PR #2, opened 2021-02-01
don't run CI twice on new pull requests

arvidn:


.github/workflows/build.yml
```diff
@@ -1,6 +1,10 @@
 name: Build
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [ main, dev ]
```
richardkiss, line 5: I find it useful to have CI run tests on all branches. It's probably okay to not upload to test pypi, but runnings tests is still good. Would this prevent it?

arvidn: this means that any new commit to these two branches will trigger CI. the `pull_request` means that any commit that is part of a PR will trigger CI. i.e. when a PR lands in one of these branches, that merge will also have CI run on it.

The previous rule meant that every commit to PR would trigger CI twice.

So, other than such duplicates, if you have some other branch, that isn't a pull request, pushing commits to it will not trigger CI. Once you make it a pull request, CI is triggered.

arvidn: If you want to keep "secret" branches and have tests run on them, we could just change this to:
```
on: push
```

richardkiss: I see, so the workflow would change to "work on a new branch, and if you want tests to run, do a draft PR". Let me try switching to that workflow and see how it goes.

PR #8, opened 2021-02-02
avoid running CI twice on pull requests

arvidn:
@hoffmang9 did you make the change deliberately to trigger CI twice for pull requests? The commit message said "On all push or pull", but the change was "On all push **and** pull", i.e. since every pull request also is a push, it's run twice.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,6 +1,9 @@
 name: Build Mac, Linux, and Windows wheels
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [ main, dev ]
+  pull_request:
 
 jobs:
   build_wheels:


hoffmang9:
The issue with that work flow is that you can't get it to run on your own branch on push. Especially when debugging CI, that's less ideal. We had been running on all pushes and PRs everywhere and the first repo that changed that was chia-blockchain over the weekend due to PR only.

arvidn:
Right, the workflow proposed with this patch is that you create a Draft PR while testing, and you don't create a PR if you don't want tests to run.

If we want to run CI on all pushes, I don't think theres any need to *also* run them for PRs. All PRs are also pushes so they should already be running.

hoffmang9:
The tradeoff is that you do want to run on community PRs. I don't have a strong opinion beyond that community members should automatically trigger a CI for their PR.

arvidn:
oh, right. PRs aren't necessarily commits in this repo

hoffmang9 (approved):
lgtm

PR #9, opened 2021-02-05
Arm64

hoffmang9:


.github/workflows/build-test-ARM64.yml
```diff
@@ -0,0 +1,128 @@
...
+    name: ARM64 Python Wheels on ARM64 Runner
+    runs-on: ARM64
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 1
+
+    - name: Build Python 3.7 wheels
```
richardkiss, line 25: Seems likely that if you have multiple `python` binaries in your path, then `maturin` will build them all at once.

hoffmang9: Having maturin try to build and uncontrolled number of python versions from the manylinux2014 docker image sounds like the very best way to get an upstream CI break when we least want it. This way we know exactly what we're getting wheel wise.

richardkiss: I believe that the only reason it's not building them all is because it doesn't find the other python instances. If we explicitly add `/opt/python/cp39-cp39/bin/` to `PATH`, that sort of "blesses" that install. Haven't explored much (or at all), but maybe it's something to consider.

hoffmang9: I just built maturin wheels for arm64 as building maturin is about 50% of runtime. I'll run a couple of races but I agree that the explicit path strategy could handle my concern nicely.

PR #11, opened 2021-02-06
implement strict- and lenient mode in serialize_and_run_program

arvidn:
This depends on an updated clvm_tools and clvm (for tests).

Specifically https://github.com/Chia-Network/clvm/pull/60 and https://github.com/Chia-Network/clvm_tools/pull/33

### DIFF
--- a/src/more_ops.rs
+++ b/src/more_ops.rs
@@ -8,10 +8,11 @@ use std::ops::BitXorAssign;
 use lazy_static::lazy_static;
 
 use crate::allocator::Allocator;
+use crate::int_allocator::IntAllocator;
 use crate::node::Node;
 use crate::number::{node_from_number, number_from_u8, Number};
 use crate::op_utils::{atom, check_arg_count, int_atom, two_ints, uint_int};
-use crate::reduction::{Reduction, Response};
+use crate::reduction::{EvalErr, Reduction, Response};
 use crate::serialize::node_to_bytes;
 
 use sha2::{Digest, Sha256};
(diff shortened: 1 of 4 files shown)


arvidn:
this implements the same behavior as https://github.com/Chia-Network/clvm/pull/60

richardkiss:
I'm still optimistic about unifying the "fast" and "slow" api... is there any chance we can defer this change for a bit?

richardkiss:
Or do we need this change very soon for some intermediate goal? I think it's best to completely encapsulate the dialect in a single structure rather than have an extra flag. But at the moment, `serialize_and_run_program` doesn't have the `OperatorDialect` argument.

arvidn:
> is there any chance we can defer this change for a bit?

My understanding is that strict and lenient mode is required for launch. A new dialect is not required until the first soft-fork.
Even in a future where the fast and slow versions are unified, this function would still be necessary, right?

src/py/run_program.rs
```diff
@@ -28,6 +36,241 @@ pub fn operator_handler2(
...
+    args: &u32,
+) -> Result<Reduction<u32>, EvalErr<u32>> {
+    if op.len() == 1 {
+        if let Some(f) = F_TABLE[op[0] as usize] {
+            let node_t: Node<IntAllocator> = Node::new(allocator, *args);
+            return f(&node_t);
+        }
+    }
+
+    // unknown opcode in lenient mode
```
richardkiss, line 98: What do you think about refactoring the stuff below here out into a separate function? We could consider putting it into `more_ops` (so that we don't have to make those costs `pub`) or even into its own file.

richardkiss: Also, I'm wondering about maybe just using the python version (which would require me to get to glue working), so we don't have to worry about discrepancies in implementation. Not convinced that there's much benefit to "speeding things up" in this case. Maybe this is a dumb idea.

arvidn: yeah, moving this code into more_ops makes sense to not have to touch the cost constants. Although, there may be value in consolidating those constants too, at some point. But I'll punt on that.

arvidn:
made the changes to move the bulk of the logic (and tests) into `more_ops.rs` and rebased on `main`

src/more_ops.rs
```diff
@@ -71,6 +72,228 @@ fn limbs_for_int(v: &Number) -> u32 {
     ((v.bits() + 7) >> 3) as u32
 }
 
+fn u32_from_u8(mut buf: &[u8]) -> Option<u32> {
```
aqk, line 75: > fn u32_from_u8

I'm surprised rust doesn't have this in their stdlib.

arvidn: well, perhaps there's something similar. this has specific overflow behavior though

src/more_ops.rs
```diff
@@ -71,6 +72,228 @@ fn limbs_for_int(v: &Number) -> u32 {
...
+        return None;
+    }
+
+    let mut ret: u32 = 0;
+    for b in buf {
+        ret <<= 8;
+        ret |= *b as u32;
+    }
+    Some(ret)
+}
```
aqk, line 96: It looks like this function will return `0` if `buf.len() == 0`

src/more_ops.rs
```diff
@@ -71,6 +72,228 @@ fn limbs_for_int(v: &Number) -> u32 {
...
+
+#[test]
+fn test_u32_from_u8() {
+    assert_eq!(u32_from_u8(&[]), Some(0));
```
aqk, line 100: Oh, there it is.

src/py/run_program.rs
```diff
@@ -36,9 +54,14 @@ pub fn serialize_and_run_program(
     quote_kw: u8,
     apply_kw: u8,
     max_cost: u32,
+    flags: u32,
```
richardkiss, line 57: I'd like to replace `flags` with an explicit reference to the function to call when we get an unknown operators, but we can do that later since it requires more complicated bridging logic mess.

richardkiss:
Cherry-picked eaa54c4495eb8a6151719e7a508a1558fec6ec29.

PR #14, opened 2021-02-08
add special cases for the pre-defined atoms 0 and 1

arvidn:


src/serialize.rs
```diff
@@ -134,6 +134,12 @@ pub fn node_from_stream<T: Allocator>(allocator: &T, f: &mut dyn Read) -> std::i
                     ops.push(ParseOp::Cons);
                     ops.push(ParseOp::SExp);
                     ops.push(ParseOp::SExp);
+                } else if b[0] == 0x00 {
+                    values.push(allocator.null());
+                } else if b[0] == 0x01 {
+                    values.push(allocator.one());
+                } else if b[0] == 0x80 {
```
richardkiss, line 141: This is a lone zero, which is NOT a null

richardkiss: Sorry, `0x80` is an empty string... it's `0x00` that a lone zero (which is not a null). I think.

arvidn: fixed

src/more_ops.rs
```diff
@@ -403,8 +403,8 @@ pub fn op_divmod<T: Allocator>(args: &Node<T>) -> Response<T::Ptr> {
         let q = &a0 / &a1;
         let r = &a0 - &a1 * &q;
 
-        let signed_quotient = (a0.sign() == Sign::Minus || a1.sign() == Sign::Minus)
-            && a0.sign() != a1.sign();
+        let signed_quotient =
+            (a0.sign() == Sign::Minus || a1.sign() == Sign::Minus) && a0.sign() != a1.sign();
```
arvidn, line 407: this formatting issue slip through earlier

richardkiss (approved):

PR #16, opened 2021-02-09
fix divmod case where quotient is 0

arvidn:
but should still be considered negative. This fixes the test case added to `clvm` here: https://github.com/Chia-Network/clvm/pull/66

### DIFF
--- a/src/more_ops.rs
+++ b/src/more_ops.rs
@@ -180,9 +180,12 @@ pub fn op_divmod<T: Allocator>(args: &Node<T>) -> Response<T::Ptr> {
         let q = &a0 / &a1;
         let r = &a0 - &a1 * &q;
 
+        let signed_quotient = (a0.sign() == Sign::Minus || a1.sign() == Sign::Minus)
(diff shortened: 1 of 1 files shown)


src/more_ops.rs
```diff
@@ -180,9 +180,12 @@ pub fn op_divmod<T: Allocator>(args: &Node<T>) -> Response<T::Ptr> {
         let q = &a0 / &a1;
         let r = &a0 - &a1 * &q;
 
+        let signed_quotient = (a0.sign() == Sign::Minus || a1.sign() == Sign::Minus)
+            && a0.sign() != a1.sign();
```
arvidn, line 184: the issue was that "zero` does not have a sign (according to `sign()`) so even though a / b was negative, `q` would not be. So this logic computes the sign explicitly, without relying on it propagating to `q`

richardkiss (approved):
Saw you added some tests to clvm, presumably at least one of them would find the thing this fixes.

PR #19, opened 2021-02-11
pass Ptr nodes by value into new_pair() to simplify ownership

arvidn:


src/allocator.rs
```diff
@@ -13,7 +13,7 @@ pub trait Allocator {
     type Ptr: Clone;
```
richardkiss, line 13: Can we get away with remove `Clone` from `type Ptr` with this change? Or does code elsewhere depend on it being cloneable?

arvidn: no, in fact I added `clone()` in a few places. I don't see how it could work without `Ptr` either being `Clone` or `Copy`

PR #22, opened 2021-02-12
simplify the complex lifetime relationship between allocator and SExp

arvidn:
Instead, ask the allocator to get the underlying bytes for an atom every time. Some of the complication here comes from the fact that the `ArcAllocator` has the `T::Ptr` own the bytes, whereas the `IntAllocator` has the allocator itself own the bytes. It especially makes it a bit tricky to unify in `Node::atom()`, that needs to return bytes whose lifetime is tied to the allocator.

Being able to break the ownership ties of a node and the allocator is important to be able to make the allocator honor mutability (i.e. make the allocating functions require a mutable allocator).

With mutable allocators, a function cannot hold an `SExp<>` (prior to this patch) when attempting to allocate a new node or pair. This is problematic because the only way to determine whether a node is a pair or an atom is to call `sexp()`, which then "burns" a reference to the allocator that cannot be undone because of the current ownership transfer.

This patch allows to break the ownership tie, and hold a mutable reference to the allocator within the match-expression.

### DIFF
--- a/src/py/arc_allocator.rs
+++ b/src/py/arc_allocator.rs
@@ -43,18 +43,30 @@ impl ArcAllocator {
 
 impl Allocator for ArcAllocator {
     type Ptr = ArcSExp;
+    type AtomBuf = Arc<Vec<u8>>;
 
     fn new_atom(&self, v: &[u8]) -> ArcSExp {
         ArcSExp::Atom(Arc::new(v.into()))
     }
 
     fn new_pair(&self, first: ArcSExp, rest: ArcSExp) -> ArcSExp {
-        ArcSExp::Pair(Arc::new(first.to_owned()), Arc::new(rest.to_owned()))
+        ArcSExp::Pair(Arc::new(first), Arc::new(rest))
     }
 
-    fn sexp<'a: 'c, 'b: 'c, 'c>(&'a self, node: &'b ArcSExp) -> SExp<'c, ArcSExp> {
+    fn atom<'a>(&'a self, node: &'a Self::Ptr) -> &'a [u8] {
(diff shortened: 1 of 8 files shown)


src/py/arc_allocator.rs
```diff
@@ -43,18 +43,30 @@ impl ArcAllocator {
...
+            _ => panic!("expected atom, got pair"),
+        }
+    }
+
+    fn buf<'a>(&'a self, node: &'a Self::AtomBuf) -> &'a [u8] {
+        &*node
+    }
+
+    fn sexp(&self, node: &ArcSExp) -> SExp<ArcSExp, Arc<Vec<u8>>> {
+        match node {
```
aqk, line 68: It's nice that it doesn't need to force you to match a "default" case.

aqk (approved):

richardkiss:
Okay, so the goal here is to get rid of `Aovec`, yes? I'd like to hold off on merging this until we're sure that going down this path until we're sure we can actually do it. But please do keep working on this.

arvidn:
> Okay, so the goal here is to get rid of Aovec, yes?

That's right. I think the motivation is a little bit stronger than just to lift the arbitrary limit on number of atoms. I would like a more efficient data structure for the atom allocations, where an arbitrary limit would be more problematic.

arvidn:
> I'd like to hold off on merging this until we're sure that going down this path until we're sure we can actually do it. But please do keep working on this.

I have a full patch that does this. Perhaps I should clean up all commits and put them in the same PR. They would most likely have to be reviewed one commit at a time though.

PR #23, opened 2021-02-12
Operator allocators

arvidn:


richardkiss:
As code, it's fine. I'd like to better understand the motivation. The way I was thinking about doing cheap substr was to add `substr` to the `Allocator` trait which would take the `Ptr` to the original string and the substr parameters, and then expand the `SExp` enum to include substr information for atoms (or have a third type which is the same as the atom type but with the saved substr info).

arvidn:
this commit sits on top of https://github.com/Chia-Network/clvm_rs/pull/22, which has motivations. This is a draft and not quite ready for review yet. But the commit message includes some motivation.

arvidn:
it's only tangentially related to substr. It's related to fixing the const-correctness of allocators, to make allocating member functions require a mutable allocator. This, in turn, will allo replacing the `Aovec` with normal vectors, which also fixes the arbitrary size limit we have right now.

It also opens the door for a more efficient allocation strategy for atoms, where they are all part of the same memory object, and all atoms, whether they are the result of a substr or not, just point into this object. I believe such representation will also remove one memory indirection for every atom lookup.

PR #25, opened 2021-02-13
Replace Aovec with Vec

arvidn:
This patch set adds support for allocators where allocating functions mutate the allocator. Specifically, this means no overlapping allocator references.

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -4,17 +4,26 @@
  *
  */
 
-pub enum SExp<'a, T> {
-    Atom(&'a [u8]),
+pub enum SExp<T, B> {
+    Atom(B),
     Pair(T, T),
 }
 
 pub trait Allocator {
     type Ptr: Clone;
+    type AtomBuf: Clone;
 
     fn new_atom(&self, v: &[u8]) -> Self::Ptr;
     fn new_pair(&self, first: Self::Ptr, rest: Self::Ptr) -> Self::Ptr;
-    fn sexp<'a: 'c, 'b: 'c, 'c>(&'a self, node: &'b Self::Ptr) -> SExp<'c, Self::Ptr>;
+
+    // The lifetime here is a bit special because IntAllocator and ArcAllocator
+    // have slightly different requirements. With IntAllocator, all buffers are
+    // owned by the allocator, with ArcAllocator all buffers have shared
+    // ownership by ArcAllocator::Ptr objects. So the returned buffer here
+    // depends on both
+    fn atom<'a>(&'a self, node: &'a Self::Ptr) -> &'a [u8];
+    fn buf<'a>(&'a self, node: &'a Self::AtomBuf) -> &'a [u8];
+    fn sexp(&self, node: &Self::Ptr) -> SExp<Self::Ptr, Self::AtomBuf>;
     fn null(&self) -> Self::Ptr;
     fn one(&self) -> Self::Ptr;
 }
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -13,8 +13,8 @@ pub trait Allocator {
     type Ptr: Clone;
     type AtomBuf: Clone;
 
-    fn new_atom(&self, v: &[u8]) -> Self::Ptr;
-    fn new_pair(&self, first: Self::Ptr, rest: Self::Ptr) -> Self::Ptr;
+    fn new_atom(&mut self, v: &[u8]) -> Self::Ptr;
+    fn new_pair(&mut self, first: Self::Ptr, rest: Self::Ptr) -> Self::Ptr;
 
     // The lifetime here is a bit special because IntAllocator and ArcAllocator
     // have slightly different requirements. With IntAllocator, all buffers are
--- a/src/node.rs
+++ b/src/node.rs
@@ -19,13 +19,13 @@ impl<'a, T: Allocator> Node<'a, T> {
         Node::new(self.allocator, node)
     }
 
-    pub fn sexp(&self) -> SExp<T::Ptr> {
+    pub fn sexp(&self) -> SExp<T::Ptr, T::AtomBuf> {
         self.allocator.sexp(&self.node)
     }
 
-    pub fn atom(&self) -> Option<&[u8]> {
+    pub fn atom(&'a self) -> Option<&'a [u8]> {
         match self.sexp() {
-            SExp::Atom(a) => Some(a),
+            SExp::Atom(_) => Some(self.allocator.atom(&self.node)),
             _ => None,
         }
     }
@@ -39,7 +39,7 @@ impl<'a, T: Allocator> Node<'a, T> {
 
     pub fn nullp(&self) -> bool {
         match self.sexp() {
-            SExp::Atom(a) => a.is_empty(),
+            SExp::Atom(a) => self.allocator.buf(&a).is_empty(),
             _ => false,
         }
     }
@@ -90,7 +90,7 @@ impl<'a, T: Allocator> PartialEq for Node<'a, T> {
             (SExp::Pair(l0, l1), SExp::Pair(r0, r1)) => {
                 self.with_node(l0) == self.with_node(r0) && self.with_node(l1) == self.with_node(r1)
             }
-            (SExp::Atom(l0), SExp::Atom(r0)) => l0 == r0,
+            (SExp::Atom(l0), SExp::Atom(r0)) => self.allocator.buf(&l0) == self.allocator.buf(&r0),
             _ => false,
         }
     }
@@ -104,7 +104,7 @@ impl<'a, T: Allocator> fmt::Debug for Node<'a, T> {
                 .field(&self.with_node(l))
                 .field(&self.with_node(r))
                 .finish(),
-            SExp::Atom(a) => a.fmt(f),
+            SExp::Atom(a) => self.allocator.buf(&a).fmt(f),
         }
     }
 }
--- a/src/node.rs
+++ b/src/node.rs
@@ -11,10 +11,6 @@ impl<'a, T: Allocator> Node<'a, T> {
         Node { allocator, node }
     }
 
-    pub fn new_atom(&self, v: &[u8]) -> Self {
-        self.with_node(self.allocator.new_atom(v))
-    }
-
     pub fn with_node(&self, node: T::Ptr) -> Self {
         Node::new(self.allocator, node)
     }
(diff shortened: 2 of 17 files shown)


arvidn:
this replaces: https://github.com/Chia-Network/clvm_rs/pull/23 and https://github.com/Chia-Network/clvm_rs/pull/22.

Each commit is made to build and pass tests individually.

arvidn:
Benchmarks run locally on MacBook. I see a pretty reliable 15% performance improvement. (I ran these 3 times each, alternating branch).

replace-aovec branch:
```
        8900023e.hex: 0.013634 0.012058 0.011530   mean: 0.012407 (+/- 0.001227)
        7312527a.hex: 0.029514 0.029361 0.027024   mean: 0.028633 (+/- 0.001609)
          substr.hex: 0.042147 0.035739 0.029708   mean: 0.035865 (+/- 0.006282)
        ca4b27bc.hex: 0.043752 0.044524 0.042186   mean: 0.043487 (+/- 0.001301)
        255822f7.hex: 0.025901 0.011738 0.013440   mean: 0.017026 (+/- 0.008875)
        c38a49ea.hex: 0.027888 0.022754 0.021611   mean: 0.024084 (+/- 0.003804)
        1f228c2e.hex: 0.008973 0.009755 0.010244   mean: 0.009657 (+/- 0.000684)
        a24336b6.hex: 0.018334 0.017690 0.018405   mean: 0.018143 (+/- 0.000453)
        4f57efb0.hex: 0.005279 0.005678 0.014421   mean: 0.008459 (+/- 0.005962)
        b8c8e6d5.hex: 0.005017 0.005007 0.010267   mean: 0.006764 (+/- 0.003503)
     large-block.hex: 0.336360 0.322661 0.317905   mean: 0.325642 (+/- 0.010718)
        1509a62d.hex: 0.052576 0.038265 0.036696   mean: 0.042512 (+/- 0.010064)
        90c890db.hex: 0.036658 0.037018 0.032114   mean: 0.035263 (+/- 0.003149)
       factorial.hex: 0.009100 0.009127 0.009044   mean: 0.009090 (+/- 0.000046)
        54fd91bb.hex: 0.019328 0.021760 0.019706   mean: 0.020265 (+/- 0.001495)
        8d4b1459.hex: 0.012321 0.013848 0.012723   mean: 0.012964 (+/- 0.000884)
       point-pow.hex: 1.291955 1.243960 1.228197   mean: 1.254704 (+/- 0.037251)
          concat.hex: 5.827664 5.088701 5.091002   mean: 5.335789 (+/- 0.491875)
        73d19fac.hex: 0.173954 0.160037 0.155193   mean: 0.163061 (+/- 0.010893)
        1c8103a4.hex: 0.026302 0.010688 0.015454   mean: 0.017481 (+/- 0.008821)
        d20e7d20.hex: 0.037612 0.033558 0.036056   mean: 0.035742 (+/- 0.002184)
TOTAL: 22.371122 s
UNCERTAINTY: 0.611080 s

TOTAL: 22.940727 s
UNCERTAINTY: 0.591909 s

TOTAL: 22.584403 s
UNCERTAINTY: 0.502554 s
```

main branch:
```
        8900023e.hex: 0.023080 0.020026 0.023892   mean: 0.022333 (+/- 0.002307)
        7312527a.hex: 0.050192 0.050810 0.050600   mean: 0.050534 (+/- 0.000342)
          substr.hex: 0.064577 0.056022 0.060157   mean: 0.060252 (+/- 0.004325)
        ca4b27bc.hex: 0.068124 0.072074 0.073395   mean: 0.071198 (+/- 0.003074)
        255822f7.hex: 0.020900 0.024043 0.023327   mean: 0.022757 (+/- 0.001857)
        c38a49ea.hex: 0.036819 0.038064 0.035401   mean: 0.036761 (+/- 0.001360)
        1f228c2e.hex: 0.021089 0.020275 0.020397   mean: 0.020587 (+/- 0.000502)
        a24336b6.hex: 0.035471 0.031480 0.035650   mean: 0.034200 (+/- 0.002720)
        4f57efb0.hex: 0.011883 0.011643 0.012619   mean: 0.012048 (+/- 0.000571)
        b8c8e6d5.hex: 0.010573 0.009761 0.011136   mean: 0.010490 (+/- 0.000729)
     large-block.hex: 1.376106 1.297808 1.328347   mean: 1.334087 (+/- 0.042019)
        1509a62d.hex: 0.066148 0.062794 0.071714   mean: 0.066885 (+/- 0.004829)
        90c890db.hex: 0.057809 0.057829 0.064509   mean: 0.060049 (+/- 0.004460)
       factorial.hex: 0.019337 0.019756 0.022606   mean: 0.020566 (+/- 0.002040)
        54fd91bb.hex: 0.034872 0.035218 0.038746   mean: 0.036279 (+/- 0.002467)
        8d4b1459.hex: 0.023382 0.029580 0.027129   mean: 0.026697 (+/- 0.003315)
       point-pow.hex: 1.284193 1.291506 1.260020   mean: 1.278573 (+/- 0.018553)
          concat.hex: 5.838707 5.059406 5.021034   mean: 5.306382 (+/- 0.532325)
        73d19fac.hex: 0.208045 0.194211 0.199933   mean: 0.200730 (+/- 0.007315)
        1c8103a4.hex: 0.039918 0.022705 0.020842   mean: 0.027822 (+/- 0.012096)
        d20e7d20.hex: 0.058394 0.063159 0.055090   mean: 0.058881 (+/- 0.004278)
TOTAL: 26.274333 s
UNCERTAINTY: 0.651483 s

TOTAL: 28.118141 s
UNCERTAINTY: 0.633593 s

TOTAL: 26.110402 s
UNCERTAINTY: 0.613428 s
```

arvidn:
The commits in this patch are split up to make sense individually and can be reviewed individually, to make reviewing easier.

richardkiss (commented):
Nice! I'm impressed you were able to pull this off. I do have some ideas to possibly simplify things a bit which might or might not be possible.

src/allocator.rs
```diff
@@ -4,17 +4,26 @@
...
+    fn new_atom(&mut self, v: &[u8]) -> Self::Ptr;
+    fn new_pair(&mut self, first: Self::Ptr, rest: Self::Ptr) -> Self::Ptr;
+
+    // The lifetime here is a bit special because IntAllocator and ArcAllocator
+    // have slightly different requirements. With IntAllocator, all buffers are
+    // owned by the allocator, with ArcAllocator all buffers have shared
+    // ownership by ArcAllocator::Ptr objects. So the returned buffer here
+    // depends on both
+    fn atom<'a>(&'a self, node: &'a Self::Ptr) -> &'a [u8];
+    fn buf<'a>(&'a self, node: &'a Self::AtomBuf) -> &'a [u8];
```
richardkiss, line 25: Maybe we can just ensure that `Self::AtomBuf : Into<'a &[u8]>` ie. is a type that supports `Into<'a &[u8]>`, and then we don't need this extra `buf<'a>` step. The type could simply keep a pointer to the allocator (if it needs it, as `IntAllocator` does), then you just go `.into()` to turn it into a `&'a [u8]`.

richardkiss: I guess the risk here is that now `sexp` might return a thing that keeps a reference to `'a`. But I think you've already done all the untangling work to make this possible.

arvidn: it cannot keep a pointer to the allocator. rust wouldn't allow creating any new atoms or pairs on the allocator while this object exists in that case. since you can't have aliases to mutable references

src/node.rs
```diff
@@ -11,21 +11,17 @@ impl<'a, T: Allocator> Node<'a, T> {
...
-    pub fn new_atom(&self, v: &[u8]) -> Self {
-        self.with_node(self.allocator.new_atom(v))
-    }
-
     pub fn with_node(&self, node: T::Ptr) -> Self {
         Node::new(self.allocator, node)
     }
 
-    pub fn sexp(&self) -> SExp<T::Ptr> {
+    pub fn sexp(&self) -> SExp<T::Ptr, T::AtomBuf> {
```
richardkiss, line 18: Another possible idea would be to tweak `Node` so it takes `mut Allocator` instead of `&'a Allocator`, so it essentially "contains" the allocator. You could only have one `Node` alive at a time, but it would essentially "store" the allocator.

arvidn: the difficutly with that is that you couldn't have `first()`, `rest()` or the ability to iterate over it, because it couldn't create more `Node` objects

richardkiss: I guess that doesn't really work because the `Result` doesn't return a `Node` in the error case, so the allocator would just be lost. Maybe it doesn't matter in that case since no more allocator references are done? Not sure. I suppose `Node` should really be renamed `Cursor` or `ProgramCounter` :) if we try this.

arvidn: I think those are the main reasons for `Node`'s existence in the first place

richardkiss:
In any case, it seems like a big win semantically since it deals with `& mut` the way it was intended. When I first wrote this code, a lot of these concepts were hazier to me than they are now.

src/allocator.rs
richardkiss: Are you sure? Since `buf` returns something `'a`, rust treats that pointer as clinging to a non-mutable lifetime of the allocator anyway. You either do something with it or discard it, but in any case it has to be gone before you call `new_pair` or `new_atom` since it has lifetime `'a`.

richardkiss: I'm pretty sure your untangling has made this possible. I really should try hacking the code though to prove it.

src/node.rs
richardkiss: I seem to think that my motivation was to provide a bunch of utility routines. A trick I didn't do was make it `Deref` to `Allocator`, which means we could call all the allocator methods on it like `new_atom`.

richardkiss: Yeah, `first` and `rest` would have to consume the old node (so it could steal the allocator) or just return a `Ptr`. Not sure how worth it it would be. In any case, I really doubt it would make the code faster, and I'm not even sure if it would make it any nicer.

richardkiss: It would get rid of the `'a` parameterization though.

src/allocator.rs
arvidn: rust lets that variable exist in the same basic block, and you can still use the aliasing mutable reference as long as that variable (that has the lifetime dependency) is not used later, in that path of the code

richardkiss: I tried changing `AtomBuf` to `(&Allocator, u32)`, but it needs a lifetime, and generic types like `type AtomBuf<'a>` are not a stable feature at the moment.

```
error[E0658]: generic associated types are unstable
  --> src/int_allocator.rs:37:5
   |
37 |     type AtomBuf<'a> = (&'a Self, u32);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #44265 <https://github.com/rust-lang/rust/issues/44265> for more information
```

richardkiss (approved):

PR #26, opened 2021-02-13
More benchmarks

arvidn:
This patch-set removes some of the really small generated benchmark programs and adds some hand-crafted ones.


benchmark/cmp.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python
+
+# tool to compare the outout of two benchmark runs
```
aqk, line 3: > outout

arvidn: fixed

aqk (approved):

PR #27, opened 2021-02-14
allocate atoms on a stack rather than individually on the heap

arvidn:
This turns the atom buffer allocator into a stack. It makes atom nodes ranges into the allocated buffer

### DIFF
--- a/src/int_allocator.rs
+++ b/src/int_allocator.rs
@@ -1,14 +1,21 @@
-use std::vec;
-
 use crate::allocator::{Allocator, SExp};
 
+#[derive(Clone, Copy)]
+pub struct IntAtomBuf {
+    start: u32,
+    end: u32,
+}
+
 enum NodePtr {
-    Atom(u32),
+    Atom(IntAtomBuf),
     Pair(u32, u32),
 }
 
 pub struct IntAllocator {
-    u8_vec: Vec<Vec<u8>>,
+    // this is effectively a grow-only stack where atoms are allocated. Atoms
+    // are immutable, so once they are created, they will stay around until the
+    // program completes
+    u8_vec: Vec<u8>,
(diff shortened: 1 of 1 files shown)


arvidn:
the windows VM on github actions is failing to make a 4 GiB allocation. This is caused by one of the benchmarks that is stress-testing heap pressure.

https://github.com/Chia-Network/clvm_rs/pull/26 is fixing the immediate failure by making the `concat` test less aggressive.

The long term fix for this is to implement heap limits on CLVM programs, in my opinion. We should probably also have size limits on single atoms, but I think that's less urgent. I'm working on a patch to allow limiting the heap size of a CLVM program.

arvidn:
This allocation strategy also opens the door for future optimizations (for platforms that support it, and have 64 bit address space) to allocate a large amount of virtual address space (with `mmap()`) and lazily commit pages as they are used, never needing to copy any memory because the arena needs to grow.

src/int_allocator.rs
```diff
@@ -1,14 +1,21 @@
-use std::vec;
-
 use crate::allocator::{Allocator, SExp};
 
+#[derive(Clone, Copy)]
+pub struct IntAtomBuf {
```
richardkiss, line 4: You might consider using tuple-type structs here like
```
pub struct IntAtomBuf (u32, u32);
```
Get the values out with `int_atom_buf.0` and `int_atom_buf.1`.

arvidn: yeah, I hadn't realized you could do that at the time. But I also like having the names. I think there's a risk the members could, incorrectly, be interpreted as `start` and `length` for instance.

richardkiss (approved):
Great! Cheap substring should be a breeze now.

PR #31, opened 2021-02-15
separate the node list into two separate vectors

arvidn:
to avoid the overhead of a variant discriminator for every item. This should decrease memory pressure.

It shows a small performance improvement in the benchmarks (about 3.5 %)

```
$ python benchmark/cmp.py before.txt after.txt
        1509a62d.hex mean: 0.022263 (-0.002348) -9.54 % (within uncertainty)
        73d19fac.hex mean: 0.046827 (-0.006502) -12.19 % (within uncertainty)
        ca4b27bc.hex mean: 0.023823 (-0.002127) -8.20 % (within uncertainty)
          concat.hex mean: 0.634571 (-0.035736) -5.33 % (within uncertainty)
      count-even.hex mean: 0.032501 (-0.003437) -9.56 % (within uncertainty)
        d20e7d20.hex mean: 0.019012 (-0.002185) -10.31 % (within uncertainty)
       factorial.hex mean: 0.178255 (-0.004526) -2.48 % (within uncertainty)
     hash-string.hex mean: 0.337750 (-0.003297) -0.97 % (within uncertainty)
       hash-tree.hex mean: 0.070640 (-0.007488) -9.58 % (within uncertainty)
     large-block.hex mean: 0.427025 (-0.032651) -7.10 % (within uncertainty)
 matrix-multiply.hex mean: 0.377796 (-0.034381) -8.34 %
       point-pow.hex mean: 1.259553 (-0.006682) -0.53 % (within uncertainty)
     pubkey-tree.hex mean: 0.463226 (-0.005521) -1.18 % (within uncertainty)
      shift-left.hex mean: 2.317158 (-0.046278) -1.96 % (within uncertainty)
     substr-tree.hex mean: 0.385590 (-0.025769) -6.26 % (within uncertainty)
          substr.hex mean: 0.116922 (-0.005582) -4.56 % (within uncertainty)
        sum-tree.hex mean: 0.594497 (-0.041585) -6.54 % (within uncertainty)
TOTAL: 365.370418 (-13.304822) -3.51 %
```

### DIFF
--- a/src/int_allocator.rs
+++ b/src/int_allocator.rs
@@ -6,9 +6,10 @@ pub struct IntAtomBuf {
     end: u32,
 }
 
-enum NodePtr {
-    Atom(IntAtomBuf),
-    Pair(u32, u32),
+#[derive(Clone, Copy)]
+pub struct IntPair {
+    first: i32,
+    rest: i32,
 }
 
 pub struct IntAllocator {
@@ -16,7 +17,14 @@ pub struct IntAllocator {
     // are immutable, so once they are created, they will stay around until the
     // program completes
     u8_vec: Vec<u8>,
-    node_vec: Vec<NodePtr>,
+
+    // storage for all pairs (positive indices)
+    pair_vec: Vec<IntPair>,
+
+    // storage for all atoms (negative indices).
+    // node index -1 refers to index 0 in this vector, -2 refers to 1 and so
+    // on.
+    atom_vec: Vec<IntAtomBuf>,
 }
 
 impl Default for IntAllocator {
@@ -29,44 +37,45 @@ impl IntAllocator {
     pub fn new() -> Self {
         let mut r = IntAllocator {
             u8_vec: Vec::new(),
-            node_vec: Vec::new(),
+            pair_vec: Vec::new(),
+            atom_vec: Vec::new(),
         };
         r.u8_vec.reserve(1024 * 1024);
+        r.atom_vec.reserve(256);
+        r.pair_vec.reserve(256);
         r.u8_vec.push(1_u8);
         // Preallocated empty list
-        r.node_vec
-            .push(NodePtr::Atom(IntAtomBuf { start: 0, end: 0 }));
+        r.atom_vec.push(IntAtomBuf { start: 0, end: 0 });
         // Preallocated 1
-        r.node_vec
-            .push(NodePtr::Atom(IntAtomBuf { start: 0, end: 1 }));
+        r.atom_vec.push(IntAtomBuf { start: 0, end: 1 });
         r
     }
 }
 
 impl Allocator for IntAllocator {
-    type Ptr = u32;
+    type Ptr = i32;
     type AtomBuf = IntAtomBuf;
 
     fn new_atom(&mut self, v: &[u8]) -> Self::Ptr {
         let start = self.u8_vec.len() as u32;
         self.u8_vec.extend_from_slice(v);
         let end = self.u8_vec.len() as u32;
-        let r = self.node_vec.len() as u32;
-        self.node_vec.push(NodePtr::Atom(IntAtomBuf { start, end }));
-        r
+        self.atom_vec.push(IntAtomBuf { start, end });
+        -(self.atom_vec.len() as i32)
     }
 
     fn new_pair(&mut self, first: Self::Ptr, rest: Self::Ptr) -> Self::Ptr {
-        let r: u32 = self.node_vec.len() as u32;
-        self.node_vec.push(NodePtr::Pair(first, rest));
+        let r = self.pair_vec.len() as i32;
+        self.pair_vec.push(IntPair { first, rest });
         r
     }
 
     fn atom<'a>(&'a self, node: &'a Self::Ptr) -> &'a [u8] {
-        match self.node_vec[*node as usize] {
-            NodePtr::Atom(IntAtomBuf { start, end }) => &self.u8_vec[start as usize..end as usize],
-            _ => panic!("expected atom, got pair"),
+        if *node >= 0 {
+            panic!("expected atom, got pair");
         }
+        let atom = self.atom_vec[(-*node - 1) as usize];
+        &self.u8_vec[atom.start as usize..atom.end as usize]
     }
 
     fn buf<'a>(&'a self, node: &'a Self::AtomBuf) -> &'a [u8] {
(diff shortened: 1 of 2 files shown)


src/int_allocator.rs
```diff
@@ -6,17 +6,25 @@ pub struct IntAtomBuf {
...
     // this is effectively a grow-only stack where atoms are allocated. Atoms
     // are immutable, so once they are created, they will stay around until the
     // program completes
     u8_vec: Vec<u8>,
-    node_vec: Vec<NodePtr>,
+
+    // storage for all pairs (positive indices)
+    pair_vec: Vec<IntPair>,
+
+    // storage for all atomes (negative indices).
```
aqk, line 24: > atomes

src/int_allocator.rs
```diff
@@ -29,62 +37,66 @@ impl IntAllocator {
...
+            atom_vec: Vec::new(),
         };
         r.u8_vec.reserve(1024 * 1024);
+        r.atom_vec.reserve(256);
+        r.pair_vec.reserve(256);
         r.u8_vec.push(1_u8);
         // Preallocated empty list
-        r.node_vec
-            .push(NodePtr::Atom(IntAtomBuf { start: 0, end: 0 }));
+        r.atom_vec.push(IntAtomBuf { start: 0, end: 0 });
```
aqk, line 48: Are these lists pointers to growable memory?

arvidn: they aren't pointers, they're indices. And yes, the vectors they point into can grow.

arvidn: indices have two main advantages:

1. the memory can grow without having to update pointers
2. we can use 32 bit integers instead of full 64 bit pointers, decreasing memory pressure

src/int_allocator.rs
arvidn: I'll force-push an update to fix this in a bit

src/int_allocator.rs
```diff
@@ -29,62 +37,66 @@ impl IntAllocator {
...
 
     fn new_atom(&mut self, v: &[u8]) -> Self::Ptr {
         let start = self.u8_vec.len() as u32;
         self.u8_vec.extend_from_slice(v);
         let end = self.u8_vec.len() as u32;
-        let r = self.node_vec.len() as u32;
-        self.node_vec.push(NodePtr::Atom(IntAtomBuf { start, end }));
-        r
+        self.atom_vec.push(IntAtomBuf { start, end });
+        -(self.atom_vec.len() as i32)
```
aqk, line 64: Can `self.atom_vec.len() == 0` in this scheme?

src/int_allocator.rs
```diff
@@ -29,62 +37,66 @@ impl IntAllocator {
     pub fn new() -> Self {
         let mut r = IntAllocator {
             u8_vec: Vec::new(),
-            node_vec: Vec::new(),
+            pair_vec: Vec::new(),
+            atom_vec: Vec::new(),
         };
         r.u8_vec.reserve(1024 * 1024);
+        r.atom_vec.reserve(256);
```
aqk, line 44: Oh, so the atoms and pairs are held on two separate stacks.

src/int_allocator.rs
```diff
@@ -29,62 +37,66 @@ impl IntAllocator {
...
         r
     }
 }
 
 impl Allocator for IntAllocator {
-    type Ptr = u32;
+    type Ptr = i32;
     type AtomBuf = IntAtomBuf;
 
     fn new_atom(&mut self, v: &[u8]) -> Self::Ptr {
```
aqk, line 59: I assume the return value of `fn new_atom` is an index / ID for that object

aqk (commented):
Looks good. Mostly I have a few questions about how things work.

It seems safe to use the sign bit, even in a fixed-width type, because the index into the in-memory buffer is wider than the possible (negative, virtual) index values.

src/int_allocator.rs
arvidn: in theory it could, but there are always two atoms that are pre-allocated. `1` and `()`, so it won't. Also, these atom-indices should only ever come back from `new_atom()`, so they should also never be out-of-range

src/int_allocator.rs
arvidn: yes, that's the core of this patch

src/int_allocator.rs
arvidn: that's right

arvidn:
fun trivia: Quake 1 used a similar representation for its BSP tree, where negative indices referred to leaves and positive to internal nodes.

src/int_allocator.rs
aqk, line 48: > // Preallocated empty list
This comment is inaccurate now

src/int_allocator.rs
```diff
@@ -29,62 +37,66 @@ impl IntAllocator {
...
 
     fn buf<'a>(&'a self, node: &'a Self::AtomBuf) -> &'a [u8] {
         &self.u8_vec[node.start as usize..node.end as usize]
     }
 
     fn sexp(&self, node: &Self::Ptr) -> SExp<Self::Ptr, Self::AtomBuf> {
-        match self.node_vec[*node as usize] {
-            NodePtr::Atom(atombuf) => SExp::Atom(atombuf),
-            NodePtr::Pair(left, right) => SExp::Pair(left, right),
+        if *node >= 0 {
```
aqk, line 86: I'm worried about atom/pair type confusion here (clvm types, not rust types).
In the last version, null was type `Atom`, but now, index `0` means "first item in the pair list", if I read this correctly. (And index `-1` is the new null)

> if *node >= 0
Then we return SExp::Pair, but I *think* that was `SExp::Atom` before.

aqk: Wait, doh. So we still return `SExp::Atom` now.

arvidn: the value `IntAllocator::Ptr` has no relation to the values of anything in clvm. In fact, this type can be an actual pointer too (when using the `ArcAllocator`). The `i32` value of `Ptr` is never available to the clvm program

aqk: Thanks, that helps!

But what I meant was: when we construct a `SExp` here, if `node >= 0`, we construct a `SExp::Pair`. However, above, we pre-declare `null` on the *atom* stack:

> r.atom_vec.push(IntAtomBuf { start: 0, end: 0 });

aqk: And then I realized, oh, but the input to this function would be `-1` to mean `null`

arvidn: right

richardkiss (approved):

PR #32, opened 2021-02-16
zero-copy substr

arvidn:
This PR sits on top of https://github.com/Chia-Network/clvm_rs/pull/31, which should land first.

This patch implements zer-copy substring, by extending the allocator with a new function to create an atom based on a part of an existing atom. Since all atoms are immutable, the new atom can simply point into the firsts memory.

There are two benchmarks that exercise `substr`, it's really only one of them that has a measurable improvement (but it's quite substantial):

```
    substr-tree.hex mean: 0.378359 (-0.008424) -2.18 % (within uncertainty)
          substr.hex mean: 0.018781 (-0.092265) -83.09 %
```

All other benchmarks are unaffected (within the random variability).

### DIFF
--- a/src/int_allocator.rs
+++ b/src/int_allocator.rs
@@ -70,6 +70,28 @@ impl Allocator for IntAllocator {
         r
     }
 
+    fn new_substr(&mut self, node: Self::Ptr, start: u32, end: u32) -> Self::Ptr {
+        if node >= 0 {
+            panic!("substr expected atom, got pair");
+        }
+        let atom = self.atom_vec[(-node - 1) as usize];
+        let atom_len = atom.end - atom.start;
+        if start > atom_len {
+            panic!("substr start out of bounds");
+        }
+        if end > atom_len {
+            panic!("substr end out of bounds");
+        }
+        if end < start {
+            panic!("substr invalid bounds");
+        }
+        self.atom_vec.push(IntAtomBuf {
+            start: atom.start + start,
+            end: atom.start + end,
+        });
+        -(self.atom_vec.len() as i32)
+    }
+
     fn atom<'a>(&'a self, node: &'a Self::Ptr) -> &'a [u8] {
         if *node >= 0 {
             panic!("expected atom, got pair");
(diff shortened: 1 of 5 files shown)


src/int_allocator.rs
```diff
@@ -70,6 +70,28 @@ impl Allocator for IntAllocator {
         r
     }
 
+    fn new_substr(&mut self, node: Self::Ptr, start: u32, end: u32) -> Self::Ptr {
+        if node >= 0 {
```
richardkiss, line 74: Maybe just wrap in `Result` rather than panic.

src/more_ops.rs
```diff
@@ -469,17 +469,14 @@ pub fn op_substr<T: Allocator>(a: &mut T, input: T::Ptr) -> Response<T::Ptr> {
     let a0 = args.first()?;
     let s0 = atom(&a0, "substr")?;
     let (n1, _, n2, _) = two_ints(&args.rest()?, "substr")?;
-    let i1: isize = isize::try_from(n1).unwrap_or(isize::max_value());
-    let i2: isize = isize::try_from(n2).unwrap_or(0);
-    let size = s0.len() as isize;
-    if i2 > size || i2 < i1 || i2 < 0 || i1 < 0 {
+    let i1 = u32::try_from(n1).unwrap_or(u32::max_value());
+    let i2 = u32::try_from(n2).unwrap_or(0);
```
richardkiss, line 473: This seems like it would pin negative numbers to `0`, which may not be what we want.

richardkiss: (I realize this isn't new code)

src/int_allocator.rs
arvidn: Yes, my next patch introduces the ability for allocations to fail, I'm fixing this there

src/more_ops.rs
arvidn: ah, yes. negative numbers should be failures. We have tests for that though, right? I'll look into this

arvidn: we have this test:
```
clvm/tests/brun/substr-3.txt
brun -c '(substr (q "abcdefghijkl") 2 5)' '(-1 12)'
FAIL: invalid indices for substr ("abcdefghijkl" -1 12)
```

arvidn: this is a real issue that isn't covered by any test case. I'll fix it and add a test case (to the `clvm` repo)

richardkiss (approved):

PR #33, opened 2021-02-16
use constants instead of function pointers in op_stack

arvidn:
to lower memory pressure and improve branch prediction. This improves some benchmarks by about 10%.

```
     large-block.hex mean: 0.405073 (-0.053889) -11.74 %
 matrix-multiply.hex mean: 0.368093 (-0.048388) -11.62 %
        sum-tree.hex mean: 0.553587 (-0.085877) -13.43 %
```

The other benchmarks are within the random variability. I think the `large-block` test carries some extra weight, as it's the kinds of programs we expect to run on the chain.

### DIFF
--- a/src/run_program.rs
+++ b/src/run_program.rs
@@ -29,8 +29,14 @@ pub type PreEval<A> = Box<
 
 pub type PostEval<T> = dyn Fn(Option<&<T as Allocator>::Ptr>);
 
-type RpcOperator<T> =
-    dyn FnOnce(&mut RunProgramContext<T>) -> Result<u32, EvalErr<<T as Allocator>::Ptr>>;
+#[repr(u8)]
+enum Operation {
+    Apply,
+    Cons,
+    Eval,
+    Swap,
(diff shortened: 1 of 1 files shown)


src/run_program.rs
```diff
@@ -41,8 +47,9 @@ pub struct RunProgramContext<'a, T: Allocator> {
     apply_kw: u8,
     operator_lookup: Box<dyn OperatorHandler<T>>,
     pre_eval: Option<PreEval<T>>,
+    posteval_stack: Vec<Box<PostEval<T>>>,
     val_stack: Vec<T::Ptr>,
-    op_stack: Vec<Box<RpcOperator<T>>>,
+    op_stack: Vec<Operation>,
```
richardkiss, line 52: You might get the same benefit (and reduce complexity) if you just unbox the functions. I didn't think about moving `PostEval` to its own stack, and it requires boxing, but the others all seem static.

richardkiss (approved):

PR #34, opened 2021-02-16
Renumber opcodes

richardkiss:
Add `deserialize_and_run_program` which takes a hash table of opcode indices and named operator references.

### DIFF
--- a/src/py/f_table.rs
+++ b/src/py/f_table.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use crate::allocator::Allocator;
 use crate::core_ops::{op_cons, op_eq, op_first, op_if, op_listp, op_raise, op_rest};
 use crate::more_ops::{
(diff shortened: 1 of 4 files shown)


.github/workflows/build-test.yml
```diff
@@ -108,7 +108,7 @@ jobs:
         git clone https://github.com/Chia-Network/clvm.git --branch=develop --single-branch
         python -m pip install ./clvm
 
-        git clone https://github.com/Chia-Network/clvm_tools.git --branch=develop --single-branch
+        git clone https://github.com/Chia-Network/clvm_tools.git --branch=deserialize_api --single-branch
```
richardkiss, line 111: This is just for testing the new api, once we're convinced it works with this branch, we can remove it and merge this (and then merge clvm_tools).

src/py/run_program.rs
```diff
@@ -53,13 +52,77 @@ pub fn serialize_and_run_program(
     apply_kw: u8,
     max_cost: u32,
     flags: u32,
+) -> PyResult<(u32, Py<PyBytes>)> {
+    let mut opcode_lookup_by_name = HashMap::<String, Vec<u8>>::new();
```
arvidn, line 56: I thought the idea that this list would be defined by `clvm`. This is here for backwards compatibility with this API, yeah?
Since this is a global, immutable, constant, is there a way to do this once, rather than every call to `serialize_and_run_program()`?

arvidn (approved):
looks good. I think perhaps an opcode > 255 shouldn't be silently ignored, but at least cause a failure

src/py/f_table.rs
```diff
@@ -11,43 +13,62 @@ type OpFn<T> = fn(&mut T, <T as Allocator>::Ptr) -> Response<<T as Allocator>::P
...
-        (27, op_lognot),
-        (28, op_ash),
-        (29, op_lsh),
-        (30, op_not),
-        (31, op_any),
-        (32, op_all),
-        (33, op_softfork),
-        (34, op_div),
+pub fn opcode_by_name<T: Allocator>(name: &str) -> Option<OpFn<T>> {
+    let opcode_lookup: [(OpFn<T>, &str); 30] = [
```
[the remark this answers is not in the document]
richardkiss: Ultimately, life will be easiest if I can actually just expose the rust implementation to python as a python object, and python can pass in a `dict` of opcodes to lookup, turn it into a `NativeOpLookup`, and then reuse that on subsequent `run_program` calls. So this ugly, less-than-perfect bit is just a temporary measure for expediency.

src/py/f_table.rs
```diff
@@ -11,43 +13,62 @@ type OpFn<T> = fn(&mut T, <T as Allocator>::Ptr) -> Response<<T as Allocator>::P
...
+    let mut f_lookup = [None; 256];
+    for (name, idx) in opcode_lookup_by_name.iter() {
+        if idx.len() == 1 {
+            let index = idx[0];
+            let op = opcode_by_name(name);
+            if op.is_none() {
+                panic!("can't find native operator {:?}", name);
+            }
+            f_lookup[index as usize] = op;
+        }
```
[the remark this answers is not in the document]
richardkiss: Again, yes, gross for expediency. Not intended to be the final api.

src/py/run_program.rs
richardkiss: Expediency, with a plan to something less gross later.

PR #38, opened 2021-02-16
heap limits

arvidn:
and implement a heap- and atom size limits

### DIFF
--- a/src/limits.rs
+++ b/src/limits.rs
@@ -1,0 +1,2 @@
+pub const MAX_CUMULATIVE_HEAP_SIZE: usize = 200 * 1024 * 1024;
+pub const MAX_ATOM_SIZE: usize = 1024 * 1024;
--- a/src/py/arc_allocator.rs
+++ b/src/py/arc_allocator.rs
@@ -1,5 +1,7 @@
 use crate::allocator::{Allocator, SExp};
-
+use crate::err_utils::err;
+use crate::limits::{MAX_ATOM_SIZE, MAX_CUMULATIVE_HEAP_SIZE};
+use crate::reduction::EvalErr;
 use std::sync::Arc;
 
 use lazy_static::*;
@@ -8,7 +10,9 @@ use pyo3::prelude::*;
 
 #[pyclass(subclass, unsendable)]
 #[derive(Clone)]
-pub struct ArcAllocator {}
+pub struct ArcAllocator {
+    cumulative_allocations: usize,
+}
 
 #[derive(Clone)]
 pub struct ArcAtomBuf {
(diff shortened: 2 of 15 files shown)


src/limits.rs
```diff
@@ -0,0 +1,2 @@
+pub const MAX_CUMULATIVE_HEAP_SIZE: usize = 200 * 1024 * 1024;
+pub const MAX_ATOM_SIZE: usize = 1024 * 1024;
```
arvidn, line 2: I think these limits should be significantly higher than we would expect any normal program to need, does this seem like reasonable limits?

richardkiss: Don't use constants, make them parameters.

src/py/arc_allocator.rs
```diff
@@ -38,10 +42,12 @@ impl Clone for ArcSExp {
 
 impl ArcAllocator {
     pub fn new() -> Self {
-        ArcAllocator {}
+        ArcAllocator {
+            cumulative_allocations: 0,
```
richardkiss, line 46: This won't work because `ArcAllocator` objects interoperate.

richardkiss:
Hard limits not associated with costs are consensus risks and make it much more likely that different implementations will have hard-to-find differences, so I'm not a big fan of some of these changes.

src/py/arc_allocator.rs
arvidn: there's more than one allocator?

arvidn:
> Hard limits not associated with costs are consensus risks and make it much more likely that different implementations will have hard-to-find differences, so I'm not a big fan of some of these changes.

Right. However, the alternative (as far as I can tell) is to have costs of operations be proportional to the cumulative allocated memory so far. Otherwise the cost doesn't grow fast enough once you start using a lot of memory.

arvidn:
Actually, the upper limit of atoms is trivial to implement and unlikely (I would say) to cause consensus divergence. The cumulative allocation is a bit easier to get wrong I suppose. As it turns out the `ArcAllocator` apparently can't keep state.

richardkiss:
Yeah, allocators that use GC like `ArcAllocator` are going to have a very hard time with state across a given program. I think artificially increasing the cost of allocation is our best bet even though it might seem a bit wasteful. We should pick a maximum amount of memory we want to allocate and adjust our per-byte costs accordingly. Mostly `concat` – not sure it's possible to waste much memory with any other operators. Maybe arithmetic.

arvidn:
I broke out, what I believe to be, the uncontroversial changes: https://github.com/Chia-Network/clvm_rs/pull/40
Thats just to support the error path from allocators.

PR #43, opened 2021-02-17
Remove our maturin wheel in preference to Pypi

hoffmang9:


.github/workflows/build-arm64-wheels.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - dev
+      - no-simple-arm
```
richardkiss, line 8: Maybe take this out now that you know it works?

hoffmang9: Meant to and got distracted.

richardkiss (approved):

PR #44, opened 2021-02-18
Enforce max cost

arvidn:
this patch adds a test that generates clvm programs pushing the limits on single operations that are expensive. It demonstrates the need for all operators that take an unbounded number of arguments to be able to fail early, when max cost is exceeded.

It fixes this problem by passing along the remaining cost into all operator functions.

### DIFF
--- a/src/py/f_table.rs
+++ b/src/py/f_table.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 
 use crate::allocator::Allocator;
 use crate::core_ops::{op_cons, op_eq, op_first, op_if, op_listp, op_raise, op_rest};
+use crate::cost::Cost;
 use crate::more_ops::{
     op_add, op_all, op_any, op_ash, op_concat, op_div, op_divmod, op_gr, op_gr_bytes, op_logand,
     op_logior, op_lognot, op_logxor, op_lsh, op_multiply, op_not, op_point_add, op_pubkey_for_exp,
@@ -9,7 +10,7 @@ use crate::more_ops::{
 };
 use crate::reduction::Response;
 
-type OpFn<T> = fn(&mut T, <T as Allocator>::Ptr) -> Response<<T as Allocator>::Ptr>;
+type OpFn<T> = fn(&mut T, <T as Allocator>::Ptr, Cost) -> Response<<T as Allocator>::Ptr>;
 
 pub type FLookup<T> = [Option<OpFn<T>>; 256];
 
--- a/src/run_program.rs
+++ b/src/run_program.rs
@@ -142,7 +142,6 @@ fn augment_cost_errors<P: Clone>(
         return r;
     }
     let e = r.unwrap_err();
-    println!("error: {}", e.1);
     if &e.1 != "cost exceeded" {
         Err(e)
     } else {
@@ -329,16 +328,17 @@ where
         &mut self,
         program: &T::Ptr,
         args: &T::Ptr,
-        mut max_cost: Cost,
+        max_cost: Cost,
     ) -> Response<T::Ptr> {
         self.val_stack = vec![self.allocator.new_pair(program.clone(), args.clone())?];
         self.op_stack = vec![Operation::Eval];
 
-        if max_cost == 0 {
-            max_cost = Cost::MAX;
-        }
+        // max_cost is always in effect, and necessary to prevent wrap-around of
+        // the cost integer.
+        let max_cost = if max_cost == 0 { Cost::MAX } else { max_cost };
+
         let max_cost_number: Number = max_cost.into();
-        let max_cost_ptr = ptr_from_number(self.allocator, &max_cost_number);
+        let max_cost_ptr = ptr_from_number(self.allocator, &max_cost_number)?;
 
         let mut cost: Cost = 0;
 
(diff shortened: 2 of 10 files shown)


src/run_program.rs
```diff
@@ -133,6 +134,22 @@ fn traverse_path<T: Allocator>(
...
+fn augment_cost_errors<A: Clone>(
+    r: Result<Cost, EvalErr<A>>,
+    max_cost: &A,
+) -> Result<Cost, EvalErr<A>> {
+    if r.is_ok() {
+        return r;
+    }
+    let e = r.unwrap_err();
+    println!("error: {}", e.1);
+    if &e.1 != "cost exceeded" {
```
arvidn, line 146: this is not very pretty. However:

1. Our error messages are fundamentally based on strings (rather than error codes e.g.) so I don't see another way to distinguish error types.
2. The alternative, as far as I can tell, would be to pass this information down all the way everywhere this error can happen, which seems worse.

arvidn: another alternative (that I'm not sure I like any more) would be to store the max cost inside the allocator, to make it available to anyone creating this kind of error

src/run_program.rs
```diff
@@ -312,11 +329,17 @@ where
...
         program: &T::Ptr,
         args: &T::Ptr,
-        max_cost: Cost,
+        mut max_cost: Cost,
     ) -> Response<T::Ptr> {
         self.val_stack = vec![self.allocator.new_pair(program.clone(), args.clone())?];
         self.op_stack = vec![Operation::Eval];
 
+        if max_cost == 0 {
+            max_cost = Cost::MAX;
```
arvidn, line 338: with this patch, `max_cost` must always be set, this is how integer wrap-arounds are prevented

arvidn: setting it to the max value effectively means infinite. I'm kind of thinking that we might want a more reasonable default than this though.

src/run_program.rs
```diff
@@ -326,9 +349,11 @@ where
                 None => break,
             };
             cost += match op {
-                Operation::Apply => self.apply_op()?,
+                Operation::Apply => {
+                    augment_cost_errors(self.apply_op(max_cost - cost), &max_cost_ptr)?
```
arvidn, line 353: the `augment_cost_errors()` is the moral equivalence of catching `CostError` exceptions, setting the `max_cost` field and rethrowing.

tests/generate-programs.py
```diff
@@ -0,0 +1,83 @@
...
+many_args('programs/args-or.clvm', 'logior', 6000)
+many_args('programs/args-xor.clvm', 'logxor', 6000)
+many_args_point('programs/args-point_add.clvm', 'point_add', 12000)
+many_args('programs/args-unknown-1.clvm', '0x7fffffff00', 300)
+many_args('programs/args-unknown-2.clvm', '0x7fff40', 300)
+many_args('programs/args-unknown-3.clvm', '0x7fff80', 300)
+many_args('programs/args-unknown-4.clvm', '0x7fffc0', 300)
+
+# this program attempts to wrap around a 64 bit cost counter
+softfork_wrap('programs/softwork-1.clvm', '0x00ffffffffffffff45')
```
arvidn, line 80: I'm pretty excited that I managed to demonstrate this. I ended up in what appeared to be an infinite loop, where cost would not grow over time. (this is fixed now, that cost is not allowed to wrap-around)

aqk (approved):

tests/generate-programs.py
aqk: That is bad-ass.

src/run_program.rs
aqk: > max_cost must always be set, this is how integer wrap-arounds are prevented
> setting it to the max value effectively means infinite

You should comment that

arvidn: yeah, I should. I should also change this code to be more true to modern programming practices:

```
let max_cost = if max_cost == 0 { Cost::MAX } else { max_cost };
```

arvidn: (so `max_cost` can be left immutable)

src/py/f_table.rs
```diff
@@ -2,14 +2,15 @@ use std::collections::HashMap;
...
+use crate::cost::Cost;
 use crate::more_ops::{
     op_add, op_all, op_any, op_ash, op_concat, op_div, op_divmod, op_gr, op_gr_bytes, op_logand,
     op_logior, op_lognot, op_logxor, op_lsh, op_multiply, op_not, op_point_add, op_pubkey_for_exp,
     op_sha256, op_softfork, op_strlen, op_substr, op_subtract,
 };
 use crate::reduction::Response;
 
-type OpFn<T> = fn(&mut T, <T as Allocator>::Ptr) -> Response<<T as Allocator>::Ptr>;
+type OpFn<T> = fn(&mut T, <T as Allocator>::Ptr, Cost) -> Response<<T as Allocator>::Ptr>;
```
richardkiss, line 13: If you make this an `enum` with two flavors of function signature, one of which doesn't take a second parameter, you can avoid making all the gratuitous signature changes to the functions that can't possibly overflow.

src/run_program.rs
richardkiss: You might use `P` instead of `A` here to suggest `Ptr`. At first I thought `A` was for `Allocator`. Not a critical fix though.

tests/generate-programs.py
richardkiss, line 80: You probably want `s/softwork/softfork/`

src/py/f_table.rs
richardkiss: Might not be worth it, not sure. But it's definitely a possibility.

richardkiss (approved):

src/py/f_table.rs
arvidn: yeah, I suspect it would end up being more complex overall though. an unused parameter is pretty simple and straight-forward (and no need to store a discriminator next to each function pointer)

tests/generate-programs.py
arvidn: oops

PR #47, opened 2021-02-19
Fix parser

arvidn:
This extends the test case generator to create serialized files with very large atoms (that are truncated on disk). This triggers the rust implementation to allocate a byte vector of 17179869183 bytes.

~~Somehow this doesn't seem to be a problem on 64 bit mac, but presumably this would fail on a 32 bit system.~~ This is a problem and causes CI to fail without the fix in this patch.

This patch replaces some `panic!()` with propagating errors up the call stack, as well as turning a few `unwrap()` into reporting errors as well.

### DIFF
--- a/src/serialize.rs
+++ b/src/serialize.rs
@@ -14,6 +14,10 @@ fn bad_encoding() -> std::io::Error {
     Error::new(ErrorKind::InvalidInput, "bad encoding")
 }
 
+fn internal_error() -> std::io::Error {
+    Error::new(ErrorKind::InvalidInput, "internal error")
+}
+
 fn encode_size(f: &mut dyn Write, size: u64) -> std::io::Result<()> {
     if size < 0x40 {
         f.write_all(&[(0x80 | size) as u8])?;
@@ -41,7 +45,7 @@ fn encode_size(f: &mut dyn Write, size: u64) -> std::io::Result<()> {
             ((size) & 0xff) as u8,
         ])?;
     } else {
-        panic!("atom size exceeded maximum {}", size);
+        return Err(Error::new(ErrorKind::InvalidData, "atom too big"));
     }
     Ok(())
 }
@@ -82,7 +86,10 @@ fn decode_size(f: &mut dyn Read, initial_b: u8) -> std::io::Result<u64> {
     // this function decodes the length prefix for an atom. Atoms whose value
     // fit in 7 bits don't have a length-prefix, so those should never be passed
     // to this function.
-    assert!((initial_b & 0x80) != 0);
+    debug_assert!((initial_b & 0x80) != 0);
+    if (initial_b & 0x80) == 0 {
+        return Err(internal_error());
+    }
 
     let mut bit_count = 0;
     let mut bit_mask: u8 = 0x80;
@@ -208,14 +215,10 @@ fn test_encode_size() {
     let mut buf = Vec::<u8>::new();
     assert!(encode_size(&mut buf, 0x3ffffffff).is_ok());
     assert_eq!(buf, vec![0b11111011, 0xff, 0xff, 0xff, 0xff]);
-}
 
-#[test]
-#[should_panic]
-fn test_encode_panic() {
     // this is too large
     let mut buf = Vec::<u8>::new();
-    assert!(encode_size(&mut buf, 0x400000000).is_ok());
+    assert!(!encode_size(&mut buf, 0x400000000).is_ok());
 }
 
 #[test]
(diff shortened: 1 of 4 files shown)


src/serialize.rs
```diff
@@ -127,7 +134,7 @@ impl<T> std::convert::From<EvalErr<T>> for std::io::Error {
 
 pub fn node_from_stream<T: Allocator>(
     allocator: &mut T,
-    f: &mut dyn Read,
+    f: &mut Cursor<&[u8]>,
```
arvidn, line 137: this is core to this patch. This enables the code below to validate the atom size against the available buffer left, thus preventing an attack with an invalid atom lenth

src/serialize.rs
```diff
@@ -41,7 +45,7 @@ fn encode_size(f: &mut dyn Write, size: u64) -> std::io::Result<()> {
             ((size) & 0xff) as u8,
         ])?;
     } else {
-        panic!("atom size exceeded maximum {}", size);
```
arvidn, line 44: this is somewhat unrelated, but we shouldn't have `panic!()`s. It's better to fail the clvm program than to terminate the process.

src/serialize.rs
richardkiss: This seems like it loses some flexibility in that it can no longer read from a stream (handing back a stream without reading it all has a use case, possibly even with `chia-blockchain`). I think you might want two versions of deserialize, one that takes a blob (which you can wrap in a cursor if you want), immediately adds it to the allocator, and then all other atoms can be a cheap `substr`. What do you think about keeping around a (less-optimal) `dyn Read` version too?

richardkiss (changes requested):
Keep old API too.

arvidn:
which API?

src/serialize.rs
arvidn: (I didn't see this message on github at first).

> What do you think about keeping around a (less-optimal) dyn Read version too?

This isn't about efficiency. It's about being exploitable. As far as I can tell, this exact API *cannot* be made safe (unless we introduce a reasonably small atom size limit I suppose).

We could make it take a stream *and* a max size that can be read from the stream. But since a stream represents a (possibly) infinite sequence of bytes, we can't allocate that many.

Also, regarding the flexibility; I think we shouldn't have that flexibility until we need it *and* we have solved this problem. I don't think it makes sense to spend time solving this problem (the exploit) for a difficult case *until* we actually need it to be solved.

arvidn: I suppose mitigation would be to internally read large atoms in small bite sizes. say, 1kiB at a time. That way, if we actually hit end-of-file, we get an error. It's not perfect though, one just have to imagine a gzip stream that can produce a very large output from a very small input to make the memory use explode.

tests/generate-programs.py
```diff
@@ -81,3 +124,9 @@ def softfork_wrap(filename, val):
 # this program attempts to wrap around a 32 bit cost counter
 softfork_wrap('programs/softfork-2.clvm', '0x00ffffff45')
 
+# tests that try to trick a parser into allocating too much memory
+serialized_atom_overflow('programs/large-atom-1.hex', 0xffffffff)
+serialized_atom_overflow('programs/large-atom-2.hex', 0x3ffffffff)
+serialized_atom_overflow('programs/large-atom-3.hex', 0xffffffffff)
+serialized_atom_overflow('programs/large-atom-4.hex', 0xfffffffffff)
```
arvidn, line 131: @richardkiss the challenge is to not crash when parsing these files

src/serialize.rs
richardkiss: How about a `max_atom_size` parameter?

arvidn: That seems like a greater risk, and I don't see what its upside is.

* If that parameter changes, the consensus rules change. We really should have unit tests to ensure all the consensus rules are what they are, that fail if any such rule change. With making parts of the rules parameterized, we invite it changing without having a unit test fail.
* There is no place where this function is called from where we might want to set different max atom size for the parser.
* If we have a max atom size (which I think is the safer option), it would make sense to not only make it apply to the parser, but also to new allocations.

tests/run-programs.py
```diff
@@ -1,12 +1,42 @@
...
+    if k not in "qa."
+)
+
+for fn in glob.glob('programs/large-atom-*.hex'):
+
+    try:
+        program_data = bytes.fromhex(open(fn, 'r').read())
+        max_cost = 100000
+
+        cost, result = deserialize_and_run_program(
```
richardkiss, line 25: Any chance we can isolate this ugly mess with the `native_opcode_names_by_opcode` and invocation of `deserialize_and_run_program` to one place in `tests`? I plan to refactor & simplify it soon anyway.

richardkiss (approved):

tests/run-programs.py
arvidn: well, I think it would make sense to bury it in the rust code, that would achieve the de-duplication

PR #51, opened 2021-02-23
Max atom size

arvidn:
This introduces a maximum size of atoms.

If atom sizes are unbounded, there would be a lot of places that would need additional overflow checks, anywhere we perform cost computations based on atom size for instance.

For example:
* https://github.com/Chia-Network/clvm_rs/blob/main/src/more_ops.rs#L369
* https://github.com/Chia-Network/clvm_rs/blob/main/src/more_ops.rs#L500

One risk of such approach is that the effective maximum atom size would be implied by which operations are performed and how costs are computed for those operations. For example, using multi-precision integers would avoid overflow of intermediate values and could raise the effective max atom size.

Proving that we could not overflow anywhere would be difficult and prone to errors.

Specifying an atom size upper limit explicitly has the benefit of:

1. simplify all logic that computes costs based on atom sizes, to know that all values are within reason and won't overflow
2. simplifies the documentation by simply stating the limit, rather than having to document individual limits for different operations.

### DIFF
--- a/src/cost.rs
+++ b/src/cost.rs
@@ -1,6 +1,10 @@
 use crate::allocator::Allocator;
 use crate::reduction::EvalErr;
 
+// this must be small enough to prevent overflow in intermediate values in cost
+// computations.
+pub const MAX_ATOM_SIZE: usize = 8 * 1024 * 1024;
+
 pub type Cost = u64;
 
 pub fn check_cost<A: Allocator>(a: &A, cost: Cost, max_cost: Cost) -> Result<(), EvalErr<A::Ptr>> {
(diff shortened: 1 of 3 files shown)


richardkiss:
This one bothers me a little. I was still really hoping that we shoehorn all our limits into costs, even if it means charging artificially high amounts for allocation (ie. wherever there is a `new_atom` call, add a premium proportional to bytes) and not just for computation. Do we have examples that prove this is necessary?

arvidn:
I think proving that this *isn't* necessary is non-trivial. And I have a hard time justifying producing such evidence. There doesn't seem to be any practical upside of defining this limit implicitly via costs.

It seem like writing security critical code without having buffer overflow checks, but instead construct the code such that overflows shouldn't happen.

arvidn:
that said, the limit should probably be greater than 1 MiB still.

PR #52, opened 2021-02-23
Ban leading zeros

arvidn:
the rust counterpart to https://github.com/Chia-Network/clvm/pull/71

Note that this contains a commit just to pull the updated tests from the `clvm` repo as well, which should be dropped before landing.

### DIFF
--- a/src/more_ops.rs
+++ b/src/more_ops.rs
@@ -12,7 +12,7 @@ use crate::cost::{check_cost, Cost};
 use crate::err_utils::{err, u8_err};
 use crate::node::Node;
 use crate::number::{number_from_u8, ptr_from_number, Number};
-use crate::op_utils::{atom, check_arg_count, int_atom, two_ints, uint_int};
+use crate::op_utils::{atom, check_arg_count, i32_atom, int_atom, two_ints, u32_from_u8};
 use crate::reduction::{Reduction, Response};
 use crate::serialize::node_to_bytes;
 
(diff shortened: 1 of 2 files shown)


src/more_ops.rs
```diff
@@ -73,53 +73,6 @@ fn limbs_for_int(v: &Number) -> usize {
     ((v.bits() + 7) / 8) as usize
 }
 
-fn u32_from_u8(mut buf: &[u8]) -> Option<u32> {
```
arvidn, line 76: this function (along with its tests) was moved into `op_utils.rs` as it seemed like a more appropriate place now that it's used in more places. It also changed semantics.

It was changed to return a *signed* int 32 and *reject* values that are too wide, even if there are leading zeros.

The rationale for not stripping leading zeros is that it would have a cost doing so, which means we would have to charge for that cost. By instead just rejecting such arguments, we don't need to charge a cost, since it's fixed at 4 bytes (or fewer)

src/more_ops.rs
```diff
@@ -217,9 +170,13 @@ pub fn op_unknown<A: Allocator>(
 
     assert!(cost > 0);
 
+    check_cost(allocator, cost, max_cost)?;
     cost *= cost_multiplier + 1;
-
-    Ok(Reduction(cost as Cost, allocator.null()))
+    if cost > u32::MAX as u64 {
```
arvidn, line 175: this check was missing previously, it was an oversight and exposed by one of the new test cases (as a discrepancy with the python implementation)

src/more_ops.rs
```diff
@@ -532,18 +485,21 @@ pub fn op_ash<T: Allocator>(a: &mut T, input: T::Ptr, _max_cost: Cost) -> Respon
...
-    let (i0, l0, i1, _) = uint_int(&args, "lsh")?;
-    let s1 = i64::try_from(&i1);
-    if match s1 {
-        Err(_) => true,
-        Ok(v) => v.abs() > 65535,
-    } {
+    check_arg_count(&args, 2, "lsh")?;
+    let a0 = args.first()?;
+    let b0 = int_atom(&a0, "lsh")?;
+    let i0 = BigUint::from_bytes_be(b0);
```
richardkiss, line 491: This is kind of an annoyance, as I think it's the only place in an operator impl that exposes which bigint library we're using underneath. Moving into `Number` might make gmp port less of a headache. But I guess we can burn that bridge when we get to it. [sic]

richardkiss (approved):
It passes the tests and seems good to me!

src/more_ops.rs
arvidn: yes. In fact, since a shift is a pretty simple operation, we might even get away with implementing it ourselves, over the byte-buffer. But I think we can punt on that

PR #54, opened 2021-02-25
Disable nested cons eval

richardkiss:


richardkiss:
A couple more tests in `clvm:develop` break without this change.

richardkiss:
Oh shoot, it breaks the benchmarks. @arvidn I can help you with these later if you need it.

richardkiss:
Basically, replace `((c PROG ARGS))` with `(a PROG ARGS)` and `(a ATOM ARGS)` with `(ATOM ARG1 ARG2 ...)` (quoting `ARGn` individually if necessary).

richardkiss:
Probably easiest just to recompile everything.

arvidn (approved):

PR #55, opened 2021-02-25
recompile benchmarks to use the new apply operator

arvidn:
~~large-block is precompiled and still doesn't work~~


arvidn:
oh, this isn't quite done yet

benchmark/point-pow.clvm
```diff
@@ -5,4 +5,4 @@
 ;  (point_power point pow)
 ;)
 
-((c (q . ((c 2 (c 2 (c 5 (c 11 (q . ()))))))) (c (q . ((c (i (= 11 (q . 1)) (q . 5) (q . (point_add 5 ((c 2 (c 2 (c 5 (c (- 11 (q . 1)) (q . ()))))))))) 1))) 1)))
+(a (q 2 2 (c 2 (c 5 (c 11 (q))))) (c (q 2 (i (= 11 (q . 1)) (q . 5) (q 29 5 (a 2 (c 2 (c 5 (c (- 11 (q . 1)) (q))))))) 1) 1))
```
richardkiss, line 8: The efficient way to do point power (which is really multiplication, since it's repeated addition) is the standard multiplication(/exponentiation) algorithm. Start with the identity (0) in an accumulator. If the integer is odd, add the point to the accumulator, halve the integer, double the point, repeat until the integer is 0.

richardkiss (approved):

richardkiss:
Tests still fail, but go ahead and merge once you fix them (or let me know if you need another review).

PR #59, opened 2021-03-03
Add double-cons `((+) 1 2 3)` => `6` syntax.

richardkiss:
Related to https://github.com/Chia-Network/clvm/pull/75

Once that one is merged, we can run the tests here again to make sure the new test cases pass again.

### DIFF
--- a/src/run_program.rs
+++ b/src/run_program.rs
@@ -244,8 +244,17 @@ where
         };
 
         let op_atom = match self.allocator.sexp(&op_node) {
-            SExp::Pair(_, _) => {
-                return Node::new(self.allocator, program.clone()).err("(()) eval disabled")
+            SExp::Pair(new_operator, must_be_nil) => {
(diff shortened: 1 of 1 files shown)


src/run_program.rs
```diff
@@ -244,8 +244,17 @@ where
         };
 
         let op_atom = match self.allocator.sexp(&op_node) {
-            SExp::Pair(_, _) => {
-                return Node::new(self.allocator, program.clone()).err("(()) eval disabled")
+            SExp::Pair(new_operator, must_be_nil) => {
+                if let SExp::Atom(_) = self.allocator.sexp(&new_operator) {
```
aqk, line 248: Is there an implicit check for is_listp(new_operator) that I'm not seeing?

richardkiss: Yeah, the `if let SExp::Atom(_) =` only succeeds if `new_operator` represents an atom.

aqk (approved):

PR #60, opened 2021-03-09
New costs

arvidn:


arvidn:
the top commit should be removed before landing this

src/core_ops.rs
```diff
@@ -4,12 +4,12 @@ use crate::node::Node;
 use crate::op_utils::{atom, check_arg_count};
 use crate::reduction::{Reduction, Response};
 
-const FIRST_COST: Cost = 8;
-const IF_COST: Cost = 31;
-const CONS_COST: Cost = 18;
-const REST_COST: Cost = 20;
-const LISTP_COST: Cost = 5;
-const CMP_BASE_COST: Cost = 16;
+const FIRST_COST: Cost = 1;
```
aqk, line 7: Well, that's dramatic!

src/more_ops.rs
```diff
@@ -18,56 +18,56 @@ use crate::serialize::node_to_bytes;
...
+const SHIFT_COST_PER_BYTE_DIVIDER: Cost = 32;
 
 const BOOL_BASE_COST: Cost = 1;
-const BOOL_COST_PER_ARG: Cost = 8;
+const BOOL_COST_PER_ARG: Cost = 3;
 
-const POINT_ADD_BASE_COST: Cost = 213;
-const POINT_ADD_COST_PER_ARG: Cost = 358;
+const POINT_ADD_BASE_COST: Cost = 316;
+const POINT_ADD_COST_PER_ARG: Cost = 4200;
```
aqk, line 67: These increases must be because we are using the Rust library, instead of our own Relic-based BLS-128.

src/more_ops.rs
```diff
@@ -218,7 +218,7 @@ fn test_unknown_op_reserved() {
 
     // a single ff is not sufficient to be treated as a reserved opcode
     let buf = vec![0xff];
-    assert_eq!(test_op_unknown(&buf, &mut a, null), Ok(Reduction(4, null)));
+    assert_eq!(test_op_unknown(&buf, &mut a, null), Ok(Reduction(1, null)));
```
aqk, line 221: In `Reduction(1,null)`, what is `1`?

aqk (approved):

src/more_ops.rs
arvidn: but it's a good point, long term we might want to use the same bls implementation

src/more_ops.rs
arvidn: that's the cost

richardkiss:
I haven't run my own tests of course, but no obvious issues, so looks good to me!

src/more_ops.rs
mariano54: Can you explain what these operations are? Which one is used for the AGG_SIG condition (which requires verifying a pairing operation). 

From what I've seen in our bls library, verifying a single pairing takes 3ms, but when batch verifying it takes 1ms (so we should use the batch verification cost). Does this kind-of align with the costs in this file?

richardkiss: Although clvm generates `(AGG_SIG ...)` and `(AGG_SIG_ME ...)` conditions as output, the actual pairing is deferred to the condition checker, which is written in python, and adds costs proportionate to the number of these conditions. So none of these costs in this PR correspond to pairing.

Instead, look here: https://github.com/Chia-Network/chia-blockchain/blob/main/src/consensus/cost_calculator.py#L38

richardkiss: Probably @arvidn should benchmark the ops here to make sure their costs are normalized to the same scale as his other benchmarks.

arvidn: the new costs represents approximately 0.1 microsecond of run-time per 1 cost, on my MacBook Pro.
If the `AGG_SIG` operation takes 1 ms (amortized) on my MacBook, it would correspond to roughly 10000 in cost.

However, looking at the link richard posted, it looks like we charge cost per item in the `cvp_list`. Is each such item its own operation, which should be charged the cost of `AGG_SIG`?

mariano54: Yes it looks like it's set at 20 cost here.
https://github.com/Chia-Network/chia-blockchain/blob/bb06f0b8d49b3525b1d91219615c769896c93d7a/src/consensus/condition_costs.py#L4

There is a 108 multiplier to compare to clvm cost: 
https://github.com/Chia-Network/chia-blockchain/blob/dev/src/consensus/default_constants.py#L54

That would imply a cost of 2100 which is probably a little bit on the cheap side. 

Also, not sure why we have two types of costs, wouldn't it be simpler to use the same units everywhere?

richardkiss: My comment above wasn't very clear, so I'll rewrite.

As we know, "running" a block generates a list of coins consumed and their corresponding condition outputs. Most of these conditions can be checked extremely quickly ("does block timestamp exceed X?"), so provide no DOS vectors. But `AGG_SIG` and `AGG_SIG_ME` invoke slow bls operations, so when those conditions are present, we have to add some cost to the block program to ensure creating a zillion of these condition outputs isn't a DOS vector.

Someone should benchmark these bls operations to ensure the cost we charge for `AGG_SIG*` is correctly proportional to all the other costs.

PR #63, opened 2021-03-12
extend cost stress tests

arvidn:
update max cost to 40000000 and make the tests larger and more expensive.

While adding tests that attempt to explode heap usage, I encountered a few stress tests that would take a lot longer than 4 seconds before failing. This turned out to be caused by the cost-per-byte not being applied against the cost until after the whole operation completes.

To ensure invalid programs fail quickly, we have to account for the per-byte cost as we go through the arguments loop as well. Since we may have rounding errors, we can't apply it by adding it to `cost` each turn through the loop, but it will always be an underestimate, since integer division rounds down.

Some of these large tests were enabled by generating a hex program directly (see `many_args()`), since `opc` isn't able to compile programs where a single call has many arguments.

I also changed the test to use the low-level clvm call since running the programs through `brun` would take too long.


tests/generate-programs.py
```diff
@@ -23,6 +24,22 @@ def many_args(filename, op, num):
...
+    else:
+        hexop = KEYWORD_TO_ATOM[op].hex()
+
+    with open(filename[:-4] + 'hex', 'w+') as f:
+        f.write('ff02ffff01ff02ff06ffff04ff02ffff04ffff02ff04ffff04ff02ffff04ff05ffff'
+            '0180808080ffff0180808080ffff04ffff01ffff02ffff03ff05ffff01ff17ffff02'
+            'ff04ffff04ff02ffff04ffff11ff05ffff010180ffff0180808080ffff018300ffff'
+            '80ffff01ff01818080ff0180ff' + hexop +
+            ('ff05' * num) + '80ff018080')
+
```
richardkiss, line 42: What does this do? Is there some way to make it more literate (or at least, comments)?

richardkiss (approved):
I don't understand all the new stuff, but it seems good.

tests/generate-programs.py
arvidn: it's the hex version of the clvm and chialisp code above. It's just that `run` can't compile it because it stack-overflows

aqk: Can you add a comment about the run stack overflow?

aqk (approved):

src/more_ops.rs
```diff
@@ -134,7 +134,11 @@ pub fn op_unknown<A: Allocator>(
                 cost += ARITH_COST_PER_ARG as u64;
                 let blob = int_atom(&arg, "unknown op")?;
                 byte_count += blob.len() as u64;
-                check_cost(allocator, cost, max_cost)?;
+                check_cost(
+                    allocator,
+                    cost + byte_count as Cost / ARITH_COST_PER_LIMB_DIVIDER,
```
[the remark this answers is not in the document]
arvidn: it is integer, yes. I'm not sure why you ask about `(cost + byte_count)`, that's not the value being divided. But since we haven't added up all the bytes yet, this is an under estimate of the total cost. But as an under estimate, it's still useful as an early abort

tests/run-programs.py
```diff
@@ -44,19 +48,68 @@
...
+            size = int(l.split('maxresident)k')[0].split(' ')[-1]) / 1024
+    if size != None:
+        print(Fore.YELLOW + ('  Resident Size: %d MiB' % size) + Style.RESET_ALL)
+
+        # TODO: tune costs to lower this a bit
+        if size > 1250:
+            ret += 1
+            print(Fore.RED + '\nTEST FAILURE: Max memory use exceeded\n' + Style.RESET_ALL)
+
+    # cost 40000000 roughly corresponds to 4 seconds
```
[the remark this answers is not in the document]
arvidn: well, it's not really meant to be relevant. I benchmarked this on an intel MacBook Pro, and it varies some, but not orders of magnitude. That's what "roughly" is supposed to mean

tests/generate-programs.py
arvidn: it's fairly well known: https://github.com/Chia-Network/clvm_tools/blob/develop/clvm_tools/cmds.py#L52

PR #65, opened 2021-04-01
Two-parameter substr.

richardkiss:


richardkiss:
@arvidn @aqk  I hacked this together. Had to factor out a function. It may be a little gross, so comments on how to make nicer encouraged.

richardkiss:
DO NOT MERGE. Last commit has to be removed prior to merge.

src/op_utils.rs
```diff
@@ -6,33 +6,34 @@ use crate::reduction::EvalErr;
...
-                        "{} takes exactly {} argument{}",
-                        name,
-                        expected,
-                        if expected == 1 { "" } else { "s" }
-                    ))
-                }
-            }
-        }
-        cnt -= 1;
+    while let Some((_, next)) = ptr.pair() {
```
arvidn, line 29: I'm unfamiliar with this. If `pair()` returns `None` so it doesn't match the `Some()`, does that make the whole `let Some((_, next)) = ...` expression itself be false? So is this essentially a short-hand for a match expression?

arvidn (approved):
looks reasonable

src/op_utils.rs
richardkiss: I only know about it because `cargo clippy` suggested it! And it worked, so I left it. It used to be `loop` with `if let` (and `if let` is a shorthand match-expression).

src/op_utils.rs
```diff
@@ -6,33 +6,34 @@ use crate::reduction::EvalErr;
...
+            name,
+            expected,
+            if expected == 1 { "" } else { "s" }
+        ))
+    } else {
+        Ok(())
+    }
+}
+
+pub fn arg_count<T: Allocator>(args: &Node<T>, return_early_if_exceeds: usize) -> usize {
```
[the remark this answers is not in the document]
richardkiss: Okay, I added some.

richardkiss:
I'm merging even though ubuntu tests are broken, as they're broken in main too, and it seems to be unrelated to this change.

PR #66, opened 2021-04-02
Various fixes to `podman` build.

richardkiss:


.github/workflows/build-test.yml
```diff
@@ -81,11 +81,12 @@ jobs:
             else export SCND_VERSION="$PY_VERSION"; fi && \
             echo "Exporting path /opt/python/cp$PY_VERSION-cp$SCND_VERSION/bin" && \
             export PATH=/opt/python/cp$PY_VERSION-cp$SCND_VERSION/bin/:$PATH && \
-            /opt/python/cp38-cp38/bin/python -m venv venv && \
+            /opt/python/cp$PY_VERSION-cp$SCND_VERSION/bin/python -m venv venv && \
```
hoffmang9, line 84: There was a reason I chose to use only 3.8 here but I can't remember what that was right now.

hoffmang9 (approved):
+1

.github/workflows/build-test.yml
richardkiss: Let me undo this change, I don't believe it's necessary anymore.

richardkiss: It worked, I'll squash & merge.

PR #67, opened 2021-04-12
increase cost granularity by 100

arvidn:
Also tweak a bit for Raspberry PI

### DIFF
--- a/src/core_ops.rs
+++ b/src/core_ops.rs
@@ -4,13 +4,16 @@ use crate::node::Node;
 use crate::op_utils::{atom, check_arg_count};
 use crate::reduction::{Reduction, Response};
 
-const FIRST_COST: Cost = 1;
-const IF_COST: Cost = 1;
-const CONS_COST: Cost = 3;
-const REST_COST: Cost = 1;
-const LISTP_COST: Cost = 1;
-const CMP_BASE_COST: Cost = 2;
-const CMP_COST_PER_BYTE_DIVIDER: Cost = 64;
+const FIRST_COST: Cost = 30;
+const IF_COST: Cost = 33;
+// Cons cost lowered from 245. It only allocates a pair, which is small
+const CONS_COST: Cost = 50;
+// Rest cost lowered from 77 since it doesn't allocate anything and it should be
+// the same as first
+const REST_COST: Cost = 30;
+const LISTP_COST: Cost = 19;
+const EQ_BASE_COST: Cost = 117;
+const EQ_COST_PER_BYTE: Cost = 1;
 
 pub fn op_if<T: Allocator>(a: &mut T, input: T::Ptr, _max_cost: Cost) -> Response<T::Ptr> {
     let args = Node::new(a, input);
@@ -67,6 +70,6 @@ pub fn op_eq<T: Allocator>(a: &mut T, input: T::Ptr, _max_cost: Cost) -> Respons
     let a1 = args.rest()?.first()?;
     let s0 = atom(&a0, "=")?;
     let s1 = atom(&a1, "=")?;
-    let cost = CMP_BASE_COST + (s0.len() as Cost + s1.len() as Cost) / CMP_COST_PER_BYTE_DIVIDER;
+    let cost = EQ_BASE_COST + (s0.len() as Cost + s1.len() as Cost) * EQ_COST_PER_BYTE;
     Ok(Reduction(cost, if s0 == s1 { a.one() } else { a.null() }))
 }
--- a/src/more_ops.rs
+++ b/src/more_ops.rs
@@ -24,59 +24,85 @@ use sha2::{Digest, Sha256};
 #[cfg(unix)]
 use openssl::sha;
 
-const ARITH_BASE_COST: Cost = 1;
-const ARITH_COST_PER_ARG: Cost = 3;
-const ARITH_COST_PER_BYTE_DIVIDER: Cost = 16;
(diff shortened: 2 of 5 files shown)


src/core_ops.rs
```diff
@@ -4,13 +4,17 @@ use crate::node::Node;
...
-const LISTP_COST: Cost = 1;
-const CMP_BASE_COST: Cost = 2;
-const CMP_COST_PER_BYTE_DIVIDER: Cost = 64;
+const FIRST_COST: Cost = 30;
+// If cost lowered from 125 since it does not allocate anything
+const IF_COST: Cost = 75;
+// Cons cost lowered from 245. It only allocates a pair, which is small
+const CONS_COST: Cost = 80;
+// Rest cost lowered from 123 since it doesn't allocate anything
+const REST_COST: Cost = 65;
```
richardkiss, line 13: How could first and rest possibly be different?

src/more_ops.rs
```diff
@@ -24,56 +24,74 @@ use sha2::{Digest, Sha256};
...
+const STRLEN_BASE_COST: Cost = 219;
+const STRLEN_COST_PER_BYTE: Cost = 1;
 
-const CONCAT_BASE_COST: Cost = 1;
-const CONCAT_COST_PER_ARG: Cost = 1;
-const CONCAT_COST_PER_BYTE_DIVIDER: Cost = 16;
+const CONCAT_BASE_COST: Cost = 142;
+const CONCAT_COST_PER_ARG: Cost = 135;
+// increased from 3 to charge more for memory allocations
+const CONCAT_COST_PER_BYTE: Cost = 6;
```
richardkiss, line 57: I'd recommend that costs that are artificially cranked higher to prevent memory DOS attacks have a name with `ALLOCATED` or something in them, and maybe even a constant multiplier with like `ANTI_DOS_MALLOC_MULTIPLIER = 2`. Something like

```
const ANTI_DOS_MALLOC_MULTIPLIER: u32 = 2;
(...snip...)
const CONCAT_COST_PER_MALLOC_BYTE: Cost = 3 * ANTI_DOS_MALLOC_MULTIPLIER;

const MUL_LINEAR_COST_PER_MALLOC_BYTE: Cost = 3 * ANTI_DOS_MALLOC_MULTIPLIER;
```
This way we can crank up or down all anti-memory DOS vectors at once.

src/run_program.rs
```diff
@@ -6,10 +6,14 @@ use crate::reduction::{EvalErr, Reduction, Response};
 
 use crate::number::{ptr_from_number, Number};
 
-const QUOTE_COST: Cost = 1;
-const TRAVERSE_COST_PER_ZERO_BYTE: Cost = 1;
-const TRAVERSE_COST_PER_BIT: Cost = 1;
-const APPLY_COST: Cost = 1;
+const QUOTE_COST: Cost = 2;
+const APPLY_COST: Cost = 2;
```
richardkiss, line 10: There's no way `QUOTE` takes as long as `APPLY`, is there?

src/core_ops.rs
arvidn: I was asking myself the same thing when I edited this. I will take a deeper look at the benchmarks and results tomorrow

src/more_ops.rs
arvidn: that's a good point. Although, we won't be changing this for that many more days, but it makes for more understandable code

src/run_program.rs
arvidn: I don't have measurements for these. You're right that the apply base cost ought to be something non-trivial.

arvidn: I suppose I should try to measure the overhead of apply somehow

src/core_ops.rs
richardkiss: It could be cache hits are more likely to come from `f` than `r`, but that would just be an artifact based on how programs tend to be formed, and wouldn't affect worst case, which is what we care about.

src/more_ops.rs
richardkiss: True, we could delay that change until later as long as the constants' values don't change.

src/run_program.rs
richardkiss: Well, both quote and apply actually just update the internal stacks, but apply pushes three things and quote pushes just one (I think).

richardkiss: In any case, I expect both to be asymptotically zero, so the exactly values probably don't much matter.

src/more_ops.rs
arvidn: I don't think the malloc cost should be a multiplier though. If it's meant to charge cost for allocating memory, it doesn't really matter how expensive the operation was to *produce* the byte, we charge for the allocation of the byte.

src/core_ops.rs
arvidn: rerunning these tests, it turned out to just be noise. I'll set them both to the same cost and probably lower it a bit compared to the measurement too

src/more_ops.rs
arvidn: part of me feels like this really should be a cost applied to `new_atom()`, based on the number of bytes. Right now, `lognot` allocates memory (as much as its input atom), but it pays less of the "malloc" cost compared to `logxor`, since the latter is binary and pays the malloc cost for both its operands.

It might be a bit late to make a change like this though.

richardkiss (approved):

PR #68, opened 2021-04-13
add stress test with an very long list

arvidn:
with the same repeated atom to not use a lot of memory

### DIFF
--- a/tests/generate-programs.py
+++ b/tests/generate-programs.py
@@ -3,6 +3,22 @@
 import os
 from clvm import KEYWORD_TO_ATOM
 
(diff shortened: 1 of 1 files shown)


arvidn:
I added this stress test to ensure it's OK to lower the cost ascribed to `c` without making a DOS attempt viable. As demonstrated [here](https://github.com/Chia-Network/clvm_rs/pull/68/checks?check_run_id=2335183818#step:8:77), it's fine now and would fail if me next batch breaks that.

aqk (approved):

PR #69, opened 2021-04-14
Release GIL in `deserialize_and_run_program`.

richardkiss:
This should allow use to use `ThreadPoolExecutor ` instead of `ProcessPoolExecutor` on `deserialize_and_run_program`.


arvidn (approved):

mariano54:
Nice. Unfortunately we already went pretty deep into multiprocessing with block validation, due to other stuff also blocking the GIL (particularly vdf validation). But it would be nice to move to threads eventually

PR #73, opened 2021-05-01
clippy fixes

sielicki:
apologies for horrible pedantic drive-by nature of this PR, but it's really just fixing stuff that clippy complained about on nightly.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -170,9 +170,9 @@ jobs:
         TWINE_PASSWORD: ${{ secrets.test_pypi_password }}
       run: twine upload --non-interactive --skip-existing --verbose 'target/wheels/*'
 
-  checks:
+  fmt:
     runs-on: ubuntu-20.04
-    name: Checks
+    name: cargo fmt
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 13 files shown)


arvidn:
This looks reasonable, but I think it would only make sense together with updating the github workflow to also use nightly clippy.

sielicki:
Do you want me to do that, or do we like being on stable?

arvidn:
I don't have a strong opinion. I take it "nightly" clippy is still reasonably stable in its checks. I just think the source code should stay in sync with CI.

@richardkiss what do you think?

richardkiss:
This is fine and all, since eventually all these complains will come to stable rust, and these will be real errors. But what would make this PR really great would be a change that updates the clippy checks to use nightly rust so that new problems are created. What do you say @sielicki? (If you're not familiar with CI – "continuous integration" – start with the scripts in `.github`.)

sielicki:
Sure -- I haven't worked with github actions specifically but I can give it a shot. 

Do you mean just rewriting the `check` [here](https://github.com/Chia-Network/clvm_rs/blob/main/.github/workflows/build-test.yml#L173-L189) to use the [clippy specific action from actions-rs](https://github.com/actions-rs/clippy-check)?

richardkiss:
Sorry about the prior comment, I hadn't read the entire thread yet. These changes look good, thank you. I'll run the work-flow to make sure all tests still pass.

richardkiss (approved):

PR #74, opened 2021-05-01
add a cargo config to help macos with linkage

sielicki:
This lets you run `cargo build` on macos without having to go through the whole hoopla of maturin.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,11 @@
+[target.x86_64-apple-darwin]
+rustflags = [
+  "-C", "link-arg=-undefined",
+  "-C", "link-arg=dynamic_lookup",
+]
+
+[target.aarch64-apple-darwin]
+rustflags = [
+  "-C", "link-arg=-undefined",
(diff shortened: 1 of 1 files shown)


arvidn:
is the MacOS specific? afaik, you can't do `cargo build` on linux either.

With this patch, do you get a python module from `cargo build`?

sielicki:
I came across this on the pyo3 docs [here](https://pyo3.rs/v0.13.2/#using-rust-from-python). My understanding is that this is just a hack required on macs and that `cargo build` does succeed on linux, but I haven't tried. This shouldn't have any impact on the behavior on other platforms (because of the `[target.<arch>-apple-darwin]`)

What you get out from running `cargo build` is a file `target/debug/libclvm_rs.dylib`, and if you make a symlink named `clvm_rs.so` that points to that, you can pop open a python shell and do `import clvm_rs`. You still probably will want to use maturin for everything -- I'm just used to working with `watch cargo build -vv` running in a terminal for quick feedback.

richardkiss (approved):

PR #75, opened 2021-05-01
add the rust gitignore from github/gitignore

sielicki:


sielicki:
if you're concerned about licensing, it came from here: https://raw.githubusercontent.com/github/gitignore/master/Rust.gitignore

https://github.com/github/gitignore/blob/master/LICENSE

arvidn (dismissed):

sielicki:
I just realized that this did not purge files which were checked in, but are now included in `.gitignore`. I just force pushed an amended commit which also purges Cargo.lock.

richardkiss:
We've had problems deleting `Cargo.lock` in the past when a dependent library played loosie-goosie with its own dependencies, so I'm scared to merge. I don't remember the exact details now, and it's likely difficult to reproduce since a lot of the state is just "what's the most recent version of library X", which changes over time.

If we do delete `Cargo.lock` and the build breaks in the future, what's the strategy for fixing it?

sielicki:
This PR is very simple to recreate so I'm going to close in the interim.

PR #76, opened 2021-05-01
run_program: ensure safe failure when stack is small

sielicki:
In the case that only one element is present and a swap or combine is
attempted, an error will raise on the second pop(), and the contents of the
first pop() will be silently dropped. Rewrite these ops such that they attempt
to put contents back on the stack before returning an error.

Signed-off-by: Nicky Sielicki <sielicki@yandex.com>

### DIFF
--- a/src/run_program.rs
+++ b/src/run_program.rs
@@ -177,21 +177,61 @@ impl<'a, 'h, T: Allocator> RunProgramContext<'a, T> {
     }
 
     fn swap_op(&mut self) -> Result<Cost, EvalErr<T::Ptr>> {
-        /* Swap the top two operands. */
-        let v2 = self.pop()?;
-        let v1 = self.pop()?;
-        self.push(v2);
-        self.push(v1);
-        Ok(0)
+        /* swap the top two operands. */
+        match self.pop() {
+            Ok(v1) => {
+                match self.pop() {
+                    Ok(v2) => {
+                        // Push back on, reversing order.
+                        self.push(v1);
+                        self.push(v2);
+                        Ok(0)
+                    }
+                    Err(e) => {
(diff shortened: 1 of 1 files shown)


sielicki:
I really don't like that I had to clone in the cons fix. Maybe we can rewrite the `new_pair` signature from:

```fn new_pair( &mut self, first: Self::Ptr, rest: Self::Ptr,) -> Result<Self::Ptr, EvalErr<Self::Ptr>>```

To something like

```fn new_pair( &mut self, first: Self::Ptr, rest: Self::Ptr,) -> Result<Self::Ptr, EvalErr<(Self::Ptr, Self::Ptr)>>```

such that we get ownership of both parameters back in the case of an error.

arvidn:
we have plans on dropping the `ArcAllocator`, which would simplify the ownerships of the Allocator trait. Then the `Ptr` type will implement `Copy` and be a primitive type.

Regarding this patch, could you add unit tests that exercise the problem you're solving?

sielicki:
I would be very excited to see `ArcAllocator` be dropped -- my whole goal yesterday was to derive [Arbitrary](https://docs.rs/arbitrary/1.0.0/arbitrary/) on structures throughout the project so that I could implement `cargo fuzz` support, but I got lost in lifetime hell before I could make any substantial progress.

arvidn:
the `ArcAllocator` has been removed now, and some owner relationships have been simplified. There are still things that can be simplified now, but I think this is enough to not have to clone node pointers (previously `Allocator::Ptr`, now called `NodePtr`).

sielicki:
Closing due to losing context on what this was trying to accomplish in the first place. Sorry!

PR #79, opened 2021-05-20
Use `&[u8]` for opcode instead of `u8`.

richardkiss:


arvidn:
this PR can be closed, right? this patch has already landed in `main`

PR #80, opened 2021-05-20
Fix bitrot in build.

richardkiss:


.github/workflows/benchmark.yml
```diff
@@ -116,7 +116,7 @@ jobs:
         python -m venv venv
         ln -s venv/bin/activate
         . ./activate
-        git clone https://github.com/Chia-Network/clvm_tools.git --branch=develop --single-branch
+        git clone https://github.com/Chia-Network/clvm_tools.git --branch=main --single-branch
```
arvidn, line 119: it might make sense to just drop the `--branch=` argument too. `main` is the default branch iiuc.

arvidn (approved):

PR #83, opened 2021-06-11
Drop ArcAllocator and simplify templates

arvidn:
I renamed `IntAllocator` to just `Allocator` and moved the file to `src/allocator.rs`, which turns into a large diff, because the original `src/allocator.rs` trait was removed.


richardkiss:
This need rebasing now, but also I'd like to look at deferring this until after my `Dialect` changes go in, since there may be additional conflict.

arvidn:
do you use any of this in that patch?

richardkiss (approved):

PR #87, opened 2021-06-25
Simplify allocator

arvidn:
With the `ArcAllocator` removed, ownership is implified and we can pass around `NodePtr` instead of `AtomBuf`. this simplifies a few things.

Additionally, since `NodePtr` is an `i32`, we shouldn't be passing it around by reference, but just copy it instead.

Now that we only have a single allocator, `SExp` and `EvalErr` no longer needs to be templates.

These are mostly independent commits, sharing the common dependency that they are simplifications of not having the allocator be a template anymore.

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -65,7 +65,7 @@ impl Allocator {
         r
     }
 
-    pub fn new_atom(&mut self, v: &[u8]) -> Result<NodePtr, EvalErr<NodePtr>> {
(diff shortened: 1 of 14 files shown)


richardkiss (approved):
Seems like a good change. Lots of simplifications. I didn't look at every line closely, just sort of tried to understand the general ideas, but I know we have good test coverage so I think it's good to merge.

PR #91, opened 2021-06-29
implement concat without intermediate buffer

arvidn:
instead of accumulating the concatenated string into a temporary buffer, and then copy it into the clvm heap, this patch copies each buffer directly from its current location. This is made possible by `Vec`s `extend_from_within()`.

I was hoping that this would have a material impact on the worst case timing of the concat benchmark, but it didn't. Comparing a few runs it seems to consistently improve the run-time however.

```
concat.hex mean: 0.602707 (-0.010977) -1.79 %
concat.hex mean: 0.602506 (-0.011178) -1.82 %
concat.hex mean: 0.602707 (-0.010664) -1.74 %
concat.hex mean: 0.602506 (-0.010865) -1.77 %
```
(this is on an M1, so I would expect the impact might be more drastic on an RPi for instance)

The greater impact is on the resident size. The `recursive-concat` test specifically.

current main:

```
/usr/bin/time -l ./run.py programs/recursive-cat.hex programs/recursive-cat.envhex
  Runtime: 0.23 s
  Resident Size: 984 MiB

/usr/bin/time -l ./run.py programs/recursive-cat.hex programs/recursive-cat.envhex
  Runtime: 0.23 s
  Resident Size: 1176 MiB
```

With this patch:

```
/usr/bin/time -l ./run.py programs/recursive-cat.hex programs/recursive-cat.envhex                                                                                                                              
  Runtime: 0.17 s                                                               
  Resident Size: 600 MiB

/usr/bin/time -l ./run.py programs/recursive-cat.hex programs/recursive-cat.envhex                                                                                                                                                            
  Runtime: 0.16 s
  Resident Size: 600 MiB
```

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -15,8 +15,11 @@ pub struct AtomBuf {
 }
 
 impl AtomBuf {
-    pub fn is_empty(self) -> bool {
+    pub fn is_empty(&self) -> bool {
         self.start == self.end
+    }
+    pub fn len(&self) -> usize {
+        (self.end - self.start) as usize
     }
 }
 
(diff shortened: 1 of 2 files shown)


arvidn:
`RPi` is about the same

richardkiss (approved):
This seems fine to merge. From what you've written, it seems like it improves things a little bit, at not much complexity cost. Go ahead and do so if you think it's worthwhile.

src/allocator.rs
```diff
@@ -15,9 +15,12 @@ pub struct AtomBuf {
 }
 
 impl AtomBuf {
-    pub fn is_empty(self) -> bool {
+    pub fn is_empty(&self) -> bool {
         self.start == self.end
     }
+    pub fn len(&self) -> usize {
```
richardkiss, line 21: You might consider exposing this on `LazyNode` with `.atom_length` which returns `int` for an atom and `None` for a pair.

arvidn: yes, I have been considering that. I will probably propose some patches to take this short cut internally first, to not go through a `&[u8]` just to get the length.

arvidn:
I do think this is worth it. The RSS reduction is quite significant and one of the things I worry about a bit.

PR #93, opened 2021-07-02
Use `&[u8]` for opcode instead of `u8`.

richardkiss:


src/run_program.rs
```diff
@@ -191,7 +188,7 @@ where
     ) -> Result<Cost, EvalErr> {
         let op_atom = self.allocator.buf(op_buf);
         // special case check for quote
-        if op_atom.len() == 1 && op_atom[0] == self.quote_kw {
+        if op_atom.len() == 1 && op_atom == self.quote_kw {
```
arvidn, line 191: it looks like you meant to drop the `op_atom.len() == 1` condition here. I don't think there are any tests covering this now

src/run_program.rs
```diff
@@ -282,7 +279,7 @@ where
             return err(operator, "internal error");
         }
         let op_atom = self.allocator.atom(operator);
-        if op_atom.len() == 1 && op_atom[0] == self.apply_kw {
+        if op_atom.len() == 1 && op_atom == self.apply_kw {
```
arvidn, line 282: here too

richardkiss:
Got rid of the now unnecessary `len == 1` checks.

arvidn:
I imagine the new clippy warning is an issue in pyo3 itself

arvidn:
it's a bit mysterious that it only hows up at this one instance though. We use that macro in a bunch of other places.

arvidn (approved):
this patch looks good. but we should probably make sure we don't have clippy warnings in `main`

richardkiss:
Oh yeah, maybe upgrading `pyo3` version would fix it. `clippy` doesn't give me any warnings locally.

PR #95, opened 2021-07-07
parse conditions

arvidn:


src/allocator.rs
```diff
@@ -8,10 +8,11 @@ pub enum SExp {
     Pair(NodePtr, NodePtr),
 }
 
-#[derive(Clone, Copy)]
+// TODO: figure out how to integrate with RangeSet without exposing these fields
+#[derive(Clone, Copy, Debug)]
 pub struct AtomBuf {
-    start: u32,
-    end: u32,
+    pub start: u32,
```
richardkiss, line 14: Consider a "getter" that returns these two values rather than making them public and giving people the wrong idea that they're mutable.

src/gen/conditions.rs
```diff
@@ -0,0 +1,1568 @@
...
+            ErrorCode::AssertHeightAbsolute,
+            range_cache,
+            flags,
+        )?)),
+        _ => Err(ValidationErr(c, ErrorCode::InvalidConditionOpcode)),
+    }
+}
+
+#[derive(Debug)]
+pub struct Npc {
```
richardkiss, line 199: `NPC` was always a terrible name, and since this isn't quite exactly the same, this gives us the opportunity to rename it. Maybe `SpendConditionsSummary`?

src/gen/conditions.rs
```diff
@@ -0,0 +1,1568 @@
...
+        return Err(ValidationErr(spend, ErrorCode::DoubleSpend));
+    }
+
+    let mut npc = Npc {
+        coin_id,
+        puzzle_hash,
+        reserve_fee: 0,
+        height_relative: 0,
+        height_absolute: 0,
+        seconds_relative: 0,
```
richardkiss, line 277: Can `(ASSERT_RELATIVE_SECONDS 0)` ever fail? If block timestamps are not purely monotonic, it could fail, which would make this a soft fork (ie. `(ASSERT_RELATIVE_SECONDS 0)` is different from no `ASSERT_RELATIVE_SECONDS` of any kind.

I know in bitcoin block timestamps are not purely monotonically increasing. @mariano54

src/gen/conditions.rs
```diff
@@ -0,0 +1,1568 @@
...
+    }
+
+    // check all the assert announcements
+    // if there are no asserts, there is no need to hash all the announcements
+    if !ann.assert_coin.is_empty() {
+        let mut announcements = HashSet::<[u8; 32]>::new();
+
+        // TODO:
+        // in order to avoid hashing the coin ID multiple times, do it once and
+        // then copy the hasher state to add the message for each announcement
```
richardkiss, line 439: Nothing happens until 512 bits are passed to `.update`, so this optimization wouldn't do very much. See also https://github.com/in3rsha/sha256-animation

src/gen/sanitize_int.rs
```diff
@@ -0,0 +1,273 @@
...
+    let mut ret: usize = 0;
+    while ret < buf.len() && buf[ret] == 0 {
+        ret += 1;
+    }
+    ret as usize
+}
+
+// this *just* counts zeros. Note that not all zeros are redundant. the last
+// zero byte may be necessary. This has to be fixed up by the caller
+fn count_zeros(a: &Allocator, range_cache: &mut RangeSet, atom_range: &AtomBuf) -> usize {
```
richardkiss, line 16: Fun (but probably not that useful) discussion of `align_to` here: https://stackoverflow.com/questions/65367552/checking-a-vecu8-to-see-if-its-all-zero

src/op_utils.rs
```diff
@@ -170,6 +170,44 @@ fn test_i32_from_u8() {
     assert_eq!(i32_from_u8(&[0x7d, 0xcc, 0x55, 0x88, 0xf3]), None);
 }
 
+pub fn u64_from_bytes(buf: &[u8]) -> u64 {
```
richardkiss, line 173: This https://doc.rust-lang.org/std/primitive.u64.html#method.from_be_bytes almost works, but you'd have to copy to an 0 array first. Might be faster, who knows. Probably doesn't matter.

tests/run_gen.py
```diff
@@ -0,0 +1,93 @@
...
+from clvm_tools import binutils
+import sys
+from run import native_opcode_names_by_opcode
+
+def run_gen(fn):
+
+    # the generator ROM from:
+    # https://github.com/Chia-Network/chia-blockchain/blob/main/chia/wallet/puzzles/rom_bootstrap_generator.clvm.hex
+    program_data = bytes.fromhex(
+        "ff02ffff01ff02ff0cffff04ff02ffff04ffff02ff05ffff04ff08ffff04ff13" \
```
richardkiss, line 16: FYI, python doesn't actually require `\` here because it's in parens `(...)`.

src/allocator.rs
arvidn: Yeah, I think the main issue is that you can construct your own instances of this now, that's the main violation of the abstraction. I'll look into fixing this

src/gen/conditions.rs
arvidn: right, my assumption is that all `ASSERT_`(`SECONDS`|`HEIGHT`)`_`(`ABSOLUTE`|`RELATIVE`) are inherently true if the argument is <= 0

src/op_utils.rs
arvidn: yeah, I considered using this function, I kind of dismissed it as I realized it took 8 bytes unconditionally. But it might end up being less code to copy it

tests/run_gen.py
arvidn: heh, I've always been confused when exactly I need these.

src/allocator.rs
arvidn: I restored this abstraction at a slight expense in complexity in `sanitize_integer`. See this commit: https://github.com/Chia-Network/clvm_rs/pull/95/commits/ccd8d073b48c1152ab57506ea2e3e57219a5ae03

src/gen/conditions.rs
arvidn: fixed

src/gen/conditions.rs
arvidn: I removed these comments

tests/run_gen.py
arvidn: removed in https://github.com/Chia-Network/clvm_rs/pull/95/commits/3cfd2f9498648702c7bc1680ff153da60c9d15fe

src/gen/sanitize_int.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    debug_assert!(start <= range.1);
+
+    if ret > 0 {
+        range_cache.add(range.0, range.0 + ret as u32);
+    }
+    ret
+}
+
+pub fn sanitize_integer<'a>(
```
richardkiss, line 64: Maybe `sanitize_uint` to emphasize that it rejects negative numbers?

src/gen/sanitize_int.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    let atom_range = match a.sexp(n) {
+        SExp::Atom(b) => b,
+        _ => {
+            panic!("unreachable");
+        }
+    };
+
+    // strip redundant leading zeros
+    let mut i = count_zeros(a, range_cache, &atom_range);
```
richardkiss, line 93: Any reason this doesn't update the range cache when it counts them? It has it borrowed `&mut`, right? Does it just make the code too messy?

richardkiss: Oh I see, it doesn't have `atom_range` available.

src/gen/validation_error.rs
```diff
@@ -0,0 +1,99 @@
...
+    ReserveFeeConditionFailed,
+    DuplicateOutput,
+    DoubleSpend,
+    CostExceeded,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct ValidationErr(pub NodePtr, pub ErrorCode);
+
+pub fn checked_add(lhs: u64, rhs: u64, n: NodePtr, e: ErrorCode) -> Result<u64, ValidationErr> {
```
richardkiss, line 34: Something similar at https://doc.rust-lang.org/std/primitive.u64.html#method.checked_add

richardkiss: You could use that in conjunction with `.ok_or` https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or

src/int_to_bytes.rs
```diff
@@ -0,0 +1,25 @@
+pub fn u64_to_bytes(n: u64) -> Vec<u8> {
+    let mut buf = Vec::<u8>::new();
+    buf.extend_from_slice(&n.to_be_bytes());
```
richardkiss, line 3: Nitpicky, but you could just go `if n < 0 buf.push(0);`

src/py/run_generator.rs
```diff
@@ -0,0 +1,290 @@
...
+    }
+}
+
+impl IntoPy<PyObject> for PyBytesObj {
+    fn into_py(self, py: Python) -> PyObject {
+        PyBytes::new(py, &self.arr).into()
+    }
+}
+
+#[pyclass(subclass, unsendable)]
```
richardkiss, line 52: I wonder if these go away if we upgrade `pyo3` to 0.14.1. Might be worth trying.

richardkiss: I'm trying it. Use `cargo +nightly clippy` to run nightly clippy locally (might have to do something like `rustup update nightly` if you haven't yet).

richardkiss: Yep, that fixes it!

src/py/run_generator.rs
```diff
@@ -0,0 +1,290 @@
...
+#[derive(Clone)]
+pub struct PyConditionWithArgs {
+    #[pyo3(get)]
+    pub opcode: ConditionOpcode,
+    #[pyo3(get)]
+    pub vars: Vec<PyBytesObj>,
```
richardkiss, line 58: This is more efficient if you use a `PyTuple`, so the pyo3 glue doesn't recreate a list each time across the python/rust bridge.

src/py/run_generator.rs
```diff
@@ -0,0 +1,290 @@
...
+use crate::py::lazy_node::LazyNode;
+
+use std::collections::HashMap;
+use std::rc::Rc;
+
+use pyo3::prelude::*;
+use pyo3::types::{PyBytes, PyDict};
+
+#[derive(Clone)]
+pub struct PyBytesObj {
```
richardkiss, line 25: You can actually use a real `PyBytes` object here, and it becomes a native python object. There are some examples in `py/run_program.rs` you can look at.

richardkiss: I guess if you're only looking at each object once from python, it doesn't matter. But for frequently-used things, you pay a conversion price each time across the bridge unless it's already a python object.

src/py/run_generator.rs
richardkiss: Just realized this all runs with threading enabled, so there's no access to `Python: py`, which means you can't do this.

richardkiss: I was wrong, threading is re-enabled before you get here (although it probably shouldn't be).

src/py/run_generator.rs
```diff
@@ -0,0 +1,290 @@
...
+                Err(ValidationErr(_, c)) => {
+                    Ok((Some(c), 0_u64, Vec::<SpendConditionSummary>::new()))
+                }
+                Ok(spend_list) => Ok((None, cost, spend_list)),
+            }
+        },
+    );
+
+    let mut ret = Vec::<PyNPC>::new();
+    match r {
```
richardkiss, line 261: You could move all this stuff up to the threaded area since you only use `py` in one of the error cases (and in that case, you could reacquire the `py` lock without much concern, since it should happen only rarely).

src/py/run_generator.rs
```diff
@@ -0,0 +1,290 @@
...
+        ret.to_object(py)
+    }
+}
+
+// returns the cost of running the CLVM program along with the list of NPCs for
+// the generator/spend bundle. Each SpendConditionSummary is a coin spend along with its
+// conditions
+#[allow(clippy::too_many_arguments)]
+#[pyfunction]
+pub fn run_generator(
```
richardkiss, line 220: Could we simply call `run_program` from python and then feed the `LazyNode` into a simplified version of this function starting at line 251? The upside is we would have less redundant api. The downside is the python could in theory retain a careless reference to the lazy node, delaying destroying the allocator (but as the doctor says when you tell him "it hurts when I do this", "Well then, don't do that.").

src/gen/conditions.rs
```diff
@@ -0,0 +1,1643 @@
...
+pub struct SpendConditionSummary {
+    pub coin_id: Arc<[u8; 32]>,
+    pub puzzle_hash: NodePtr,
+    // conditions
+    // The sum of all reserve fee conditions
+    pub reserve_fee: u64,
+    // the highest height/time conditions (i.e. most strict)
+    pub height_relative: u32,
+    pub height_absolute: u32,
+    pub seconds_relative: u64,
```
richardkiss, line 208: I'm still very scared about the implicit `(ASSERT_SECONDS_AGE_EXCEEDS 0)` this code generates. How hard would it be to change `seconds_relative` to `Option<u64>`? In particular, if that assert can ever fail, this is a soft fork (and it could fail if block times are not guaranteed to be monotonic, or if the "exceeds" condition is `>` rather than `>=`).

Similar concerns for `height_relative`.

richardkiss: There has been discussion in `chia_network#consensus` about this, but I'm not 100% convinced. I would prefer @mariano54 weigh in also, or research the two questions yourself: (1) is "exceeds" non-strict, ie. >= 0 is okay; and (2) are blocks monotonic.

richardkiss: This seems to suggest transaction block timestamps are purely monotonic increasing https://github.com/Chia-Network/chia-blockchain/blob/main/chia/consensus/block_header_validation.py#L822

richardkiss: This seems to suggest we're okay with `>` vs `>=` https://github.com/Chia-Network/chia-blockchain/blob/7d55c39ab18609f47ce90106314244f445496947/chia/full_node/mempool_check_conditions.py#L98

richardkiss: This seems to suggest we're okay with `>` vs `>=` for block height https://github.com/Chia-Network/chia-blockchain/blob/7d55c39ab18609f47ce90106314244f445496947/chia/full_node/mempool_check_conditions.py#L66

richardkiss: I'm still nervous and would prefer is someone else also review this.

src/gen/conditions.rs
```diff
@@ -0,0 +1,1643 @@
...
+    pub agg_sigs: Vec<Condition>,
+}
+
+struct AnnounceState {
+    // hashing of the announcements is deferred until parsing is complete. This
+    // means less work up-front, in case parsing/validation fails
+    coin: HashSet<(Arc<[u8; 32]>, NodePtr)>,
+    puzzle: HashSet<(NodePtr, NodePtr)>,
+
+    // the assert announcements are checket once everything has been parsed and
```
richardkiss, line 222: `s/checket/checked/`

src/gen/conditions.rs
```diff
@@ -0,0 +1,1643 @@
...
+    // all create coins. Duplicates are consensus failures
+    pub create_coin: HashSet<(Vec<u8>, u64)>,
+    // Agg Sig conditions
+    pub agg_sigs: Vec<Condition>,
+}
+
+struct AnnounceState {
+    // hashing of the announcements is deferred until parsing is complete. This
+    // means less work up-front, in case parsing/validation fails
+    coin: HashSet<(Arc<[u8; 32]>, NodePtr)>,
```
richardkiss, line 219: Maybe `announcement_by_coin_id`? Is that too long? I feel like `coin` doesn't quite capture this. Same with `puzzle`

src/int_to_bytes.rs
arvidn: `n` is unsigned, so it would have to be `if n as i64 { buf.push(0); }` which would make me a bit nervous about whether rust checks for overflow in casts like that.. I suppose it doesn't

src/py/run_generator.rs
arvidn: I was having a hard time converting things to python `bytes` objects myself, that's why I went with using a `Vec` that can convert via a trait.

iirc, the issue was that the py03 `Py*` objects aren't movable or something strange like that.

src/py/run_generator.rs
arvidn: I was having a hard time getting that to work

src/py/run_generator.rs
arvidn: right. My original intention was to use the `Py` object to create native python objects, but I couldn't get that to work

src/py/run_generator.rs
arvidn: I think it's a net increase in complexity to do that. Especially to ensure we don't accidentally do something expensive, either CPU- or memory wise.

src/gen/conditions.rs
arvidn: there is no condition created for 0. 0 is assumed to always be true and is ignored.
See the places where these conditions are turned into conditions on their way to python:

https://github.com/Chia-Network/clvm_rs/pull/95/files#diff-2c8dc65d12ea32516876ce48b208f38ea2107a7f5a4f74d966a297a165cf345fR139

src/gen/conditions.rs
arvidn: fixed

src/gen/conditions.rs
arvidn: I figured since they are already part of `AnnounceState`, it's implied that they are announcements.

src/gen/conditions.rs
arvidn: if the assumption is wrong, that `ASSERT_TIME_RELATIVE 0` isn't always true. The case that would fail would be:

* a coin is created with an explicit `ASSERT_TIME_RELATIVE 0`
* in the same block, that coin is being spent

With my patch, that would be OK, the condition would never be checked.
With the current code, I believe it's also ok. but [this check](https://github.com/Chia-Network/chia-blockchain/blob/7d55c39ab18609f47ce90106314244f445496947/chia/full_node/mempool_check_conditions.py#L111) *would* fail if `expected` and `timestamp` are the same.

arvidn: @Yostra as far as I can tell, my example above *would* fail currently. Would you mind taking a look as well?

* The timestamp of the ephemeral coin is `block.foliage_transaction_block.timestamp` (see [here](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/consensus/block_body_validation.py#L375)).
* The timestamp passed in to `mempool_check_conditions_dict` is the same, `block.foliage_transaction_block.timestamp` (see [here](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/consensus/block_body_validation.py#L465)).
* The conditions checker checks current-time < expected (not <=). See [here](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/full_node/mempool_check_conditions.py#L111)

If you have an ephemeral coin with a condition `ASSERT_SECONDS_RELATIVE 0`, it would fail, because the check for that condition is that the current timestamp is *less* than the expected timestamp for the block (which is the same for ephemeral coins) + the relative time, which is 0. i.e. you end up testing `current_time < current_time`, which is false.

i.e. `ASSERT_SECONDS_RELATIVE 0` is not unconditionally true.

arvidn: @richardkiss do you agree that it's *not* safe to assume `ASSERT_SECONDS_RELATIVE 0) to be true unconditionally? However, `ASSERT_SECONDS_RELATIVE -1` *is*

arvidn: the same goes for `ASSERT_HEIGHT_RELATIVE`. 0 means it cannot be spent in the same block (ifaict)

richardkiss: I see. Rather than use 0 as the indicator to not create a condition, you might consider using `None`... any reason to not do that?

I'm reasonably convinced it the way you have it not would not change behaviour. But, it still feels a bit icky. Using `None` rather than `0` just seems a bit more explicit.

arvidn: the main reason is that `Option<uint64>` is more expensive than just `u64`. but it seems 0 is a valid condition that affects ephemeral coins, so I will need to change this. I just need a test first

src/gen/conditions.rs
arvidn: this has been updated for `height_relative`. Since 0 *is* relevant, I use an `Option<u32>` there now

src/op_utils.rs
arvidn: the logic to set up a temporary buffer for this is about as complex as this code. It might be slightly faster but I don't think there's really a strong enough motivation to make this change now. I think we should look into all of these functions (for `i32`, `u32` and `u64`) to use a temporary buffer and the built-in cast in the future.

src/gen/validation_error.rs
arvidn: done

src/gen/conditions.rs
arvidn: this has been changed for `height_relative`, to use `Option<u32>`. I expect the `Option` to double the size of the field, but 0 is relevant for relative height

arvidn:
I probably shouldn't have been so trigger happy on the squashing. But I think this is in a good place now

src/gen/sanitize_int.rs
arvidn: I don't have a strong opinion. It makes sense since it *does* treat negative numbers as an error (even if the higher level functions may chose not to). I'll make this change.

richardkiss (dismissed):

arvidn:
squashed and signed

richardkiss (approved):

PR #98, opened 2021-07-21
First crack at polyglot `Cargo.toml` for building python, npm package, and cargo crate.

richardkiss:


.github/workflows/benchmark.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         git clone https://github.com/Chia-Network/clvm_tools.git --branch=main --single-branch
         python -m pip install ./clvm_tools
         python -m pip install colorama
-        maturin develop --release
+        maturin develop --release --cargo-extra-args="--features extension-module,openssl"
```
arvidn, line 57: what does this do?

README.md
```diff
@@ -5,7 +5,7 @@ Install into current virtualenv with
 
 ```
 $ pip install maturin
-$ maturin develop --release
+$ maturin develop --release --cargo-extra-args='--features extension-module,openssl'  # on Windows, drop `openssl`
```
arvidn, line 8: there is no way to make this the default? It seems like quite a long command line to enter

pyproject.toml
```diff
@@ -1,10 +1,11 @@
 [build-system]
-requires = ["maturin"]
+requires = ["maturin>=0.11,<0.12"]
 build-backend = "maturin"
 
 [tool.maturin]
 bindings = "pyo3"
 manylinux = "2010"
+cargo-extra-args = "--features extension-module,openssl"
```
arvidn, line 8: this suggests it *is* the default. do you still need to specify it?

.github/workflows/benchmark.yml
richardkiss: I'm splitting out options like python API and OpenSSL support into "feature" flags. See also <https://doc.rust-lang.org/cargo/reference/features.html>. Seems to be the cargo way of doing it (rather than Unix vs Windows, which leaves webasm out in the cold).

README.md
richardkiss: I haven't been able to find one. I'm open to ideas.

pyproject.toml
richardkiss: Python `pep517` builds look at this file, but `cargo` doesn't. I don't think `maturin` does either.

.github/workflows/benchmark.yml
arvidn: it seems like quite inconvenient to have to type all this to build. Is there a good reason not to enable these features by default?

arvidn: I expect the end-game is to have separate projects (maybe even in separate repos) for the core `clvm_rs`, the python bindings and the wasm bindings. At that point `extension-module` wouldn't be a feature anymore, at least.

I would think `openssl` could have a default that depends on the operating system too, at least for now.

Cargo.toml
```diff
@@ -13,21 +13,18 @@ crate-type = ["cdylib"]
 [profile.release]
 lto = true
 
-[dependencies.pyo3]
-version = "0.13.2"
-
 [features]
-extension-module = ["pyo3/extension-module"]
-default = ["extension-module"]
+default = ["sha2"]
```
arvidn, line 17: afaict from the docs, if you add `"extension-module"` to this list, it will be enabled by default.

README.md
arvidn: I don't see `openssl` declared as a "feature" in Cargo.toml. Are all dependencies implicitly features also, if you say "optional=true"?

richardkiss:
Unfortunately there are problems with this approach, so I'm converting to draft for now.

richardkiss:
This feature is now in main via a different PR.

PR #100, opened 2021-08-07
Get webasm building with minimal changes.

richardkiss:


.github/workflows/build-crate-and-npm.yml
```diff
@@ -0,0 +1,81 @@
...
+    - name: Set up rust
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: nightly
+        components: rustfmt, clippy
+
+    - name: fmt
+      run: cargo +nightly fmt -- --files-with-diff --check
+    - name: clippy
+      run: cargo +nightly clippy
```
arvidn, line 41: we already have a workflow that runs `fmt` and `clippy`, right? is there a good reason to add another one?

src/wasm/api.rs
```diff
@@ -0,0 +1,113 @@
...
+        if self.strict {
+            err(o, "unimplemented operator")
+        } else {
+            op_unknown(allocator, o, argument_list, max_cost)
+        }
+    }
+}
+
+#[wasm_bindgen]
+pub fn run_clvm(program: &[u8], args: &[u8]) -> Vec<u8> {
```
arvidn, line 49: it would be nice with some smoke tests of this from javascript at least

src/wasm/api.rs
```diff
@@ -0,0 +1,113 @@
...
+    let quote_kw: u8 = 1;
+    let apply_kw: u8 = 2;
+    let max_cost: Cost = 1_000_000_000_000_000;
```
arvidn, line 52: this is very high. you don't think we should set this to at most the max block cost?

arvidn (approved):

.github/workflows/build-crate-and-npm.yml
```diff
@@ -0,0 +1,81 @@
...
+      run: cargo install wasm-pack
+
+    - name: wasm-pack
+      run: wasm-pack build && wasm-pack pack
+
+    - name: Upload artifacts
+      uses: actions/upload-artifact@v2
+      with:
+        name: npm-pkg
+        path: ./pkg/clvm_rs-0.1.9.tgz
```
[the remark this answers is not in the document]
richardkiss: Good point. Will use wildcards.

.github/workflows/build-crate-and-npm.yml
richardkiss: Mostly it's so that if there's a problem, it won't upload a half-baked version to `crates.io` or `npmjs.com` (which isn't yet implemented anyway).

src/wasm/api.rs
richardkiss: This is just supposed to be a proof-of-concept. Someone needs to write a real bridge to JavaScript, with real API. This api should actually be removed, since it's missing a lot of important stuff like `max_cost` and handling `EvalErr`.

src/wasm/api.rs
richardkiss: (I agree. See above.)

arvidn (dismissed):

arvidn (dismissed):

PR #106, opened 2021-08-19
collect all the state used while parsing

arvidn:
and validating conditions into ParseState, to simplify signature of parse_spend_conditions.

This avoids a clippy warning of `parse_spend_conditions()` taking too many parameters


richardkiss (approved):
Seems reasonable.

PR #107, opened 2021-08-20
add higher level function to run chia programs

arvidn:
where you don't need to specify operator->function mapping.


arvidn:
@richardkiss does this make sense?
The idea would be to pull this out into the future `libchia`

PR #109, opened 2021-08-23
Rust-based Dialect api.

richardkiss:


src/chia_dialect.rs
```diff
@@ -0,0 +1,95 @@
...
+use crate::err_utils::err;
+use crate::f_table::{f_lookup_for_hashmap, FLookup};
+use crate::more_ops::op_unknown;
+use crate::reduction::Response;
+use crate::run_program::OperatorHandler;
+
+const QUOTE_KW: [u8; 1] = [1];
+const APPLY_KW: [u8; 1] = [2];
+
+pub struct OperatorHandlerWithMode {
```
arvidn, line 15: this doesn't look like it belongs in the *chia* dialect. The chia dialect is known and doesn't need to be determined at run-time. It doesn't need to keep a run-time lookup-table. In fact, the run-time lookup-table seems like it belongs in the python binding, as it's a trick to allow mixing python functions with rust functions. In the rust API, there's no need for that.

src/chia_dialect.rs
```diff
@@ -0,0 +1,95 @@
...
+        h.insert(v.to_string(), [k as u8].into());
+    }
+    h
+}
+
+pub fn chia_op_handler(strict: bool) -> OperatorHandlerWithMode {
+    OperatorHandlerWithMode::new_with_hashmap(chia_opcode_mapping(), strict)
+}
+
+pub fn chia_dialect(strict: bool) -> Dialect {
```
arvidn, line 93: as far as I can tell, this is the only function that needs to be public in this file.

src/dialect.rs
```diff
@@ -0,0 +1,55 @@
+use crate::allocator::{Allocator, NodePtr};
+use crate::cost::Cost;
+use crate::reduction::Response;
+use crate::run_program::{OperatorHandler, PreEval, RunProgramContext};
+
+pub struct Dialect {
+    quote_kw: Vec<u8>,
+    apply_kw: Vec<u8>,
+    op_handler: Box<dyn OperatorHandler>,
```
arvidn, line 9: I don't see why this needs to be a run-time interface instead of a compile-time trait, not why it needs to be heap allocated. In the chia dialect case this class only contains one bool of state (strict mode).

`Dialect` seems like a perfect candidate for being a trait rather than a concrete type.

src/dialect.rs
```diff
@@ -0,0 +1,55 @@
...
+
+impl OperatorHandler for Dialect {
+    fn op(
+        &self,
+        allocator: &mut Allocator,
+        op: NodePtr,
+        args: NodePtr,
+        max_cost: Cost,
+    ) -> Response {
+        self.op_handler.op(allocator, op, args, max_cost)
```
arvidn, line 20: instead of forwarding this call, could the `Dialect` trait and `OperatorHandler` traits just be merged? They seem simple enough and related enough that it could just be one `Dialect` trait

src/dialect.rs
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+
+    pub fn run_program_with_pre_eval(
+        &self,
+        allocator: &mut Allocator,
+        program: NodePtr,
+        args: NodePtr,
+        max_cost: Cost,
+        pre_eval: Option<PreEval>,
```
arvidn, line 39: this doesn't need to be optional anymore here, since there's another version of this call for when you don't want the `pre_eval`

src/lib.rs
```diff
@@ -1,19 +1,21 @@
...
+pub mod number;
+pub mod op_utils;
 #[cfg(not(any(test, target_family = "wasm")))]
 mod py;
-mod reduction;
-mod run_program;
-mod serialize;
+pub mod reduction;
+pub mod run_program;
+pub mod serialize;
```
arvidn, line 18: it makes me a bit nervous to expose so much from the library. Is it necessary?

arvidn (commented):
I think, generally, a single interface/trait would suffice and make this API simpler. I think `Dialects` should be that trait that then has two implementations, one `ChiaDialect` which does not keep a lookup table for functions, and maybe the other one could be called `CustomDialect`, or something like that, which is configurable via a string map.

richardkiss:
Closing — ancient PR, no longer relevant.

PR #112, opened 2021-08-31
First crack at `Dialect`.

richardkiss:


src/dialect.rs
```diff
@@ -0,0 +1,73 @@
...
+            &self.apply_kw,
+            self,
+            program,
+            args,
+            max_cost,
+            Some(pre_eval),
+        )
+    }
+
+    pub fn run_program(
```
arvidn, line 55: is there a good reason to wrap `run_program()` in this class? Why not pass in the dialect to `run_program()` directly?

src/py/run_generator.rs
```diff
@@ -208,31 +207,19 @@ pub fn run_generator(
     py: Python,
     program: &[u8],
     args: &[u8],
-    quote_kw: u8,
-    apply_kw: u8,
-    opcode_lookup_by_name: HashMap<String, Vec<u8>>,
```
arvidn, line 213: I don't think we should change this API without also changing the version number of `clvm_rs` (but I absolutely think we should change this API in this direction, I just think we should be careful to stay API compatbile within version numbers, even when building from git)

arvidn: and I have queued up other changes as well. For now, maybe we can leave these as unused parameters until we cut over

tests/run_gen.py
```diff
@@ -56,9 +56,6 @@ def run_gen(fn):
     return run_generator(
         program_data,
         env_data,
-        KEYWORD_TO_ATOM["q"][0],
-        KEYWORD_TO_ATOM["a"][0],
-        native_opcode_names_by_opcode,
```
arvidn, line 61: again, I think we should hold off on this API change, and bump the version number at the same time

arvidn (commented):
I think this is a step in the right direction. I still think `f_lookup` belongs in the python binding though. But this patch is smaller in scope

src/dialect.rs
richardkiss: Mostly convenience... `Dialect` was always intended to be the highest level object... so most users will just go ```
let dialect = chia_dialect(true);
let r = dialect.run_program(...)
```

Comment (arvidn) [reply src/dialect.rs:55]:
```
let r = run_program(chia_dialect(true), ...);
```
is pretty convenient too. The problem with member functions is that every dialect will have to implement these forwarding functions, or derive from a class that that does (I'm not sure exactly how that works in rust, but I'm sure it's possible)

richardkiss: I sketched out what it would look like making `Dialect` a `trait`, but it didn't really make things simpler, so still it's a (very simple) `struct`. So making it a type parameter to `RunProgramContext` would be a circular dependency between `dialect.rs` and `run_program.rs`, which smells bad to me.

arvidn: you get the circular dependency because you add these member functions to `Dialect`. The idea of `Dialect` being a trait is that *that's* the interface/customization point.

arvidn (approved):
I still think this is a step in the right direction. Although, with `Dialect` being a template, and hence `run_program` also needing to be a template, I don't see the value of requiring state in the `Dialect` class. You can avoid templates by moving state to run-time, or you can avoid state by making it a template. Now we have both.

richardkiss:
> I still think this is a step in the right direction. Although, with `Dialect` being a template, and hence `run_program` also needing to be a template, I don't see the value of requiring state in the `Dialect` class. You can avoid templates by moving state to run-time, or you can avoid state by making it a template. Now we have both.

Okay. I'll merge this for now. If you want to do a PR changing `Dialect` to a trait, I'll definitely look at it.

PR #114, opened 2021-09-01
update run_generator API. Return a more compact representation of conditions

arvidn:
The main change is the internal and returned data structure representing conditions and the result of running a generator program.

There are two levels, the top level object are spend-bundle-wide conditions, like `reserve_fee` and `height_absolute` conditions. These do not depend on which specific spend issued them, so we just need to keep one at the end. This is called `SpendBundleConditions`. It also contains the list of `agg_sig_unsafe` keys and messages, since these also don't depend on the specific spend issuing them.

There's a list of spends (of type `Spend`) which represents a coin being spent by the spend bundle. Each spend has a coin-id, puzzle hash and some other conditions. For example `height_relative` (which depends on when the spent coin was created) or `agg_sig_me`, which depends on the coind-id of the spent coin.

This data structure is much more compact since conditions no longer are kept in lists, and coalesced immediately at the appropriate level.

This compact data structure is also returned to python and will require patching `chia-blockchain` to take advantage of these simplifications.


src/py/run_generator.rs
```diff
@@ -193,78 +123,66 @@ impl IntoPy<PyObject> for ErrorCode {
...
 
-// returns the cost of running the CLVM program along with the list of NPCs for
-// the generator/spend bundle. Each SpendConditionSummary is a coin spend along with its
-// conditions
+// returns the cost of running the CLVM program along with conditions and the list of
+// spends
 #[allow(clippy::too_many_arguments)]
 #[pyfunction]
-pub fn run_generator(
+pub fn run_generator2(
```
arvidn, line 136: The reason I didn't keep the previous `run_generator()` around for backwards compatibility was that it would require a fair amount of work (and tests) to convert the new internal data structure to the previous returned structure. It didn't seem worth the effort for not much practical benefit.

src/gen/conditions.rs
```diff
@@ -269,12 +290,13 @@ impl ParseState {
 }
 
 fn parse_spend_conditions(
+    ret: &mut SpendBundleConditions,
```
richardkiss, line 293: You can make this a method of `SpendBundleConditions`.
```suggestion
impl SpendBundleConditions {
    fn parse_spend_conditions(&mut self, ...
```

arvidn: yes, but I don't see any benefits with doing so

arvidn:
The change to chia-blockchain to use this new API are here: https://github.com/Chia-Network/chia-blockchain/pull/9645

richardkiss (approved):
I didn't review this in detail, but in principal the idea to more concisely summarize conditions based on a set of coin spends is good.

arvidn:
yeah, it also separates out conditions that don't depend on which coin introduced them. `AGG_SIG_UNSAFE`, `RESERVE_FEE` and the absolute height and time conditions. Which means the way they are coalesced is more straight forward.

PR #115, opened 2021-09-02
Dialect trait

arvidn:
I had something like this in mind.

1. merged `OperatorHandler` and `Dialect` and made it a trait.
2. implemented a hard-coded chia dialect as `ChiaDialect`
3. implemented the previous behavior as `RuntimeDialect` which takes a map of strings and opcodes as well as the quote and apply opcodes as actual (run-time) buffers. This is part of the python binding.
4. moved all handling of strings -> function pointer mapping int into the python binding subdirectory
5. made the wasm binding use the `ChiaDialect`, since it doesn't allow customizing the dialect right now anyway.
6. added a `run_chia_program()` that uses the `ChiaDialect`, without customizing operators

Using `run_chia_program()` in the benchmark seemed to give a small speedup of about 1% (compared to head of `main`):

```
      block-2000.hex mean: 0.005213 (-0.000210) -3.87 % (within uncertainty)
 compressed-2000.hex mean: 0.607450 (-0.051221) -7.78 %
          concat.hex mean: 0.648865 (+0.004932) +0.77 % (within uncertainty)
      count-even.hex mean: 0.028684 (-0.001319) -4.40 % (within uncertainty)
       factorial.hex mean: 0.235109 (+0.000326) +0.14 % (within uncertainty)
     hash-string.hex mean: 0.011903 (+0.001001) +9.18 % (within uncertainty)
       hash-tree.hex mean: 0.044673 (-0.002115) -4.52 %
     large-block.hex mean: 0.122354 (-0.006917) -5.35 %
        loop_add.hex mean: 4.440183 (-0.114218) -2.51 % (within uncertainty)
        loop_ior.hex mean: 4.519143 (-0.081650) -1.77 % (within uncertainty)
        loop_not.hex mean: 3.806513 (-0.018100) -0.47 % (within uncertainty)
        loop_sub.hex mean: 4.900199 (-0.038738) -0.78 % (within uncertainty)
        loop_xor.hex mean: 4.603299 (+0.020249) +0.44 % (within uncertainty)
 matrix-multiply.hex mean: 0.219568 (-0.003707) -1.66 % (within uncertainty)
       point-pow.hex mean: 0.986889 (+0.000685) +0.07 % (within uncertainty)
     pubkey-tree.hex mean: 0.363839 (-0.002005) -0.55 % (within uncertainty)
      shift-left.hex mean: 3.195101 (-0.002293) -0.07 % (within uncertainty)
     substr-tree.hex mean: 0.168737 (-0.004107) -2.38 %
          substr.hex mean: 0.007326 (-0.000455) -5.85 % (within uncertainty)
        sum-tree.hex mean: 0.305883 (-0.011012) -3.47 %
TOTAL: 146.104665 (-1.554372) -1.05 %
```


src/run_program.rs
```diff
@@ -277,7 +267,7 @@ impl<'a, Handler: OperatorHandler> RunProgramContext<'a, Handler> {
             return err(operator, "internal error");
         }
         let op_atom = self.allocator.atom(operator);
-        if op_atom == self.apply_kw {
+        if op_atom == self.dialect.apply_kw() {
```
richardkiss, line 270: Might be worth seeing whether this compiles to a function call or gets inlined. A function call for each `q` check would be quite a bit more expensive.

arvidn: it's a static call returning a constant value, so this certainly *could* be cheaper than comparing with an actual `Vec`

richardkiss (dismissed):
Seems good. Two potential ideas: 

1. maybe copying `apply_kw` and `quote_kw` to `RunProgramContext` to remove repeated calls to `.dialect.quote_kw` could speed things up a bit (but also, maybe not)
2. it'd probably be best to have `brun` use `run_chia_program` so we know all our `clvm` tests still work before we create a new release of clvm_rs with these changes

arvidn:
I can look into analyzing the binary, but I'm not very hopeful. My experience so far is that the debug information rust produces isn't of very high quality. I would *suspect* that the static calls with compile-time constants combined with LTO would be more efficient than comparing the `Vec` to an actual heap allocated `Vec` though.

I agree that we should make `brun` use `run_chia_program` as well. In fact, I was looking into ways of making `clvm_rs` not depend on `clvm_tools` for its tests at all. I think that would be ideal. The issue is that we can't parse the ChiaLisp without `clvm_tools`. But maybe we could at least have a dumbed-down version of `brun` in this repo that just pulls in the parsing from `clvm_tools`

src/chia_dialect.rs
```diff
@@ -1,30 +1,27 @@
...
+pub struct ChiaDialect {
     strict: bool,
 }
 
-impl OperatorHandlerWithMode {
-    pub fn new_with_hashmap(hashmap: HashMap<String, Vec<u8>>, strict: bool) -> Self {
-        let f_lookup: FLookup = f_lookup_for_hashmap(hashmap);
-        OperatorHandlerWithMode { f_lookup, strict }
+impl ChiaDialect {
+    pub fn new(strict: bool) -> ChiaDialect {
```
richardkiss, line 19: `mempool`?

arvidn: yeah, I'm planning a separate PR to rename strict-mode

richardkiss (approved):

PR #117, opened 2021-09-07
failing to terminate a list with nil is an error

arvidn:


.github/workflows/build-test.yml
```diff
@@ -196,7 +196,7 @@ jobs:
       - name: Install rust
         uses: actions-rs/toolchain@v1
         with:
-            toolchain: nightly
+            toolchain: stable
```
richardkiss, line 199: Is this intentional? I'm not sure what the rational was for using `nightly` for `fmt` and `clippy`, but it does seem like using nightly could be a little chaotic.

src/gen/conditions.rs
```diff
@@ -767,6 +766,65 @@ fn cond_test(input: &str) -> Result<(Allocator, Vec<SpendConditionSummary>), Val
...
+    assert_eq!(
+        cond_test("((8 ))").unwrap_err().1,
+        ErrorCode::InvalidCondition
+    );
+}
+
+#[test]
+fn test_invalid_spend_list_terminator() {
+    assert_eq!(
+        cond_test("((({h1} ({h2} (123 (()) 8 ))").unwrap_err().1,
```
aqk, line 823: Is this `(h1 h2 123 (()) 8)` ?

I understand this notation can have one or two values within a paren pair, and if the second one is omitted, it is filled in a s nil.

The sequence `(123 (()) 8 )` seems to have 3 top-level elements.

`cond_test("((({h1} ({h2} (123 (() 8 )))").unwrap_err().1,` ?

.github/workflows/build-test.yml
arvidn: yes, CI was failing last night. Specifically, it was emitting errors for `num-bigint`

richardkiss (approved):
This seems reasonable to me. Note that I haven't inspected the python code, although I did author the `.as_atom_list()` iterator which is used heavily in the python implementation, and was designed to allow any atom terminator.

src/gen/conditions.rs
```diff
@@ -767,6 +766,65 @@ fn cond_test(input: &str) -> Result<(Allocator, Vec<SpendConditionSummary>), Val
...
+#[test]
+fn test_invalid_condition_list2() {
+    assert_eq!(
+        cond_test("((({h1} ({h2} (123 ((8 ))))").unwrap_err().1,
+        ErrorCode::InvalidCondition
+    );
+}
+
+#[test]
+fn test_invalid_condition_list_terminator() {
```
aqk, line 786: I think `test_invalid_condition_list_terminator` needs more comments.
The name of the test suggests to the reader that the test should fail, but we understand that it is just the list that is invalid, not the test case.

This is also a great place to refer to the equivalent test in 

https://github.com/Chia-Network/chia-blockchain/pull/8393/files#diff-9439f549ee25ef5e6c0c4986a8710a0d5bb2a18d8b3c492ef41edbeb4e48c88fR2064

where you use `generator_condition_tester("(80 50 . 1)", ...)`
Translating that syntax in a comment might help someone familiar with clvm syntax.

aqk (approved):
Okay great - I don't see any other pertinent test cases.

PR #121, opened 2021-09-22
parse an optional 3rd parameter to the CREATE_COIN condition

arvidn:
as a 32 byte hash


src/gen/conditions.rs
```diff
@@ -233,7 +273,7 @@ pub struct SpendConditionSummary {
     pub seconds_relative: u64,
     pub seconds_absolute: u64,
     // all create coins. Duplicates are consensus failures
-    pub create_coin: HashSet<(Vec<u8>, u64)>,
+    pub create_coin: HashSet<NewCoin>,
```
arvidn, line 276: a plain tuple isn't sufficient anymore. The hint cannot be part of what's hashed for this `HashSet`. The point of putting the coins in a hash set is to ensure there are not duplicate outputs (i.e. identical puzzle hash and identical amount). The hint doesn't contribute to make a coin unique, so it has to be ignored for this purpose.

src/gen/conditions.rs
```diff
@@ -123,6 +126,18 @@ fn parse_args(
                 range_cache,
                 flags,
             )?;
+            // CREATE_COIN takes an optional 3rd parameter, which is a 32 byte
+            // hash. If we find anything else, that's still OK, since garbage is
+            // ignored.
+            if let Ok(c) = rest(a, c) {
+                if let Ok(param) = first(a, c) {
+                    if let SExp::Atom(b) = a.sexp(param) {
+                        if a.buf(&b).len() == 32 {
```
richardkiss, line 135: How about this?
```suggestion
                        if a.buf(&b).len() <= 32 {
```

Shouldn't be much worse, and then supports shorter hints too.

richardkiss (approved):
It's a bit scary to someone like me who is not expert with this code because it introduces new types and new cases, but near as I can tell it seems like it shouldn't change any consensus rules.

src/gen/conditions.rs
Yostra: I guess we don't have to be strict on length being exactly 32 bytes, less or equal gives some extra flexibility to clients so let's go with it

arvidn: One consequence of making this proposed change is that an *empty* atom (as long as it isn't the canonical `null` value) will still count as a *set* hint. but the canonical `null` won't. That's quite unintuitive. So, if we want hints to be any size, an empty atom should probably not count as a hint.

Yostra: that makes sense

PR #122, opened 2021-09-23
slight simplification; merge CreateCoin and CreateCoinWithHint

arvidn:


richardkiss (approved):
Good simplification. It basically makes `null` the default hint. Do you want to make the `== 32` a `<= 32` and then just have @Yostra ensure that 0-length hints don't go into the DB (since there will be a lot of them).

arvidn:
this patch only affects the structure internal to rust. In the python binding, a hint == `a.null()` are still not passed on as empty byte strings. But, yes, that might be a good idea to simplify/regularize this code a bit more

PR #123, opened 2021-09-23
bump version to 0.1.12

arvidn:


arvidn:
I think we need to update the tag to point to this commit too, once it lands

wjblanke (approved):
aok

PR #124, opened 2021-09-23
stress test with malicious generators

arvidn:


richardkiss (approved):
Excellent.

richardkiss:
(To be clear, I didn't review the correctness of the edge cases described in the document, mostly because it's written in an authoritative tone that suggest that you've examined this a lot closer than I have.)

arvidn:
yeah, ideally those test cases would be written in high level Chialisp. It's not good to have opaque tests like this. The problem, though, is that they use certain techniques not available from chialisp. For instance, chialisp always does constant folding/inlining eagerly. These programs rely on being small, but still executing large loops that *could* all be computed at compile-time, but doing so would make them have too high cost from space or be an ineffective attack since they would just boil down to a constant.

arvidn:
these were actually written by tweaking the output from the compiler, so it's not so easy to automate. Maybe if we have a more sophisticated compiler one day, we could improve them.

PR #125, opened 2021-09-24
simplify handling of CREATE_COIN hint

arvidn:
to always include it in what's returned to python. Then it's up to the caller to ignore empty hints

### DIFF
--- a/src/py/run_generator.rs
+++ b/src/py/run_generator.rs
@@ -98,17 +98,14 @@ fn convert_spend(
 
     let mut new_coins = Vec::<PyConditionWithArgs>::new();
     for new_coin in spend_cond.create_coin {
-        let mut cwa = PyConditionWithArgs {
+        new_coins.push(PyConditionWithArgs {
             opcode: CREATE_COIN,
             vars: vec![
                 PyBytes::new(py, &new_coin.puzzle_hash).into(),
                 int_to_pybytes(py, new_coin.amount),
+                node_to_pybytes(py, a, new_coin.hint),
             ],
-        };
-        if new_coin.hint != a.null() {
-            cwa.vars.push(node_to_pybytes(py, a, new_coin.hint));
-        }
-        new_coins.push(cwa);
+        });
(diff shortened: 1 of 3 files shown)


tests/test-generators.py
```diff
@@ -19,7 +19,14 @@
             output += f"  {c[0]}\n"
             for cwa in sorted(c[1], key=lambda x: (x.opcode, x.vars)):
                 output += f"    {cwa.opcode}"
-                for a in cwa.vars:
+
+                # special case to omit an empty hint from CREATE_COIN, to preserve the
+                # output from the test cases not using a hint
+                var = list(cwa.vars)
+                if cwa.opcode == 51 and len(cwa.vars) == 3 and len(cwa.vars[2]) == 0:
+                    var.pop(2)
```
arvidn, line 27: without this special case, I would have to insert an extra space " " at the end of every `CREATE_COIN` expected-output line in the existing test cases. I think that would muddy the waters a bit and not make it as obvious that all existing test cases are unaffected. It does have the issue that there's no test case ensuring that we *do* get empty hints unconditionally though.

However, I'm hoping to change this API to, instead of returning a list of list of condition codes + args, returns a concrete type with all conditions having concrete types.

src/py/run_generator.rs
```diff
@@ -98,17 +98,14 @@ fn convert_spend(
...
-        let mut cwa = PyConditionWithArgs {
+        new_coins.push(PyConditionWithArgs {
             opcode: CREATE_COIN,
             vars: vec![
                 PyBytes::new(py, &new_coin.puzzle_hash).into(),
                 int_to_pybytes(py, new_coin.amount),
+                node_to_pybytes(py, a, new_coin.hint),
             ],
-        };
-        if new_coin.hint != a.null() {
```
richardkiss, line 108: This would only work for some nulls anyway, since a clvm empty string `""` would not match.

richardkiss (approved):

PR #126, opened 2021-09-28
treat the optional 3rd create_coin parameter as a list of hints/memos

arvidn:
Still only pass back the first element to callers


src/gen/conditions.rs
```diff
@@ -1462,7 +1470,7 @@ fn test_create_coin_invalid_puzzlehash() {
 #[test]
 fn test_create_coin_with_hint() {
     // CREATE_COIN
-    let (a, spend_list) = cond_test("((({h1} ({h2} (123 (((51 ({h2} (42 ({h1} )))))").unwrap();
+    let (a, spend_list) = cond_test("((({h1} ({h2} (123 (((51 ({h2} (42 (({h1}) )))))").unwrap();
```
arvidn, line 1473: this test illustrates this change. the hint "{h1}" is now a list. (note that these tests don't use proper lisp syntax for simplicity, and to not have to depend on a lisp parser).

In this syntax "(" means a cons-box. the next two elements are its left and right side. ")" means nil, to terminate a list for instance.

PR #130, opened 2021-10-26
Changes to make `chia_rs` possible

richardkiss:


arvidn (dismissed):
I'm not very familiar with the python aspects of this. If this ABI is in fact compatible with 3.7 and later, how come it's not always used unconditionally by everyone? what would anyone gain from using a 3.8 or 3.9 specific ABI?

richardkiss:
bump

arvidn (approved):

PR #134, opened 2021-11-01
add unit tests for all operators

arvidn:
This adds unit test infrastructure to easily add unit tests for all operators.


arvidn:
ideally the cost would be verified by those test cases as well

richardkiss (approved):
Nice

PR #137, opened 2021-11-05
when initializing the GROUP_ORDER constant

arvidn:
don't parse a hexadecimal string, just use a known sequence of bytes.

The correctness of this constant is covered by several unit tests. I ensured they fail if I corrupt one of the bytes.

### DIFF
--- a/src/more_ops.rs
+++ b/src/more_ops.rs
@@ -825,8 +825,12 @@ pub fn op_softfork(a: &mut Allocator, input: NodePtr, max_cost: Cost) -> Respons
 
 lazy_static! {
     static ref GROUP_ORDER: Number = {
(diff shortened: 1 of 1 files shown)


richardkiss (approved):
What's the advantage here? This only gets run once

arvidn:
the main advantage is that it shrinks the interface required by the bignum library. this is the only place (outside of tests) where we parse a string into a bignum. This function is used in other places where we interpret an atom as an unsigned int as well (`lsh`)

arvidn:
which means we don't need a unit test for the string-parsing

richardkiss:
Ah, got it. We probably should have just used `hex::decode(...).unwrap()` as it yields `Vec<u8>` and is a lot more readable. I might do this as a PR.

arvidn:
although, the `hex` crate is only used by tests currently, so it would introduce one more dependency to production.

PR #138, opened 2021-11-05
fix M1 CI build

arvidn:


arvidn:
this doesn't fully fix the problem, but it takes one step towards fixing it. The next issue that's exposed now is:

```
ImportError: dlopen(/Users/chia/actions-runner/_work/clvm_rs/clvm_rs/venv/lib/python3.9/site-packages/blspy.cpython-39-darwin.so, 2): Library not loaded: /opt/homebrew/opt/gmp/lib/libgmp.10.dylib
  Referenced from: /Users/chia/actions-runner/_work/clvm_rs/clvm_rs/venv/lib/python3.9/site-packages/blspy.cpython-39-darwin.so
  Reason: image not found
```

it appears `blspy` expects homebrew to be installed and have `libgmp` installed as a shared library.

AmineKhaldi (approved):

.github/workflows/build-m1-wheel.yml
```diff
@@ -28,7 +39,7 @@ jobs:
 
     - name: Build m1 wheels
       run: |
-        arch -arm64 python -m venv venv
+        arch -arm64 python3 -m venv venv
```
arvidn, line 42: the first problem was that this invoked `python2` which didn't have the `venv` module

cmmarslender (approved):

PR #140, opened 2021-11-05
round-trip tests for Number

arvidn:
The bulk of these tests are round-tripping bytes -> number and back. This is primarily ensuring the bignum behaves as we expect it to.
This also adds a few more test cases to the `l` operator, suggested by @Quexington .


src/f_table.rs
```diff
@@ -64,9 +64,7 @@ pub fn f_lookup_for_hashmap(opcode_lookup_by_name: HashMap<String, Vec<u8>>) ->
         if idx.len() == 1 {
             let index = idx[0];
             let op = opcode_by_name(name);
-            if op.is_none() {
-                panic!("can't find native operator {:?}", name);
-            }
+            assert!(!op.is_none(), "can't find native operator {:?}", name);
```
richardkiss, line 67:
```suggestion
            assert!(op.is_some(), "can't find native operator {:?}", name);
```
is a tad more readable.

src/more_ops.rs
```diff
@@ -107,6 +107,8 @@ fn limb_test_helper(bytes: &[u8]) {
 #[test]
 fn test_limbs_for_int() {
     limb_test_helper(&[]);
+    limb_test_helper(&[0x1]);
+    limb_test_helper(&[0x80]);
```
richardkiss, line 111: maybe `0x81` too?

src/test_ops.rs
```diff
@@ -633,6 +633,9 @@ l => FAIL
 l ( 1 2 ) 1 => FAIL
 l ( 1 2 3 ) => 1
 l 1 => 0
+l null => null
```
richardkiss, line 636: I would recommend against using`null`, as it's not used anywhere else around CLVM (just use `0`).

richardkiss (approved):

src/test_ops.rs
arvidn: it is used on the allocator, which has a member function `null()`, but I see your point. The reason I ended up using it here was because the first version of this test framework didn't support decimal numbers.

arvidn: making this change is rather big, and I'm not 100% sold on it. right now there is a certain simplicity of having a special token for what `Allocator::null()` returns, rather than depending on (the quite involved) logic of parsing integers. On the other hand, that integer parsing better be correct as well..

src/more_ops.rs
arvidn: added

src/test_ops.rs
richardkiss: Ah, I see what motivates this now. My only concern is that it might be confusing to someone looking at it coming from a clvm use background rather than a clvm implementation background. A compromise might be to look for the string `0` (instead of `null`) and call `allocator.null`, and also assert that it matches `0` as a string.

richardkiss (approved):

src/test_ops.rs
arvidn: yeah, I like that

PR #141, opened 2021-11-05
deprecate op_div

arvidn:


tests/generators/duplicate-height-absolute-div.clvm
```diff
@@ -0,0 +1,6 @@
+(q ((0x0101010101010101010101010101010101010101010101010101010101010101 (a (q 2 2 (c 2 (c (q 83 100) (c 5 ())))) (c (q 2 (i 11 (q 4 5 (a 2 (c 2 (c 5 (c (/ 11 (q . 2)) ()))))) ()) 1) (q 0x7ff))) 123 (() (q 0x7f)))))
```
arvidn, line 1: this program invokes `/`, and hence fails in strict mode

src/test_ops.rs
```diff
@@ -629,7 +641,7 @@ point_add 0x97f1d3a73197d7942695638c4fa9ac0fc3688c4f9774b905a14e3a3f171bac586c55
 point_add 0x97f1d3a73197d7942695638c4fa9ac0fc3688c4f9774b905a14e3a3f171bac586c55e83ff97a1aeffb3af00adb22c6bb00 => FAIL
 point_add null => FAIL
 
-; they point must be an atom
+; the point must be an atom
```
aqk, line 644: "the argument to point_add" is a little easier to read

src/more_ops.rs
```diff
@@ -441,6 +441,11 @@ pub fn op_div(a: &mut Allocator, input: NodePtr, _max_cost: Cost) -> Response {
     }
 }
 
+pub fn op_div_deprecated(a: &mut Allocator, input: NodePtr, _max_cost: Cost) -> Response {
+    let args = Node::new(a, input);
+    args.err("div operator is deprecated")
```
aqk, line 446: This is a nice way to deprecate it

tests/run_gen.py
```diff
@@ -53,15 +53,19 @@ def run_gen(fn, flags=0):
...
+        return run_generator(
+            program_data,
+            env_data,
+            KEYWORD_TO_ATOM["q"][0],
+            KEYWORD_TO_ATOM["a"][0],
+            native_opcode_names_by_opcode,
+            max_cost,
+            flags,
+        )
+    except Exception as e:
```
aqk, line 66: Should you catch a more specific `Exception` here?

aqk (approved):

tests/run_gen.py
arvidn: this is part of the test, so I don't think it would make much sense in allowing the test to terminate with an exception. This makes it possible to test for a specific failure (as the new generator test is an example of)

richardkiss (approved):

src/chia_dialect.rs
```diff
@@ -46,7 +46,7 @@ impl OperatorHandler for OperatorHandlerWithMode {
     }
 }
 
-pub fn chia_opcode_mapping() -> HashMap<String, Vec<u8>> {
+pub fn chia_opcode_mapping(strict: bool) -> HashMap<String, Vec<u8>> {
```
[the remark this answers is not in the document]
arvidn: I agree with this, but I think this name need to be changed in one more place too, and accepting this here would invalidate the approvals this PR has. I'll make a separate PR for this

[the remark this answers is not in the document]
arvidn: https://github.com/Chia-Network/clvm_rs/pull/144

PR #142, opened 2021-11-08
Make `openssl` a feature rather than a platform choice.

richardkiss:


arvidn:
all benchmarks are slowing down with this patch too.. it's mysterious

arvidn (approved):

PR #143, opened 2021-11-08
Build fewer wheels using `abi3`

richardkiss:


richardkiss:
@arvidn can you eyeball the benchmarks to make sure there is nothing obviously worse? I'd also like to try doing a full sync with this binary before we merge it.

arvidn:
the change looks good as far as I can tell, but for some reason the benchmarks are a lot worse than main.

Windows: 316s -> 357s
Ubuntu: 166s -> 185s
Mac: 271s -> 287s

Windows shouldn't be affected by this change, right? so maybe this is just noise (it seems like a lot of noise though).

richardkiss:
I removed the `openssl` part of the patch to see how it affects the benchmarks.

arvidn (approved):

PR #146, opened 2021-11-10
use div_mod_floor from num-bigint

arvidn:
this allows for a few simplifications of the code.

@richardkiss can you think of more edge cases to test for `pubkey_for_exp`?


richardkiss:
> this allows for a few simplifications of the code.
> 
> @richardkiss can you think of more edge cases to test for `pubkey_for_exp`?

May I recommend more negative tests... -1 to -4 or so (I see -2 curiously), and -GROUP_ORDER and within three or four of either side that value too.

richardkiss (approved):

PR #152, opened 2021-11-12
use libgmp for bignum instead of num-bigint

arvidn:
the `mpz_import` and `mpz_export` functions are quite a bit slower than the `num-bigint` counterparts, so this slows down the benchmarks somewhat.

The rationale for this is:

1. use more main-stream code, with lower security risks
2. (possibly) use the same instance of libgmp as we already use in `chiavdf` and `bls-signatures`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,11 +83,9 @@ name = "clvm_rs"
 version = "0.1.15"
 dependencies = [
  "bls12_381",
+ "gmp-mpfr-sys",
  "hex",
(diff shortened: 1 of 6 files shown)


arvidn:
interestingly, there seems to be a rule preventing building the FFI layer on windows: https://github.com/Chia-Network/clvm_rs/runs/4189087870?check_suite_focus=true#step:7:104

That probably has to be addressed before we can try to get MPIR working.

PR #158, opened 2021-12-11
fix test-generator test

arvidn:
I'm not sure why `sorted()` used to work in `run_gen.py` and `test-generators.py`. The type we export from `clvm_rs` does not implement `<`. This patch fixed that and also includes a few blocks from the mainnet chain as regression tests.


tests/test-generators.py
```diff
@@ -73,6 +73,16 @@ def parse_output(result, error_code):
             limit = 9
         elif "negative-reserve-fee.clvm" in g:
             limit = 4
+        elif "block-834752" in g:
```
richardkiss, line 76: Might want to (in future) just make this a list of exceptions that you iterate through. Might be a bit cleaner.

richardkiss: Something like

```python
TEST_LIMITS = [("duplicate-coin-announce", 9), ("negative-reserve-fee", 4), ...]
...
for k, v in TEST_LIMITS:
    if k in g:
        limit = v
        break
```

richardkiss (dismissed):

tests/test-generators.py
arvidn: Yeah, I'm hoping to improve the performance to make all these run in less than 1.5s. Maybe if this list gets longer before that happens...

richardkiss (approved):

PR #163, opened 2022-01-12
split out the different features enabled by strict mode

arvidn:
and rename it to mempool-mode. This is to allow enabling them independently, for scheduled soft forks

### DIFF
--- a/src/chia_dialect.rs
+++ b/src/chia_dialect.rs
@@ -11,13 +11,20 @@ use crate::more_ops::{
 };
 use crate::reduction::Response;
 
+// division with negative numbers are disallowed
+pub const NO_NEG_DIV: u32 = 0x0001;
+
+// unknown operators are disallowed
+// (otherwise they are no-ops with well defined cost)
+pub const NO_UNKNOWN_OPS: u32 = 0x0002;
+
 pub struct ChiaDialect {
-    strict: bool,
+    flags: u32,
 }
 
 impl ChiaDialect {
-    pub fn new(strict: bool) -> ChiaDialect {
-        ChiaDialect { strict }
+    pub fn new(flags: u32) -> ChiaDialect {
+        ChiaDialect { flags }
     }
 }
 
@@ -31,7 +38,7 @@ impl Dialect for ChiaDialect {
     ) -> Response {
         let b = &allocator.atom(o);
         if b.len() != 1 {
-            return if self.strict {
+            return if (self.flags & NO_UNKNOWN_OPS) != 0 {
                 err(o, "unimplemented operator")
             } else {
                 op_unknown(allocator, o, argument_list, max_cost)
@@ -55,7 +62,7 @@ impl Dialect for ChiaDialect {
             17 => op_subtract,
             18 => op_multiply,
             19 => {
-                if self.strict {
+                if (self.flags & NO_NEG_DIV) != 0 {
                     op_div_deprecated
                 } else {
                     op_div
@@ -79,7 +86,7 @@ impl Dialect for ChiaDialect {
             // 35 ---
             36 => op_softfork,
             _ => {
-                if self.strict {
+                if (self.flags & NO_UNKNOWN_OPS) != 0 {
                     return err(o, "unimplemented operator");
                 } else {
                     return op_unknown(allocator, o, argument_list, max_cost);
--- a/src/gen/flags.rs
+++ b/src/gen/flags.rs
@@ -1,0 +1,8 @@
+// flags controlling to condition parsing
+
+// in conditions output, integers must use canonical encoding (i.e. no redundant
+// leading zeros)
+pub const COND_CANON_INTS: u32 = 0x010000;
+
+// unknown condition codes are disallowed
+pub const NO_UNKNOWN_CONDS: u32 = 0x20000;
--- a/src/py/api.rs
+++ b/src/py/api.rs
@@ -7,7 +7,10 @@ use super::run_program::{
     __pyo3_get_function_deserialize_and_run_program2, __pyo3_get_function_run_chia_program,
     __pyo3_get_function_serialized_length,
 };
-use crate::run_program::STRICT_MODE;
+use crate::chia_dialect::{NO_NEG_DIV, NO_UNKNOWN_OPS};
+use crate::gen::flags::{COND_CANON_INTS, NO_UNKNOWN_CONDS};
+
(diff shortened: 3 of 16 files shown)


src/dialect.rs
```diff
@@ -2,6 +2,26 @@ use crate::allocator::{Allocator, NodePtr};
...
+pub const COND_CANON_INTS: u32 = 2;
+
+// unknown condition codes are disallowed
+pub const NO_UNKNOWN_CONDS: u32 = 4;
+
+// unknown operators are disallowed
+// (otherwise they are no-ops with well defined cost)
+pub const NO_UNKNOWN_OPS: u32 = 8;
+
+// this constant is exported by the python binding, but when building and
```
richardkiss, line 19: Consider defining it in the python bindings code, so we don't need the `dead_code` label.

src/dialect.rs
```diff
@@ -2,6 +2,26 @@ use crate::allocator::{Allocator, NodePtr};
...
 use crate::reduction::Response;
 
+// division with negative numbers are disallowed
+pub const NO_NEG_DIV: u32 = 1;
+
+// in conditions output, integers must use canonical encoding (i.e. no redundant
+// leading zeros)
+pub const COND_CANON_INTS: u32 = 2;
+
+// unknown condition codes are disallowed
```
richardkiss, line 12: The scope is wrong here. `NO_UNKNOWN_CONDS` and `COND_CANON_INTS` are not properties of the dialect; they're used by `run_generator` which includes a dialect and other flags that affect post-processing. Can you move these somewhere which makes it clear it's not part of the dialect?

As an example, if we move `run_generator` to `chia_rs`, then these constants will have to move there anyway.

src/chia_dialect.rs
```diff
@@ -12,12 +13,12 @@ use crate::more_ops::{
 use crate::reduction::Response;
 
 pub struct ChiaDialect {
-    strict: bool,
+    flags: u32,
```
richardkiss, line 16: Any reason not to simply take a list of booleans? I know `flags` is quite traditional, but let me ask you: when have you seen an API that takes `flags` as a parameter and been pleased about it?

src/dialect.rs
arvidn: yeah, I'm hoping we can do that too. I can move the definitions of these flags, but I don't want to prepare that transition too much before we have a better sense of what it will look like and what will work

src/chia_dialect.rs
arvidn: generally I find flags better in all respects.

1. They take less space
2. The enforce names to be used (rather than 'true' and 'false')
3. They are easy to pass on to the next function

e.g.

```
f(true, true, false);
```

vs.

```
f(COND_CANON_INTS | NO_NEG_DIV);
```

Granted, in this specific place, rust will enforce names to be used to initialize the struct, but if used as parameters to any function, I think a series of bools would be inferior because of (1) and (3).

These flags are not type safe right now, but I expect that to be a small change to make.

src/chia_dialect.rs
```diff
@@ -2,6 +2,7 @@ use crate::allocator::{Allocator, NodePtr};
 use crate::core_ops::{op_cons, op_eq, op_first, op_if, op_listp, op_raise, op_rest};
 use crate::cost::Cost;
 use crate::dialect::Dialect;
+use crate::dialect::{NO_NEG_DIV, NO_UNKNOWN_OPS};
```
richardkiss, line 5: These flags belong in this file, not in `dialect.rs`, as they only modify `ChiaDialect`.

src/chia_dialect.rs
richardkiss: Maybe consider using [bitflags](https://docs.rs/flags/latest/bitflags/) or something to get typed flags. (I've not used this library so can't vouch for its goodness or badness.)

src/dialect.rs
```diff
@@ -2,6 +2,13 @@ use crate::allocator::{Allocator, NodePtr};
 use crate::cost::Cost;
 use crate::reduction::Response;
 
+// division with negative numbers are disallowed
+pub const NO_NEG_DIV: u32 = 0x0001;
```
richardkiss, line 6: As mentioned, these flags belong in `chia_dialect.rs`.

src/gen/flags.rs
```diff
@@ -0,0 +1,8 @@
+// flags controlling to condition parsing
+
+// in conditions output, integers must use canonical encoding (i.e. no redundant
+// leading zeros)
+pub const COND_CANON_INTS: u32 = 0x1000;
+
+// unknown condition codes are disallowed
+pub const NO_UNKNOWN_CONDS: u32 = 0x1001;
```
richardkiss, line 8: Why does this flag have two bits set?

src/gen/flags.rs
```diff
@@ -0,0 +1,8 @@
+// flags controlling to condition parsing
+
+// in conditions output, integers must use canonical encoding (i.e. no redundant
+// leading zeros)
+pub const COND_CANON_INTS: u32 = 0x1000;
```
richardkiss, line 5: You could probably dump these flags into `run_gen`.

src/py/api.rs
```diff
@@ -7,15 +7,22 @@ use super::run_program::{
     __pyo3_get_function_deserialize_and_run_program2, __pyo3_get_function_run_chia_program,
     __pyo3_get_function_serialized_length,
 };
-use crate::run_program::STRICT_MODE;
+use crate::dialect::{NO_NEG_DIV, NO_UNKNOWN_OPS};
+use crate::gen::flags::{COND_CANON_INTS, NO_UNKNOWN_CONDS};
+
+pub const MEMPOOL_MODE: u32 = NO_NEG_DIV | COND_CANON_INTS | NO_UNKNOWN_CONDS | NO_UNKNOWN_OPS;
```
richardkiss, line 13: This seems odd to merge flags across two separate namespaces. Are you sure you really want to do this?

src/chia_dialect.rs
arvidn: yeah, I've only had a brief look at it and wasn't very impressed so far.

src/gen/flags.rs
arvidn: hm, that's not right

src/py/api.rs
arvidn: not in the long run, just for now

src/gen/flags.rs
arvidn: they don't really belong to the python binding either though.

richardkiss (approved):

richardkiss (approved):

PR #167, opened 2022-01-27
nil-terminated arg lists

arvidn:
add flag to require AGG_SIG_ME and AGG_SIG_UNSAFE conditions to have a propertly nil terminated argument list

### DIFF
--- a/src/gen/conditions.rs
+++ b/src/gen/conditions.rs
@@ -4,10 +4,10 @@ use super::condition_sanitizers::{
 };
 use super::rangeset::RangeSet;
 use super::sanitize_int::sanitize_uint;
-use super::validation_error::{first, next, rest, ErrorCode, ValidationErr};
+use super::validation_error::{check_nil, first, next, rest, ErrorCode, ValidationErr};
 use crate::allocator::{Allocator, NodePtr, SExp};
 use crate::cost::Cost;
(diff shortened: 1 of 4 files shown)


src/gen/conditions.rs
```diff
@@ -96,23 +96,35 @@ fn parse_args(
...
+            if (flags & COND_ARGS_NIL) != 0 {
+                // make sure there aren't more than two
+                check_nil(a, rest(a, c)?)?;
+                Ok(Condition::AggSigUnsafe(pubkey, message))
+            } else {
+                // but the argument list still doesn't need to be terminated by NIL,
+                // just any atom will do
+                match a.sexp(rest(a, c)?) {
+                    SExp::Pair(_, _) => Err(ValidationErr(c, ErrorCode::InvalidCondition)),
+                    _ => Ok(Condition::AggSigUnsafe(pubkey, message)),
```
richardkiss, line 108: Consider doing the `if flags` check here... seems like it might be a smaller change.

(Obviously not a critical change.)

richardkiss (approved):

src/gen/conditions.rs
arvidn: true. in my mind, the main reason to make this change is the opportunity to remove the existing logic down the line. So that's what I had in mind, what this code would look like with just the new rule

PR #169, opened 2022-02-01
New release

richardkiss:


richardkiss:
`cargo` doesn't like the crate and the wheel to both be named `clvm_rs` so the crate is now known as `clvmr`. (The name `clvm` is not currently available on `crates.io`.)

arvidn (approved):
maybe the rust package could be `chialisp`

richardkiss:
> maybe the rust package could be `chialisp`

I think we've been using that name for the high level language, to distinguish it from `clvm`, which is just the virtual machine (so really `clvm_tools` could be called that).

PR #170, opened 2022-02-02
Remove `pyo3` from `clvmr` crate.

richardkiss:


arvidn (commented):
I'm curious about `f_table.rs`. That's part of the python binding, right?

richardkiss:
> I'm curious about `f_table.rs`. That's part of the python binding, right?

The goal of this patch was to remove `pyo3` from the root crate (which is uploaded to crates.io as `clvmr`), so it can focus on providing the best clvm implementation without distraction. This file has no dependencies on `pyo3`, and it might be useful for other FFI implementations, so it can go in the root. We might consider renaming it `ffi_utilities` or something though.

arvidn (approved):

PR #176, opened 2022-03-01
add new function, run_chia_program2()

arvidn:
that only takes a program, no arguments. To run programs with arguments, it can be combined with the apply operation

### DIFF
--- a/src/run_program.rs
+++ b/src/run_program.rs
@@ -287,8 +287,8 @@ impl<'a, D: Dialect> RunProgramContext<'a, D> {
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/run_program.rs
```diff
@@ -329,15 +329,27 @@ impl<'a, D: Dialect> RunProgramContext<'a, D> {
 }
 
 pub fn run_program<'a, D: Dialect>(
+    allocator: &'a mut Allocator,
```
richardkiss, line 332: maybe `run_program_with_program_arg_pair` ? The thing being passed in is not what I think of as a program, but rather a program/argument pair.

arvidn:
I think I will abandon this patch

PR #177, opened 2022-03-01
instead of depending on test cases from clvm and clvm_tools

arvidn:
rely on the existing unit tests in src/test_ops.rs

This is a step towards untangling `clvm_rs` from `clvm` and `clvm_tools`, which will allow us to change `clvm_tools`'s `brun` to report accurate costs without breaking the tests in this repository.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -112,35 +112,10 @@ jobs:
         python -m pip install pytest
         python -m pip install blspy
 
-    - name: install clvm & clvm_tools
+    - name: Ensure clvm_rs is installed
       run: |
         . ./activate
-        git clone https://github.com/Chia-Network/clvm.git --branch=main --single-branch
-        python -m pip install ./clvm
-
-        git clone https://github.com/Chia-Network/clvm_tools.git --branch=main --single-branch
-        python -m pip install ./clvm_tools
-
-    - name: Ensure clvm, clvm_rs, clvm_tools are installed
-      run: |
-        . ./activate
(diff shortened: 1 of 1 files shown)


richardkiss:
Taking out the `clvm_tools` tests makes perfect sense, since the test results are ignored due to stage 2 not working with `clvm_rs`. I do have concerns about taking out the `clvm` tests permanently. We could for now create a branch of `clvm` specifically to be imported by `clvm_rs` (maybe call it `clvm_rs_tests`).

arvidn:
right. running the `clvm` tests still depend on `clvm_tools` though. I think long term, those test cases should be executed in a more direct way, that doesn't go through `brun`.

If I would go through and make sure all test cases in the existing `.txt` files are present in the `test_ops.rs` test cases, do you think that would be a good enough substitute?

richardkiss:
Sure. The more similar the test formats can be between the two, the easier it will be to keep the two in sync if we add more tests or change existing ones. So I'd recommend trying to keep the same text input/output format, maybe even the exact same files, reading them at runtime.

arvidn:
I don't see the value in tying `clvm_rs` together with `clvm`. We don't really need to ensure they move in lock step anymore, do we? e.g. the `op_div` softfork hasn't been implemented in `clvm` yet, nor is it tested by those test cases.

To me it seems it would only keep us bogged down, slowing down updates to `clvm_rs` and `clvm_tools`.

Had they been part of the same repository, and changes could be made to all 3 projects in single commits, it would have been different, I think. At least simpler.

PR #179, opened 2022-03-02
test cost in operations unit tests

arvidn:
this makes the tests more effective regression tests


richardkiss (approved):
Interesting choice to use "|" as a separator.

PR #180, opened 2022-03-02
Remove `gen`, which is now in `chia_rs`.

richardkiss:


wheel/src/api.rs
```diff
@@ -2,31 +2,21 @@ use pyo3::prelude::*;
...
 use super::run_program::{
     __pyo3_get_function_deserialize_and_run_program2, __pyo3_get_function_run_chia_program,
     __pyo3_get_function_serialized_length,
 };
 use clvmr::chia_dialect::{NO_NEG_DIV, NO_UNKNOWN_OPS};
-use clvmr::gen::flags::{COND_ARGS_NIL, COND_CANON_INTS, NO_UNKNOWN_CONDS};
 
-pub const MEMPOOL_MODE: u32 =
-    NO_NEG_DIV | COND_CANON_INTS | NO_UNKNOWN_CONDS | NO_UNKNOWN_OPS | COND_ARGS_NIL;
+pub const MEMPOOL_MODE: u32 = NO_NEG_DIV | NO_UNKNOWN_OPS;
```
richardkiss, line 11: I had to remove some flags here, which changes the value of `MEMPOOL_MODE`. Should we just remove `MEMPOOL_MODE` and use the one in `chia_rs`? @arvidn

richardkiss: I decided to put back `NO_UNKNOWN_OPS` (which was just an oopsie) and remove `MEMPOOL_MODE`.

arvidn: I think it's reasonable that `clvm_rs` and `chia_rs` both have `MEMPOOL_MODE` flags, with different flags enabled. At least for now, we may want to turn it into separate flags at some point

arvidn (approved):

wheel/src/api.rs
richardkiss: The concern I have is that someone uses the `clvm_rs.MEMPOOL_MODE` flags in `chia_rs.run_generator` and then doesn't get all the strict checking they were hoping for because they don't realize the values are different.

PR #187, opened 2022-04-29
Wheel's `run`-like methods now return `(str, LazyNode)` in failure case.

richardkiss:
This removes dependency on `clvm` python wheel.


richardkiss:
This is a fairly simple change. The main benefit is that it no longer does a clunky call out to python with `import clvm` in the rust code. The main disadvantage is that we always return a pair `(cost_or_message, clvm_object)` where `cost_or_message` is an `int` on success and a `str` on failure. So python code that calls these functions should look for a `str` coming back and do a `raise EvalError` if it wants to maintain compatibility.

@arvidn  Should I deprecate `deserialize_and_run_program2`? Not sure if anything uses it anymore.

arvidn:
I think it makes sense to remove `deserialize_and_run_program2`, yes.

Is it typical python practice to return a "variant" object like that? does the caller need to use `type(cost_or_message)` to know whether the call succeeded or not? it seems a bit strange.

Is the CI failure happening in `main` as well, do you know?

wheel/src/run_program.rs
```diff
@@ -75,7 +94,7 @@ pub fn run_chia_program(
     args: &[u8],
     max_cost: Cost,
     flags: u32,
-) -> PyResult<(Cost, LazyNode)> {
+) -> PyResult<(PyObject, LazyNode)> {
```
arvidn, line 97: this seems risky without also changing the name of the function. The return value will be backwards compatible, so if the python code doesn't update, it might still appear to work, as long as no errors happen

trepca:
@richardkiss shall we close this?

PR #189, opened 2022-06-02
Make single argument raise return its only argument for ease of visual inspection (re quexington)

prozacchiwawa:
A request was made that when x is given a single argument, the single argument is returned, making the bubbled result in python easier to inspect visually.

### DIFF
--- a/src/core_ops.rs
+++ b/src/core_ops.rs
@@ -60,26 +60,23 @@ pub fn op_listp(a: &mut Allocator, input: NodePtr, _max_cost: Cost) -> Response 
 
 pub fn op_raise(a: &mut Allocator, input: NodePtr, _max_cost: Cost) -> Response {
     let args = Node::new(a, input);
-    // Quexington requested single argument raise to return the single argument
-    // rather than the full list of arguments.  brun also used to behave this
-    // way... it does make sense to ensure that the full context of the exception
-    // is passed on rather than a more cryptic error in the case of more arguments
-    // though (IMO).  We restrict the single argument case to atoms since it
-    // ensures that a single argument with a pair is distinguishable from multiple
-    // independent arguments (as new, re: richard).
+    // if given a single argument we should raise the single argument rather
+    // than the full list of arguments. brun also used to behave this way.
+    // if the single argument here is a pair then don't throw it unwrapped
+    // as it'd potentially look the same as a throw of multiple arguments.
     let throw_value = args
         .pair()
         .as_ref()
         .and_then(|(_first, _rest)| {
-            _first.atom().map(|_| {
+            _first.atom().and_then(|_| {
                 if _rest.nullp() {
-                    args.first().unwrap()
+                    Some(args.first().unwrap())
                 } else {
-                    args.clone()
+                    None
                 }
             })
         })
-        .unwrap_or_else(|| args);
+        .unwrap_or(args);
 
     throw_value.err("clvm raise")
 }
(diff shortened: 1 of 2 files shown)


richardkiss:
At first, I didn't like this because I thought it made it impossible to determine what was passed to `x` if it returns just a single item. But then I realized the way it works now, `x` always returns a pair in the `sexp` field (as opposed to an atom). There is still ambiguity in this case: `(x (q . (foo bar)))` because after this patch, it would return the same as `(x (q . foo) (q . bar))` even though they're clearly different.

HOWEVER... if the sole parameter passed to `x` is a single atom, then I think it's okay to strip it out. Because then it's no longer ambiguous. So `(x (q . foo))` can return `foo` as the failure context because it's now obvious what's happened.

If you can change this patch to also check that it's of the form `(x VALUE)` then the stripping is done iff `VALUE` is an atom, then I'll approve it.

prozacchiwawa:
ok it now isolates an atom as the single argument, and only then returns the bare atom.

src/core_ops.rs
```diff
@@ -60,7 +60,28 @@ pub fn op_listp(a: &mut Allocator, input: NodePtr, _max_cost: Cost) -> Response
...
 pub fn op_raise(a: &mut Allocator, input: NodePtr, _max_cost: Cost) -> Response {
     let args = Node::new(a, input);
-    args.err("clvm raise")
+    // Quexington requested single argument raise to return the single argument
+    // rather than the full list of arguments.  brun also used to behave this
+    // way... it does make sense to ensure that the full context of the exception
+    // is passed on rather than a more cryptic error in the case of more arguments
+    // though (IMO).  We restrict the single argument case to atoms since it
+    // ensures that a single argument with a pair is distinguishable from multiple
+    // independent arguments (as new, re: richard).
```
richardkiss, line 69: This might be good history, but I'd prefer if the comment just explained what it the code is doing.

src/core_ops.rs
```diff
@@ -60,7 +60,28 @@ pub fn op_listp(a: &mut Allocator, input: NodePtr, _max_cost: Cost) -> Response
...
+    let throw_value = args
+        .pair()
+        .as_ref()
+        .and_then(|(_first, _rest)| {
+            _first.atom().map(|_| {
+                if _rest.nullp() {
+                    args.first().unwrap()
+                } else {
+                    args.clone()
```
richardkiss, line 78: Do we need this clone?

src/core_ops.rs
```diff
@@ -60,7 +60,28 @@ pub fn op_listp(a: &mut Allocator, input: NodePtr, _max_cost: Cost) -> Response
...
+                }
+            })
+        })
+        .unwrap_or_else(|| args);
+
+    throw_value.err("clvm raise")
```
richardkiss, line 84: Can you fix the clippy warning?

src/core_ops.rs
prozacchiwawa: i think so.  i can do it a different way.

richardkiss (changes requested):
Looks good, just a few minor nits.

src/core_ops.rs
prozacchiwawa: ok i think all those are done

prozacchiwawa:
ok i think all those are done

arvidn:
This changes the output from 3 of the regression tests in `clvm` (the ones testing `x`). It wasn't discovered by CI when this PR was posted because of two bugs causing `brun` to silently run the python implementation instead of the rust one.

With those issues fixed, we now need to correct this discrepancy.

arvidn:
https://github.com/Chia-Network/clvm/pull/127

PR #192, opened 2022-06-30
Serialization with backrefs

richardkiss:
This PR extends serialization to a "version 2" of serialization that uses the (currently unused) byte `0xfe` to mean "back-reference".

A serialization of an atom follows the `0xfe`, and indicates (using standard clvm path semantics) where in the "partial deserialization stack" we should fetch a reused clvm subtree. This is very effective for large duplicate subtrees, as is common when a block contains multiple puzzle reveals of the same type, ie. have the same `mod`.

The patch includes:

- implementation of serialization with back-references (in rust)
- implementation of deserialization (in clvm)
- a command-line tool `swbr` ("serialize with back-references"), basically for testing in the shell, that takes hex serialization and re-serializes with back-references, and optionally wraps it in the clvm deserialization code so you it becomes the clvm equivalent of a "self-extracting archive"

The goal is to use this in block compression. I've tried it on a hundred thousand or so blocks prior to 400000. Compression generally takes less than a second on my MacBook Air. I haven't yet done much investigation in to how much cheaper blocks become when they're compressed, but the extra clvm code included is quite minimal and runs cheaply, and the compression is much better than what we have now for most blocks (currently, we only do back-refs to the `mod` from the standard transaction, so NFTs and CATs should benefit greatly). I've seen some large blocks compress to 20% or less of their original size.

It might have been useful to mitigate some expansion attacks across process boundaries, since an attacker could make a tiny generator that expands into a massive serialization by repeating subtrees, but work that @arvid has done has already addressed that. The only clear use case now is block compression. Because of that, it might actually be reasonable to put this new code in `chia_rs` instead of `clvm_rs`.


arvidn:
the `Bytes32` type seems to be very similar to the one in `chia_rs`. perhaps some types from there should be moved into `clvm_rs`. Or treehash should be moved to `chia_rs` (and the cache would use some other hashing)

arvidn:
@richardkiss Please add a description to this PR. Thins that aren't obvious to me are:

1. what are all the new dependencies for? Are they all transitive dependencies from bumping the crypto package version?
2. There appears to be an implementation of this in CLVM also, why?
3. what's "swbr" short for?

src/read_cache_lookup.rs
```diff
@@ -0,0 +1,232 @@
...
+            paths.sort();
+            paths.truncate(1);
+            paths.pop()
+        } else {
+            None
+        }
+    }
+}
+
+fn reversed_path_to_vec_u8(path: &[u8]) -> Vec<u8> {
```
arvidn, line 182: could you add a comment describing what this function does?

src/read_cache_lookup.rs
```diff
@@ -0,0 +1,232 @@
...
+    parent_lookup: HashMap<Bytes32, Vec<(Bytes32, u8)>>,
+}
+
+impl Default for ReadCacheLookup {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl ReadCacheLookup {
```
arvidn, line 43: this class should have unit tests

arvidn (changes requested):
I'm interested in benchmarks of code.
1. how much space savings does it give us? (say, on average in the last 100000 blocks).
2. What's the trade-off between saving bytes and paying higher CLVM cost for decompression?
3. how much more expensive is it to decompress? (I imagine that's the most relevant metric)
4. What's the worst case scenario? Are there any sensitive parts that could be a DoS vector? having some worst-case benchmarks would be great. The worst case memory usage would also be interesting. I think there's some potential for reducing the memory footprint (and the worst case is already pretty high without compression).

There should also be coverage of the compression end of things. It's important farmers don't fail to produce a block.

Generally, there are a lot of missing unit tests. I would also expect somewhat extensive round-trip tests, to ensure all edge cases in the compression can be decompressed.

I think both the rust and clvm implementation of decompression should have a fuzzer driver. We have all the infrastructure in place for that, so it shouldn't be a bit deal to add.
Maybe even the serializer could be fuzzed, it's a bit tricky though, since the tree structure would need to be built from a noise input.

Lastly, the serialization format needs to be documented.

arvidn:
It seems likely that we would get more savings by applying this compression at the db layer and leave the farmed blocks as they are.

* We wouldn't have to wrap the generator in the decompressor program (which is a lot of repetitive code)
* We could decompress it in rust, which would be faster and use less memory

richardkiss:
> @richardkiss Please add a description to this PR. Thins that aren't obvious to me are:
> 
> 1. what are all the new dependencies for? Are they all transitive dependencies from bumping the crypto package version?
> 2. There appears to be an implementation of this in CLVM also, why?
> 3. what's "swbr" short for?

1. The new dependencies are likely all for the new command-line tool `swbr` (argument parsing, etc.)
2. I wrote this in python first as a proof-of-concept, then ported it to rust. The speed-up was surprisingly marginal, and in fact, `debug` rust is *slower* than python, which makes me think maybe there's some issue in the rust. With `--release`, it's several times faster though – but not 10x, which you might expect. I can't find a problem though, and the outputs are byte-for-byte identical in my tests.
3. `streaming with back-references`

richardkiss:
> It seems likely that we would get more savings by applying this compression at the db layer and leave the farmed blocks as they are.
> 
> * We wouldn't have to wrap the generator in the decompressor program (which is a lot of repetitive code)
> * We could decompress it in rust, which would be faster and use less memory
> 
> EDIT: I suppose it depends on the cost of decompresison in CLVM. it looks like the format was chosen to make it cheap to decompress.

This is intended to be used as a generator compressor only, by farmers, on a going-forward basis (since we commit the tree hash of the generator, not the generator's output), where we are restricted to writing a decompressor in clvm. Each generator must include the decompressor program. However, it's not actually very large, and it still seems to be quite a large win.

For the DB, probably `zstd` is what you want (although you might make it optional, since some file systems like zfs support `zstd` compression at the file system layer). I expect `zstd` will be able to further compress even generators with back-references.

trepca:
@richardkiss is this still being worked on?

richardkiss:
Duplicate with https://github.com/Chia-Network/clvm_rs/pull/222

PR #194, opened 2022-07-06
Simplify various things.

richardkiss:


src/run_program.rs
```diff
@@ -46,7 +46,7 @@ struct RunProgramContext<'a, D> {
     op_stack: Vec<Operation>,
 }
 
-impl<'a, 'h, D: Dialect> RunProgramContext<'a, D> {
+impl<'a, D: Dialect> RunProgramContext<'a, D> {
```
richardkiss, line 49: This was a `clippy` suggestion.

src/serialize.rs
```diff
@@ -1,57 +1,71 @@
...
 use crate::node::Node;
+use crate::reduction::EvalErr;
 
 const MAX_SINGLE_BYTE: u8 = 0x7f;
 const CONS_BOX_MARKER: u8 = 0xff;
 
-fn bad_encoding() -> std::io::Error {
-    Error::new(ErrorKind::InvalidInput, "bad encoding")
+fn bad_encoding() -> io::Error {
+    io::Error::new(ErrorKind::InvalidInput, "bad encoding")
```
richardkiss, line 12: Usually a bare `Error` refers to the `std::error::Error` trait, so we avoid importing `std::io::Error` here, and use `io::Error` (as per the recommendation for `std::io:Result` found [here](https://doc.rust-lang.org/std/io/type.Result.html).

src/serialize.rs
```diff
@@ -271,6 +281,11 @@ fn test_serialized_length_from_bytes() {
     );
 }
 
+#[cfg(test)]
+fn encode_size(f: &mut dyn Write, size: u64) -> io::Result<()> {
```
richardkiss, line 285: Some tests use this now-deprecated function, so we reimplement it in terms of a non-deprecated replacement.

src/serialize.rs
```diff
@@ -1,57 +1,80 @@
...
+fn write_atom_encoding_prefix(f: &mut dyn io::Write, atom: &[u8]) -> io::Result<()> {
+    let u8_0 = if !atom.is_empty() { atom[0] } else { 0 };
+    write_atom_encoding_prefix_with_size(f, u8_0, atom.len() as u64)
+}
+
+/// serialize an atom
+fn write_atom(f: &mut dyn io::Write, atom: &[u8]) -> io::Result<()> {
+    write_atom_encoding_prefix(f, atom)?;
+    f.write_all(atom)
+}
```
arvidn, line 74: these two functions should have unit tests, with the interesting edge cases.

arvidn (changes requested):
looks good apart from tests are missing

src/serialize.rs
```diff
@@ -271,6 +281,11 @@ fn test_serialized_length_from_bytes() {
     );
 }
 
+#[cfg(test)]
+fn encode_size(f: &mut dyn io::Write, size: u64) -> io::Result<()> {
```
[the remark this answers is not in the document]
richardkiss: Only by tests. There was a public function with this name that I removed above. This keeps the tests working.

src/serialize.rs
richardkiss: I've added tests (and removed a no-longer necessary intermediate function).

src/serialize.rs
```diff
@@ -1,57 +1,80 @@
...
+    io::Error::new(ErrorKind::InvalidInput, "internal error")
 }
 
-fn encode_size(f: &mut dyn Write, size: u64) -> std::io::Result<()> {
-    if size < 0x40 {
-        f.write_all(&[(0x80 | size) as u8])?;
+/// write the correct prefix for an atom of size `size` whose first byte is `atom_0`.
+/// If the atom is of size 0, use any placeholder first byte, as it's ignored anyway.
+
+fn write_atom_encoding_prefix_with_size(
```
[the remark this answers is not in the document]
richardkiss: I added them (and removed the `encode_size` tests).

.github/workflows/benchmark.yml
```diff
@@ -73,7 +73,7 @@ jobs:
         git clone https://github.com/Chia-Network/clvm_tools.git --branch=main --single-branch
         python -m pip install ./clvm_tools
         python -m pip install colorama
-        maturin develop -m wheel/Cargo.toml --release --cargo-extra-args="--features=openssl"
+        maturin develop -m wheel/Cargo.toml --release --features=openssl
```
richardkiss, line 76: Version 0.13 of `maturin` requires these changes.

src/serialize.rs
```diff
@@ -1,57 +1,74 @@
...
-use std::io::Write;
-use std::io::{Error, ErrorKind, SeekFrom};
+use std::io;
+use std::io::{Cursor, ErrorKind, Read, Seek, SeekFrom};
 
 use crate::allocator::{Allocator, NodePtr, SExp};
 use crate::node::Node;
+use crate::reduction::EvalErr;
 
 const MAX_SINGLE_BYTE: u8 = 0x7f;
```
arvidn, line 8: is this constant still used anywhere now?

arvidn (dismissed):

src/serialize.rs
richardkiss: I tried removing it to see and it is being used, in `impl From<EvalErr>`. It could arguably be moved to `src/reduction.rs`. Maybe I'll do this.

arvidn (approved):

PR #199, opened 2022-07-29
Use trait from `sha2` crate instead of home-grown.

richardkiss:


src/sha2.rs
```diff
@@ -43,13 +14,13 @@ pub struct Sha256 {
 impl Sha256 {
```
arvidn, line 14: shouldn't you specify the trait you're implementing here?

arvidn (approved):
looks good, but I though the reason was to actually implement the sha2 trait

src/sha2.rs
richardkiss: I thought `Sha256` was going to implement some simple trait, but it turns out to be a type of [`CoreWrapper`](https://docs.rs/digest/0.10.3/digest/core_api/struct.CoreWrapper.html), which implements a ton of things. So I'm just going to make the two options duck-type compatible by implement the three functions we need in `more_ops.rs`.

src/sha2.rs
```diff
@@ -1,37 +1,8 @@
...
-#[cfg(not(openssl))]
-impl Sha256 {
-    pub fn new() -> Sha256 {
-        Sha256 { ctx: Ctx::new() }
-    }
-    pub fn update(&mut self, buf: &[u8]) {
-        self.ctx.update(buf);
-    }
-    pub fn finish(self) -> [u8; 32] {
-        self.ctx.finalize().into()
```
arvidn, line 24: this PR is problematic. For example, the `finish()` function used to always return `[u8: 32]`, but now its return type depends on whether you use the openssl or `sha2` implementation. `sha2` returns a `GenericArrat<>` object that can be converted to a slice.

src/sha2.rs
```diff
@@ -1,37 +1,8 @@
 #[cfg(not(openssl))]
-use sha2::{Digest, Sha256 as Ctx};
+pub use sha2::{Digest, Sha256};
```
arvidn, line 2: this is another problem. In order to call `finalize()` on the `sha2::Sha256` object, the `Hasher` trait needs to be in scope. i.e. users need to import it. However, this won't be available when using the openssl backend.

arvidn:
@richardkiss I'm tempted to revert this PR and maybe *just* change the names to match `sha2`

PR #200, opened 2022-08-02
add function to compute the tree-hash

arvidn:
from a CLVM structure in serialized form

The second commit is unrelated. it's just running `cargo fmt` on the fuzz targets.

### DIFF
--- a/src/serialize.rs
+++ b/src/serialize.rs
@@ -231,6 +231,171 @@ pub fn serialized_length_from_bytes(b: &[u8]) -> io::Result<u64> {
     Ok(f.position())
 }
 
+use crate::sha2::{Digest, Sha256};
+
+fn hash_atom(buf: &[u8]) -> [u8; 32] {
+    let mut ctx = Sha256::new();
+    ctx.update(&[1_u8]);
+    ctx.update(buf);
+    ctx.finalize().into()
+}
+
+fn hash_pair(left: &[u8; 32], right: &[u8; 32]) -> [u8; 32] {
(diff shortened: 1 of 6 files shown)


richardkiss (approved):
Seems reasonable.

src/serialize.rs
```diff
@@ -231,6 +231,171 @@ pub fn serialized_length_from_bytes(b: &[u8]) -> io::Result<u64> {
     Ok(f.position())
 }
 
+use crate::sha2::{Digest, Sha256};
```
richardkiss, line 234: Does `clippy` complain about this not being at the top? Seems like it should.

src/serialize.rs
```diff
@@ -231,6 +231,171 @@ pub fn serialized_length_from_bytes(b: &[u8]) -> io::Result<u64> {
...
+                    values.push(hash_atom(&[]));
+                } else if b[0] <= MAX_SINGLE_BYTE {
+                    values.push(hash_atom(&b));
+                } else {
+                    let blob_size = decode_size(f, b[0])?;
+                    let blob = &f.get_ref()[f.position() as usize..];
+                    if (blob.len() as u64) < blob_size {
+                        return Err(bad_encoding());
+                    }
+                    f.set_position(f.position() + blob_size);
```
richardkiss, line 279: You could use `.seek` here, possibly slightly more clear. But maybe not. Probably no speed difference.

src/serialize.rs
arvidn: it doesnt

PR #202, opened 2022-08-04
First crack at `CLVMTree`

richardkiss:


src/serialize.rs
```diff
@@ -88,43 +88,165 @@ pub fn node_to_stream(node: &Node, f: &mut dyn io::Write) -> io::Result<()> {
...
+            },
+        }
+    }
+    Ok(r)
+}
+
 /// decode the length prefix for an atom. Atoms whose value fit in 7 bits
 /// don't have a length prefix, so those should be handled specially and
 /// never passed to this function.
-fn decode_size(f: &mut dyn io::Read, initial_b: u8) -> io::Result<u64> {
```
arvidn, line 94: I missed that this turned into `dyn` at some point. changing it back seems like a good idea in itself

richardkiss:
Duplicate

PR #203, opened 2022-08-09
Wasm update

ChiaMineJP:
Trying to make wasm version of `run_chia_program`.


wasm/src/api.rs
```diff
@@ -2,23 +2,35 @@ use wasm_bindgen::prelude::*;
...
-// works at all.
-//
-// TODO: replace the below with something more robust and feature-filled
+#[wasm_bindgen]
+pub struct Flag;
+
+#[wasm_bindgen]
+impl Flag {
+    #[wasm_bindgen(getter)]
+    pub fn no_neg_div(&self) -> u32 { _no_neg_div }
```
arvidn, line 24: technically we don't need this anymore. This was made mandatory in the soft-fork, and it had not been exercised on the chain before the soft-fork.

wasm/src/api.rs
```diff
@@ -2,23 +2,35 @@ use wasm_bindgen::prelude::*;
...
+
+    #[wasm_bindgen(getter)]
+    pub fn no_unknown_ops(&self) -> u32 { _no_unknown_ops }
```
arvidn, line 27: you may want to bind the `mempool` flag as well (which is a combination of all settings in mempool-mode`

Also, are you sure you want `&self` in these getters? it's a bit odd since `Flag` is just used as a namespace , with no state.

wasm/src/api.rs
```diff
@@ -40,3 +53,82 @@ pub fn run_clvm(program: &[u8], args: &[u8]) -> Vec<u8> {
...
+        program,
+        args,
+        max_cost,
+        None,
+    );
+    match r {
+        Ok(reduction) => {
+            let cost = reduction.0;
+            let node_bytes = node_to_bytes(
+                &Node::new(&allocator, reduction.1)
```
arvidn, line 92: this is vulnerable to a DoS attack. It's possible to (cheaply) produce a return value that does not use a lot of memory in its internal clvm_rs representation, but that balloons into using a *lot* of memory when serialized. As long as only trusted programs are run, this is probably fine. There should probably be some kind of warning label though.

wasm/src/api.rs
```diff
@@ -40,3 +53,82 @@ pub fn run_clvm(program: &[u8], args: &[u8]) -> Vec<u8> {
...
+/**
+ * Return serialized result of clvm program with cost
+ *
+ * cost will be available at the first 8 bytes of returned Vec<u8>(=Uint8Array).
+ * bytes of node will be available from 8th byte offset of returned Vec<u8>.
+ *
+ * @example
+ * const result = run_chia_program(...); // Uint8Array
+ * const cost = vec_u8_to_u64(result.subarray(0, 8)) // BigInt;
+ * const serialized_node = result.subarray(8); // Uint8Array
```
arvidn, line 67: this sounds a bit awkward. you can't return tuples back to javascript?

If you end up concatenating, you may want to be careful with how you do it. Right now you end up copying `node_bytes` into a new buffer. `nodes_bytes` is potentially quite large. You could for instance insert the 8 bytes of cost at the beginning instead. It might end up copying still, but it might also `memmov()` the bytes if there's enough capacity in the vector.

wasm/src/api.rs
```diff
@@ -40,3 +53,82 @@ pub fn run_clvm(program: &[u8], args: &[u8]) -> Vec<u8> {
...
+                u64_to_vec_u8(cost),
+                node_bytes,
+            ].concat()
+        },
+        Err(_eval_err) => format!("{:?}", _eval_err).into(),
+    }
+}
+
+#[wasm_bindgen]
+pub fn u64_to_vec_u8(num: u64) -> Vec<u8> {
```
arvidn, line 105: instead of this function, you could just use:

```
cost.to_be_bytes().to_vec()
```

wasm/src/api.rs
```diff
@@ -40,3 +53,82 @@ pub fn run_clvm(program: &[u8], args: &[u8]) -> Vec<u8> {
...
+        ((num & 0x0000ff00) >> 8) as u8,
+        ((num & 0x000000ff) >> 0) as u8,
+    ];
+
+    u8_vec
+}
+
+#[wasm_bindgen]
+pub fn vec_u8_to_u64(nums: &[u8]) -> u64 {
+    let u64_value =
```
arvidn, line 122: this function could be:
```
u64::from_be_bytes(nums).expect("not enough bytes")
```

Also, there's no need to create a variable to return, just omit the `;` at the end.

arvidn (changes requested):
I think the DoS issue should be highlighted somehow at least

ChiaMineJP:
Thank you for reviewing @arvidn !
I'll check what you pointed!

wasm/src/api.rs
ChiaMineJP: > you may want to bind the mempool flag as well (which is a combination of all settings in mempool-mode`

It's weird that there are no `MEMPOOL_MODE` is available in the code at `main` branch.
wheel also doesn't expose `mempool` flag. Only `NO_NEG_DIV` and `NO_UNKNOWN_OPS` is exported.

Did you mean `mempool` as this one?
https://github.com/Chia-Network/clvm_rs/pull/163/files#diff-fab6a71cbebf342f08d70eab416d15f1ce5f4d2cb159a0de994bcf890868864eR13

-------

> Also, are you sure you want &self in these getters? it's a bit odd since Flag is just used as a namespace , with no state.

There are a number of restrictions `wasm-pack` imposes.
`wasm-bindgen` only supports to export `function`, `struct`, `impl`, `enum` and it does not allow to export just a variable to JS.
I chose `struct` and `impl` to convey the flag values to JS world.

## Current implementation
Export `impl` like this.
```rust
#[wasm_bindgen]
impl Flag {
    #[wasm_bindgen(getter)]
    pub fn no_neg_div(&self) -> u32 { _no_neg_div }

    #[wasm_bindgen(getter)]
    pub fn no_unknown_ops(&self) -> u32 { _no_unknown_ops }
}
```
The above is compiled to JavaScript like this:
```javascript
export class Flag {
    ...
    /**
    * @returns {number}
    */
    get no_neg_div() {
        var ret = wasm.flag_no_neg_div(this.ptr);
        return ret >>> 0;
    }
    /**
    * @returns {number}
    */
    get no_unknown_ops() {
        var ret = wasm.flag_no_unknown_ops(this.ptr);
        return ret >>> 0;
    }
}
```
This way, you can get the values in JS like this:
`new Flag().no_neg_div` / `new Flag().no_unknown_ops`
Umm. It doesn't look good.

## Case 1
Export `impl` like this. (Removed `&self` from method argument)
```rust
#[wasm_bindgen]
impl Flag {
    #[wasm_bindgen(getter)]
    pub fn no_neg_div() -> u32 { _no_neg_div }

    #[wasm_bindgen(getter)]
    pub fn no_unknown_ops() -> u32 { _no_unknown_ops }
}
```
When running `wasm-pack build --release`, I encountered an error:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `2`,
 right: `1`', crates\cli-support\src\js\binding.rs:165:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: Running the wasm-bindgen CLI
Caused by: failed to execute `wasm-bindgen`: exited with exit code: 101
```

## Case 2
Export `impl` like this. (Removed `&self` and `#[wasm_bindgen(getter)]` => `#[wasm_bindgen]`)
```rust
#[wasm_bindgen]
impl Flag {
    #[wasm_bindgen]
    pub fn no_neg_div() -> u32 { _no_neg_div }

    #[wasm_bindgen]
    pub fn no_unknown_ops() -> u32 { _no_unknown_ops }
}

```
The above is compiled like this:
```javascript
export class Flag {
    ...
    /**
    * @returns {number}
    */
    static no_neg_div() {
        var ret = wasm.flag_no_neg_div();
        return ret >>> 0;
    }
    /**
    * @returns {number}
    */
    static no_unknown_ops() {
        var ret = wasm.flag_no_unknown_ops();
        return ret >>> 0;
    }
}
```
This way, you can get the values in JS like this:
`Flag.no_neg_div()` / `Flag.no_unknown_ops()`.
It still does not look good but this is a little bit better than the current implementation.

wasm/src/api.rs
ChiaMineJP: In my environment, this didn't work.
```rust
#[wasm_bindgen]
pub fn vec_u8_to_u64(nums: &[u8]) -> u64 {
    u64::from_be_bytes(nums).expect("not enough bytes")
}
```

So I tried this one:
```rust
#[wasm_bindgen]
pub fn vec_u8_to_u64(nums: &[u8]) -> u64 {
    u64::from_be_bytes(nums.try_into().expect("not enough bytes"))
}
```

wasm/src/api.rs
ChiaMineJP: Thanks! It's working!

wasm/src/api.rs
ChiaMineJP: There are no built-in `tuple` type in JavaScript so we need to use JavaScript `Array` as a tuple.
Creating and returning JavaScript's `Array` in Rust/Wasm code seems heavy.
Reference: https://rustwasm.github.io/docs/wasm-bindgen/reference/arbitrary-data-with-serde.html

If we want to return a tuple like `[cost, node_bytes]`, the type of returning tuple would be `[u64, vec<u8>]` and the composite type like this seems requiring JSON serialization/deserialization with `serde` rust package.
So instead of doing such a slow task, I preferred a bit painful but performant way to get cost/node_bytes.

wasm/src/api.rs
arvidn: > It's weird that there are no MEMPOOL_MODE is available in the code at main branch.
> wheel also doesn't expose mempool flag. Only NO_NEG_DIV and NO_UNKNOWN_OPS is exported.
> 
> Did you mean mempool as this one?

Ah, right. Some of the mempool-mode treatment is in the parsing of the generator output too.

wasm/src/api.rs
arvidn: I agree that JSON would probably not carry its weight here.

wasm/src/api.rs
ChiaMineJP: As for the DoS attack you mentioned. could you correct me if my understanding is wrong?
My understanding:
- `SExp` instance on clvm_rs code first reserves memory by `Allocator`. (1MB maybe?)
- This reserved memory is not actually consumes RAM
- But when `node_to_bytes` is called, reserved memory is "materialized" and consumes RAM

I'm still looking for the actual problem and fix for now.

arvidn: no, the issue can be illustrated by this program:
```
(mod ()
(defun make_list (leaf n)
  (if n (c leaf (make_list leaf (- n 1))) leaf)
)

(make_list (lsh (q . 1) (q . 65535)) (q . 1000))
)
```

This creates one (somewhat) large atom, then builds a list of 1000 elements with that one atom. In memory, there's only a single atom, that's referenced by every cons-box/pair in the list. But if you serialize the list, you'll make a copy of that atom for every element in the list. i.e. you blow up memory usage.

The way we solved this in the interface to python is that we don't return a the resulting tree as actual python `SExp` objects, we return a native rust object that provides a view into the nodes allocated in the corresponding allocator. 

And in the critical parts we don't even return that, we parse the generator program output directly in rust and validates some of it immediately.

ChiaMineJP: I'm investigating whether `wasm-pack` can handle something like `LazyNode` in Python wheel.
As for the DoS issue, I think even after implementing `LazyNode` in wasm, the DoS attack can be effective everytime `serialization` is required.
So implementing `LazyNode` seems to just lower the chance of the attack and not completely preventing it.

arvidn: right, whoever serializes a structure has to trust it. I don't believe we ever do though, we serialize *programs* we don't necessarily trust, not their output.

Also, aren't all programs we run in the wallet trusted? Maybe the wasm bindings are meant for other applications than wallet.

ChiaMineJP: Oh I'm completely convinced. We serialize _program_, not its _output_.
The last remaining issue is to implement wasm version of `LazyNode`. I'll continue investigaing on that.

ChiaMineJP: @arvidn 
I've finally managed to implement wasm version of `LazyNode`.

This is a sample test code.
(You need to have NodeJS installed)

Prepare
```bash
git clone https://github.com/Chia-Network/clvm_rs
cd clvm_rs
wasm-pack build --release --target=nodejs wasm
node
```
(The code below is in the context of NodeJS)
```javascript
const wasm = require("./wasm/pkg/clvm_wasm.js");

// (q . 127)
prog = Uint8Array.from(Buffer.from("ff017f", "hex"));
arg = new Uint8Array([0]);

result = wasm.run_chia_program(prog, arg, BigInt("100000000000"), 0);
cost = result[0];
sexp = result[1];
console.log(cost, sexp, sexp.atom.toString() === "127");
```

ChiaMineJP:
> it would be great to exercise the wasm build in nodejs in CI, to ensure it works and won't break in the future

I'll add a CI action to run compiled wasm on nodejs context.

ChiaMineJP:
@arvidn I've added simple test code which confirms that wasm runs successfully on NodeJS runtime context.
`wasm/tests/index.js` is the one.
This test also runs on CI (`build_create-and-npm.yml`). If the test fails, CI also fails.

arvidn:
> @arvidn I've added simple test code which confirms that wasm runs successfully on NodeJS runtime context. `wasm/tests/index.js` is the one. This test also runs on CI (`build_create-and-npm.yml`). If the test fails, CI also fails.

great!

wasm/src/api.rs
```diff
@@ -1,24 +1,40 @@
...
+    pub fn no_neg_div() -> u32 { _no_neg_div }
+
+    #[wasm_bindgen]
+    pub fn no_unknown_ops() -> u32 { _no_unknown_ops }
+}
+
+#[wasm_bindgen]
+pub fn serialized_length(program: &[u8]) -> u64
+{
+    serialized_length_from_bytes(program).unwrap()
```
arvidn, line 36: if `program` has invalid serialization, this will take down the whole process. that seems risky, rather than propagating the error

arvidn (dismissed):

wasm/src/api.rs
ChiaMineJP: @arvidn I've updated the code as you requested. Please confirm when you have time.

arvidn (approved):
sorry for the delay. looks great!

PR #212, opened 2022-10-21
add unit tests for Allocator

arvidn:


src/allocator.rs
```diff
@@ -183,3 +183,154 @@ impl Allocator {
         -2
     }
 }
+
+#[test]
+fn test_null() {
+    let a = Allocator::new();
+    assert_eq!(a.atom(a.null()), b"");
+    assert_eq!(
```
aqk, line 191: This second assert would be a little easier to read as two statements.

aqk (dismissed):
lgtm. Might want to add equality for Pair(Pair, Pair) - pair with pair child on left and right.

aqk (approved):

PR #215, opened 2022-10-31
add option to limit heap usage

arvidn:


src/allocator.rs
```diff
@@ -74,12 +102,12 @@ impl Allocator {
...
     pub fn new_atom(&mut self, v: &[u8]) -> Result<NodePtr, EvalErr> {
         let start = self.u8_vec.len() as u32;
-        if ((u32::MAX - start) as usize) < v.len() {
+        if (self.heap_limit - start as usize) < v.len() {
             return err(self.null(), "out of memory");
         }
         self.u8_vec.extend_from_slice(v);
         let end = self.u8_vec.len() as u32;
-        if self.atom_vec.len() == i32::MAX as usize {
+        if self.atom_vec.len() == self.atom_limit {
```
prozacchiwawa, line 110: Not for now necessarily, but there'd be no reason to have a separate atom limit (unless you want it for other kinds of profiling) if empty atoms all pointed to the same empty string space like in coff and elf string tables.  there'd be no way to create a new atom without taking heap_limit unless it was a clone of the empty atom which would be free.

src/chia_dialect.rs
```diff
@@ -103,4 +107,12 @@ impl Dialect for ChiaDialect {
     fn apply_kw(&self) -> &[u8] {
         &[2]
     }
+
+    fn val_stack_limit(&self) -> usize {
+        if (self.flags & LIMIT_HEAP) != 0 {
+            10000000
```
prozacchiwawa, line 113: seems low for what many coins do with treehashing.

src/allocator.rs
arvidn: yeah, that's a good point. I was thinking about that when working on this patch too. We don't implement this check yet though, to ensure that all empty atom outputs reuse the one canonical null atom. Same thing could be done for "1".

src/chia_dialect.rs
prozacchiwawa, line 113: something that might be useful here is to take some big singleton inner puzzle and do treehash on it under this limit.  i'm thinking back to the recent performance improvement in clvm that might change the order of retirements from the stack.  i don't definitely know there'll be a lot of values left on the stack before the final sha256, but it'd be good to have a smoke test that it generally isn't blowing up with the limit on.

arvidn: I tested the whole chain with this limit, so we don't have any puzzles this large that have run on the chain so far at least

prozacchiwawa (dismissed):

src/chia_dialect.rs
```diff
@@ -103,4 +107,12 @@ impl Dialect for ChiaDialect {
     fn apply_kw(&self) -> &[u8] {
         &[2]
     }
+
+    fn val_stack_limit(&self) -> usize {
```
aqk, line 111: Is there a relationship between `val_stack_limit` and atom(pair/heap)_limit that could be expressed in code?

arvidn: I prefer to specify the actual limits, than trying to be clever and set a single limit that then translates to various limits

src/chia_dialect.rs
```diff
@@ -18,6 +18,10 @@ pub const NO_NEG_DIV: u32 = 0x0001;
 // (otherwise they are no-ops with well defined cost)
 pub const NO_UNKNOWN_OPS: u32 = 0x0002;
 
+// When set, limits the number of atom-bytes allowed to be allocated, as well as
+// the number of pairs
+pub const LIMIT_HEAP: u32 = 0x0004;
```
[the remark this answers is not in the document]
arvidn: I wanted to keep this patch as small as possible, and not affect existing possible uses of the `new()` constructor. Also, it's non obvious that one limit is `i32::MAX` and one is `i32::MAX - 1` (because we use the negative value space of a signed integer, which is larger)

src/chia_dialect.rs
aqk: Oh sure, I meant more like enforcing e.g. `assert heap_size > atom_limit * MIN_ATOM_SIZE` types of checks.

aqk (approved):

PR #219, opened 2022-11-04
optimize parsing of serialized CLVM by saving a copy of atoms

arvidn:
it saves a heap allocation and copy of the parsed atom when parsing

### DIFF
--- a/src/serialize.rs
+++ b/src/serialize.rs
@@ -127,6 +127,7 @@ fn decode_size(f: &mut dyn io::Read, initial_b: u8) -> io::Result<u64> {
     Ok(v)
 }
 
+#[repr(u8)]
 enum ParseOp {
     SExp,
     Cons,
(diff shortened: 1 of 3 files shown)


src/serialize.rs
```diff
@@ -158,12 +159,13 @@ pub fn node_from_stream(allocator: &mut Allocator, f: &mut Cursor<&[u8]>) -> io:
...
+                    let pos = f.position() as usize;
+                    if f.get_ref().len() < pos + blob_size as usize {
                         return Err(bad_encoding());
                     }
-                    let mut blob: Vec<u8> = vec![0; blob_size as usize];
-                    f.read_exact(&mut blob)?;
-                    values.push(allocator.new_atom(&blob)?);
+                    let blob = &f.get_ref()[pos..(pos + blob_size as usize)];
+                    f.seek(SeekFrom::Current(blob_size as i64))?;
+                    values.push(allocator.new_atom(blob)?);
```
prozacchiwawa, line 168: nice, makes sense to hold just the slice ref since it's going to be copied into allocator via new_atom anyway.

prozacchiwawa (approved):

PR #221, opened 2022-11-07
turn a dynamic-dispatch call to a static call in the clvm deserializer

arvidn:
for improved efficiency

### DIFF
--- a/src/serialize.rs
+++ b/src/serialize.rs
@@ -91,7 +91,7 @@ pub fn node_to_stream(node: &Node, f: &mut dyn io::Write) -> io::Result<()> {
 /// decode the length prefix for an atom. Atoms whose value fit in 7 bits
 /// don't have a length prefix, so those should be handled specially and
 /// never passed to this function.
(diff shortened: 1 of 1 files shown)


src/serialize.rs
```diff
@@ -91,7 +91,7 @@ pub fn node_to_stream(node: &Node, f: &mut dyn io::Write) -> io::Result<()> {
 /// decode the length prefix for an atom. Atoms whose value fit in 7 bits
 /// don't have a length prefix, so those should be handled specially and
 /// never passed to this function.
-fn decode_size(f: &mut dyn io::Read, initial_b: u8) -> io::Result<u64> {
+fn decode_size(f: &mut Cursor<&[u8]>, initial_b: u8) -> io::Result<u64> {
```
richardkiss, line 94: What do you think of this instead:

```suggestion
fn decode_size<R: io::Read>(f: &mut R, initial_b: u8) -> io::Result<u64> {
```

Requires fewer changes and should be just as fast, since it's a template. I may want to use the `Read` trait in future.

richardkiss:
@223 includes this change, so you can close this once it merges.

PR #222, opened 2022-11-07
First crack at deserialization with back-refs.

richardkiss:


src/serde/compressor.rs
```diff
@@ -0,0 +1,73 @@
+use crate::allocator::{Allocator, NodePtr};
```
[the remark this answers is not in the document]
arvidn: I don't have a strong opinion. it seems nice to separate serialization and deserialization into separate files. The name "serde" risks making people think we're actually implementing the serde crate's traits.

It's also not obvious to me that the compression belongs in the clvm crate

src/serde/compressor.rs
```diff
@@ -0,0 +1,73 @@
...
+    node_to_bytes(&Node::new(allocator, decompression_program_ptr))
+}
+
+fn wrap_atom_with_decompression_program(
+    allocator: &mut Allocator,
+    node_ptr: NodePtr,
+) -> Result<NodePtr, std::io::Error> {
+    let apply_node = allocator.new_atom(&[2])?;
+    let quote_node = allocator.one();
+    let serialized_backrefs_program = include_bytes!("deserialize_w_backrefs.bin");
```
arvidn, line 45: I think CI needs to ensure this checked-in binary matches the output of compiling the `.cl` file.

src/serde/compressor.rs
```diff
@@ -0,0 +1,73 @@
...
+    let compressed_block = allocator.new_pair(quote_node, node_ptr)?;
+    let program = allocator.new_pair(quote_node, program)?;
+    let list = allocator.null();
+    let list = allocator.new_pair(compressed_block, list)?;
+    let list = allocator.new_pair(program, list)?;
+    let list = allocator.new_pair(apply_node, list)?;
+    Ok(list)
+}
+
+pub fn decompress(allocator: &mut Allocator, blob: &[u8]) -> Result<NodePtr, std::io::Error> {
```
arvidn, line 59: these functions don't have any unit tests. I would also like to see a fuzzer that tests round-tripping of arbitrary structures

src/serde/ser_br.rs
```diff
@@ -0,0 +1,84 @@
...
+
+pub fn node_to_stream_backrefs(node: &Node, f: &mut dyn io::Write) -> io::Result<()> {
+    let allocator = node.allocator;
+    let mut read_op_stack: Vec<ReadOp> = vec![ReadOp::Parse];
+    let mut write_stack: Vec<NodePtr> = vec![node.node];
+
+    let mut read_cache_lookup = ReadCacheLookup::new();
+
+    let mut thc = ObjectCache::new(allocator, treehash);
+    let mut slc = ObjectCache::new(allocator, serialized_length);
```
arvidn, line 36: is there a good reason not to make `ObjectCache` a non-template and have its values be both the hash and the serialized lenght?
That would seem to be both simpler code and more efficient

src/serde/ser_br.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// these test cases were produced by:
+
+// from chia.types.blockchain_format.program import Program
+// a = Program.to(...)
+// print(bytes(a).hex())
+// print(a.get_tree_hash().hex())
+
+pub fn node_to_stream_backrefs(node: &Node, f: &mut dyn io::Write) -> io::Result<()> {
```
arvidn, line 28: I think you should make `f` a template instead of using dynamic dispatch. I think you can do that like this:
```
pub fn<S: io::Write> node_to_stream_backrefs(node: &Node, f: &mut S) -> io::Result<()> {
```

arvidn: I proposed this, related, change: https://github.com/Chia-Network/clvm_rs/pull/221

wheel/src/api.rs
```diff
@@ -1,19 +1,31 @@
...
     __pyo3_get_function_serialized_length,
 };
-use clvmr::chia_dialect::{NO_NEG_DIV, NO_UNKNOWN_OPS, LIMIT_HEAP};
+use clvmr::chia_dialect::{LIMIT_HEAP, NO_NEG_DIV, NO_UNKNOWN_OPS};
+use clvmr::serde::compressor::reserialize;
 
 pub const MEMPOOL_MODE: u32 = NO_NEG_DIV | NO_UNKNOWN_OPS | LIMIT_HEAP;
 
+#[pyfunction]
+fn reserialize_with_backreferences<'p>(
```
arvidn, line 16: you don't think "compress" would be a better name?

src/serde/read_cache_lookup.rs
```diff
@@ -0,0 +1,232 @@
...
+            paths.sort();
+            paths.truncate(1);
+            paths.pop()
+        } else {
+            None
+        }
+    }
+}
+
+fn reversed_path_to_vec_u8(path: &[u8]) -> Vec<u8> {
```
arvidn, line 182: could you add a comment explaining what this function does?

src/serde/read_cache_lookup.rs
```diff
@@ -0,0 +1,232 @@
...
+    parent_lookup: HashMap<Bytes32, Vec<(Bytes32, u8)>>,
+}
+
+impl Default for ReadCacheLookup {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl ReadCacheLookup {
```
arvidn, line 43: This class needs unit tests

src/serde/object_cache.rs
```diff
@@ -0,0 +1,154 @@
...
+                    }
+                }
+            }
+        }
+    }
+}
+
+/// calculate the standard `sha256tree` has for a node
+
+pub fn treehash(
```
arvidn, line 105: this function needs unit tests

src/serde/object_cache.rs
```diff
@@ -0,0 +1,154 @@
...
+                .map(|right_value| hash_blobs(&[&[2], left_value, right_value])),
+        },
+        SExp::Atom(atom_buf) => Some(hash_blobs(&[&[1], allocator.buf(&atom_buf)])),
+    }
+}
+
+/// calculate the serialized length (without backrefs) of a node. This is used
+/// to check if using backrefs is actually smaller.
+
+pub fn serialized_length(
```
arvidn, line 124: this function needs unit tests

src/serde/object_cache.rs
```diff
@@ -0,0 +1,154 @@
...
+    /// If a pair is passed and one of the children does not have its T value cached
+    /// in `ObjectCache` yet, return `None` and f will be called with each child in turn.
+    /// Don't recurse in f; that's part of the point of this function.
+    f: CachedFunction<T>,
+}
+
+/// turn a `NodePtr` into a `usize`. Positive values become even indices
+/// and negative values become odd indices.
+
+fn node_to_index(node: &NodePtr) -> usize {
```
arvidn, line 30: I'd like to see unit tests of this function.

src/serde/object_cache.rs
```diff
@@ -0,0 +1,154 @@
...
+    let node = *node;
+    if node < 0 {
+        (-node - node - 1) as usize
+    } else {
+        (node + node) as usize
+    }
+}
+
+impl<'a, T: Clone> ObjectCache<'a, T> {
```
arvidn, line 39: this class should have unit tests, but I also have a feeling it could be simplified (made more concrete) to fit the one use we have for it.

arvidn (changes requested):
The movements of code into different files is not controversial. That could land as a separate PR (which might also make it easier to focus on the compression code in this PR).

Apart from the comments I've made here, I think the ones I made on your other PR are still pertinent: https://github.com/Chia-Network/chia_rs/pull/55

src/serde/ser_br.rs
richardkiss: I've found myself writing this pattern many times while developing clvm, so I thought it would be good to create it. I suppose I could build both the tree hash and length caches simultaneously. I am pretty skeptical it would provide much of a performance boost.

richardkiss:
> The movements of code into different files is not controversial. That could land as a separate PR (which might also make it easier to focus on the compression code in this PR).
> 
> Apart from the comments I've made here, I think the ones I made on your other PR are still pertinent: [Chia-Network/chia_rs#55](https://github.com/Chia-Network/chia_rs/pull/55)

Okay, let's do that.

src/serde/object_cache.rs
richardkiss: Added tests. As I've noted previously, the complexity came out of my frustration having written this code several times for different things.

cl/deserialize_w_backrefs.cl
```diff
@@ -0,0 +1,109 @@
+(mod input
```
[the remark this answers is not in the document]
richardkiss: I've removed this code with the intent to put it into `chia_rs`.

src/serde/compressor.rs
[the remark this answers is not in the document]
richardkiss: I dropped the `compression`-named API in favour of moving it to `chia_rs`, and left `*_br` (for "back-refs") api in `clvm_rs`.

src/serde/compressor.rs
richardkiss: Dropped.

src/serde/ser_br.rs
richardkiss: This has been turned into a template, so it's the best of both worlds (at the cost of possibly slower compilation).

src/serde/read_cache_lookup.rs
richardkiss: Tests added.

src/serde/object_cache.rs
richardkiss: Done

src/serde/object_cache.rs
richardkiss: Done

src/serde/read_cache_lookup.rs
richardkiss: Done.

src/serde/object_cache.rs
```diff
@@ -0,0 +1,268 @@
...
+        let index = node_to_index(node);
+        if index >= self.cache.len() {
+            self.cache.resize(index + 1, None);
+        }
+        self.cache[index] = Some(v)
+    }
+
+    /// calculate the function's value for the given node, traversing uncached children
+    /// as necessary
+    fn calculate(&mut self, root_node: &NodePtr) {
```
arvidn, line 77: it doesn't look like this would end up being recursive. Is that right?
The functions you instantiate this template with has to make sure to only call `get_from_cache()` and never `calculate()` or `get_or_calculate()`. If the function does, it would end up being recursive.

It may not be obvious to future readers of this code. I think it would be good to have a test case that would fail in case either of the two function you use (tree_hash and serialized_length) end up being recursive. i.e. a stress test with a very deep tree structure, or a long list perhaps.

arvidn (commented):
I think it would be good to address the clippy warnings

src/serde/ser_br.rs
richardkiss: > I think it would be good to address the clippy warnings

Done

richardkiss: > as it is now, you end up computing tree hashes twice, right? That would be my main concern. In the light wallet tree hash computations shows up in the profile

If you use `ObjectCache` to calculate the hash tree of an clvm cons, it should calculate every subtree's hash exactly once. It might make sense later to provide api to pre-populate known hashes for particular subtrees. But I don't think it will be useful for compression, which is the first application here.

src/serde/object_cache.rs
```diff
@@ -0,0 +1,268 @@
...
+    check("ffff0102ff0304", 2); // ((1 . 2) . (3 . 4))
+    check("ff01ff02ff03ff04ff05ff0680", 6); // (1 2 3 4 5 6)
+}
+
+#[test]
+fn test_treehash() {
+    let check = |a, b| check_cached_function(a, Bytes32::from_hex(&b).unwrap(), treehash);
+    check(
+        "ff83666f6f83626172",
+        "c518e45ae6a7b4146017b7a1d81639051b132f1f5572ce3088a3898a9ed1280b",
```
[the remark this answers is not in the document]
richardkiss: I think I wrote this before I decided I needed a `deserialize_wbr` function. It would be a lot easier to simply do a round-trip and make sure it's idempotent. I'll make that change.

Note that you can use `opc -H (foo . bar)` to quickly get a tree hash (and in fact, that's what I did).

src/serde/object_cache.rs
```diff
@@ -0,0 +1,268 @@
...
+        } else {
+            None
+        }
+    }
+
+    /// set the cached value for a node
+    fn set(&mut self, node: &NodePtr, v: T) {
+        let index = node_to_index(node);
+        if index >= self.cache.len() {
+            self.cache.resize(index + 1, None);
```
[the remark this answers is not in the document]
richardkiss: For consensus, naw, because we're charging reasonable costs for the decompression, and the memory usage is pretty trivial compared to DOS attack.

In general, I think memory usage for legitimate actions is going to be dwarfed by DOS attacks. In other words, if we're doing a real, useful chia contract, memory and CPU usage will be quite reasonable. But DOS attacks can use as much as they want and we can't really do much about it (beside soft forking in further limits in future).

src/serde/object_cache.rs
```diff
@@ -0,0 +1,268 @@
...
+    match allocator.sexp(node) {
+        SExp::Pair(left, right) => match cache.get_from_cache(&left) {
+            None => None,
+            Some(left_value) => cache
+                .get_from_cache(&right)
+                .map(|right_value| 1 + left_value + right_value),
+        },
+        SExp::Atom(atom_buf) => {
+            let buf = allocator.buf(&atom_buf);
+            let lb = buf.len();
```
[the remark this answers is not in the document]
richardkiss: We still need `buf[0]` a few lines later, so there's not really any getting around calling `.buf`. I can make this change, but not sure it's a win.

src/serde/object_cache.rs
richardkiss: Correct, it uses that trick we use in many places where we build our own stack of objects on the heap to eliminate the need to use the thread's stack for recursion (and the resulting possibility to blow out that stack).

I'll add a test with a very long list.

src/serde/de_br.rs
```diff
@@ -0,0 +1,144 @@
...
+const BACK_REFERENCE: u8 = 0xfe;
+const CONS_BOX_MARKER: u8 = 0xff;
+
+#[repr(u8)]
+enum ParseOp {
+    SExp,
+    Cons,
+}
+
+fn parse_path(f: &mut Cursor<&[u8]>) -> io::Result<Vec<u8>> {
```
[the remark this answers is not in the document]
richardkiss: Oh yeah, I guess I didn't do that because I've always been a bit reticent to use `Cursor` api rather than `Read` api, as I just feel like restoring `Read` api might be useful someday. But I can cross that bridge if I come to it. I'll use your idea and make your change. I think I can actually unify the two functions too.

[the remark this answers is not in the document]
richardkiss: Done in @a21a6a223d19d0d9923bf7fe578bac74052a9de0

src/serde/object_cache.rs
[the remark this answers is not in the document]
richardkiss: Done in @f70809ee3f33dce00327eb9a97139082bd1dc54d

src/serde/object_cache.rs
richardkiss: Done

src/serde/de_br.rs
```diff
@@ -0,0 +1,144 @@
...
+            return Err(bad_encoding());
+        }
+        let blob = &f.get_ref()[pos..(pos + blob_size as usize)];
+        f.seek(SeekFrom::Current(blob_size as i64))?;
+        Ok(blob.into())
+    }
+}
+
+/// deserialize a clvm node from a `std::io::Cursor`
+pub fn node_from_stream_backrefs(
```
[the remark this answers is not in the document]
richardkiss: I created a target for `node_from_bytes_backrefs`, then added code to do a round-trip and it took just a few seconds for the fuzzer to find a crash. So I have more work to do.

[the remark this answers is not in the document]
richardkiss: I located & fixed the main fuzzing issue, which had to do with unsigned integer operations overflowing 6e863557052e51c9b6c06930881fa698f7b5c0d9

src/serde/ser_br.rs
arvidn: I believe I was under the impression the tree-hash was the key in the cache, but it's not `NodePtr`

arvidn (commented):
parse_atom ended up pretty nice!
I think this all looks good now! My only remaining concern is the round-trip fuzzer. It doesn't look like the current fuzzer would get you very good coverage for round-trip tests

fuzz/fuzz_targets/deserialize_br.rs
```diff
@@ -0,0 +1,26 @@
...
+use clvmr::serde::node_to_bytes_backrefs;
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    let mut allocator = Allocator::new();
+    let program = match node_from_bytes_backrefs(&mut allocator, data) {
+        Err(_) => {
+            return;
+        }
+        Ok(r) => r,
```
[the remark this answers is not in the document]
richardkiss: Since I actually found problems through fuzzing, it's definitely finding valid serialization. Note that if the serialization completes and there are still bytes remaining at the end, it passes, so that leniency helps a lot.

arvidn (approved):

fuzz/fuzz_targets/deserialize_br.rs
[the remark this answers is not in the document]
richardkiss: Done in 31218ee4c5530fe5b2b69ee0f7ffb3286104cc55

src/serde/read_cache_lookup.rs
```diff
@@ -29,10 +30,10 @@ pub struct ReadCacheLookup {
     /// the tree hashes of the contents on the left and right
     read_stack: Vec<(Bytes32, Bytes32)>,
 
-    count: HashMap<Bytes32, u32>,
+    count: HashMap<Bytes32, u32, Bytes32Hasher>,
```
arvidn, line 33: wouldn't it make more sense to `impl Hash for Bytes32`?
This also highlights a potential attack, if the hashing is consistent, hash collisions can be forced by an attacker, causing the hash table to have O(n) lookups. My (perhaps optimistic) guess is that the default hasher mixes in some secret session key to make it impossible to attack.

arvidn: I was a bit confused. If you want a cheap hash function, you need to implement the `Hasher` yourself, of course. However, this hasher doesn't really have anything to do with `Bytes32`, so I think the name is a bit misleading. `XorHasher` or something would make more sense I think.

richardkiss:
> I think everything looks good, but the last commit probably should have been a separate PR. It introduced a clippy warning

Okay, I've rolled this one back for now.

arvidn (approved):

PR #223, opened 2022-11-10
Serde refactor

richardkiss:


richardkiss:
I think the only questionable part is the rename of `serialize` to `serde`. The name conflicts with the `serde` crate, but it's short for "serialization/deserialization" (à la "codec"), so really should be considered a generic term. That said, leaving it as `serialize` makes the patch smaller, since some imports don't need to change.

I'm on the fence. I guess I really did it mostly because I love the term `serde`.

arvidn (approved):

PR #224, opened 2022-11-17
Further refactor serde.

richardkiss:


src/serde/de.rs
```diff
@@ -77,21 +31,8 @@ pub fn node_from_stream(allocator: &mut Allocator, f: &mut Cursor<&[u8]>) -> io:
...
                 } else {
-                    let blob_size = decode_size(f, b[0])?;
-                    let pos = f.position() as usize;
-                    if f.get_ref().len() < pos + blob_size as usize {
-                        return Err(bad_encoding());
-                    }
-                    let blob = &f.get_ref()[pos..(pos + blob_size as usize)];
-                    f.seek(SeekFrom::Current(blob_size as i64))?;
-                    values.push(allocator.new_atom(blob)?);
+                    values.push(parse_atom(allocator, b[0], f)?);
```
arvidn, line 35: most of this patch is moving functions to new files, but this is one place where the code is refactored. Is it the only place, or is there some other places that warrant extra review too?

src/serde/parse_atom.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
+    Ok(v)
+}
+
+/// parse an atom from the stream
+/// At this point, the first byte has already been read to ensure it's
+/// not a special code like `CONS_BOX_MARKER` = 0xff, so it must be
+/// passed in too
+
+pub fn parse_atom(
```
arvidn, line 53: there's a great opportunity to add unit tests for this function now that it's been factored out. e.g. I believe the case where the input buffer is too short may not be covered well

arvidn (commented):
I take it the vast majority of this patch is whole functions being moved to new files. I found one place where you factored out `parse_atom()`. It looks good, but if there's any other places where you factored in or -out code, I'd like to review that too.

src/serde/parse_atom.rs
richardkiss: Added.

src/serde/de.rs
richardkiss: I believe this is the only new non-test function.

arvidn (approved):

PR #225, opened 2022-11-28
simplify run_program

arvidn:
by removing an op_stack entry that was only used once at start


richardkiss:
Maybe consider just swapping the stack elements manually and using `swap_eval_op` rather than having special-purpose code above the loop? I'm fine either way though.

arvidn:
I kind of like not having to create that extra pair too (just to match it again immediately following)

arvidn:
I realized there was one more significant simplification that could be done. I added one more commit

richardkiss (approved):
Looks good.

PR #226, opened 2022-11-30
run_program simplifications

arvidn:
some minor simplifications, saving lines of code and indentation levels.

These patches are probably best reviewed one at a time. They are all independent.


richardkiss (approved):
Seems fine.

PR #227, opened 2022-11-30
Optimize env stack

arvidn:
When we prepare calling an operator, we push all of its arguments along with the environment onto the `val_stack`, then we push the same number of `SwapEval` operations on the `op_stack`.

For any given operator call, the environment is the same for every single argument. It's unnecessary to duplicate this element on the `val_stack`.

With this patch, I add a separate `env_stack` where the environment is pushed every time we push an `Apply` operation, and popped every time we execute the apply operation. This eliminates the duplication of the environment reference on the stack.

In one stress test I used to measure the improvements, `/usr/bin/time` reported the following numbers:

| metric | before | after | after / before |
| --- | --- | --- | --- | 
| wall clock time | 0:05.34 | 0:04.11 | 75% |
| Max RSS | 4241124 | 2471488 | 58.3% |
| Minor faults | 1057391 | 618185 | 58.5% |

### DIFF
--- a/src/run_program.rs
+++ b/src/run_program.rs
@@ -37,6 +37,7 @@ struct RunProgramContext<'a, D> {
     pre_eval: Option<PreEval>,
     posteval_stack: Vec<Box<PostEval>>,
     val_stack: Vec<NodePtr>,
+    env_stack: Vec<NodePtr>,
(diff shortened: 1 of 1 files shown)


richardkiss (approved):
Oh nice, that's clever!

PR #231, opened 2022-12-08
Front load eval cost

arvidn:
This patch moves some of the cost spent when evaluating an operand to when we push it onto the stack, from when we pop it and evaluate it.
The second commit does the same thing for applying operators (which has minimal impact)

### DIFF
--- a/src/run_program.rs
+++ b/src/run_program.rs
@@ -14,6 +14,10 @@ const QUOTE_COST: Cost = 20;
 const APPLY_COST: Cost = 90;
 // mandatory base cost for every operator we execute
 const OP_COST: Cost = 1;
+// this is the minimum cost of evaluating an operator argument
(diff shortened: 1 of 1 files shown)


src/run_program.rs
```diff
@@ -14,6 +14,13 @@ const QUOTE_COST: Cost = 20;
 const APPLY_COST: Cost = 90;
 // mandatory base cost for every operator we execute
 const OP_COST: Cost = 1;
+// this is the minimum cost of evaluating an operator argument
+// we charge this cost up-front, when pushing the expression onto
+// the stack
+const EARLY_EVAL_COST: Cost = 20;
```
richardkiss, line 20: Why not just move the entire cost up front?

richardkiss (approved):

PR #233, opened 2022-12-19
revert front-load patches. add option to limit stack size

arvidn:
This reverts two patches that attempted to front-load the cost of evaluating arguments pushed onto the stack. They were both broken. The failures were detected by one of the fuzzers, but we don't run those on CI (yet, there's a separate PR for that).

In addition to reverting the two bad patches, a third commit splits outs parts if the `LIMIT_HEAP` flag into a `LIMIT_STACK`. This way the stack and heap limits can be employed independently.


AmineKhaldi (approved):

arvidn:
this PR (running the fuzzers) demonstrate the failure these reverted patches caused: https://github.com/Chia-Network/clvm_rs/pull/235

PR #237, opened 2022-12-21
move the mempool-mode flag into the rust crate

arvidn:
instead of defining it in the python interface. This enables code deduplication when the same flag is exported from the chia_rs wheel

### DIFF
--- a/wheel/src/api.rs
+++ b/wheel/src/api.rs
@@ -6,9 +6,7 @@ use super::run_program::{
     __pyo3_get_function_deserialize_and_run_program2, __pyo3_get_function_run_chia_program,
     __pyo3_get_function_serialized_length,
 };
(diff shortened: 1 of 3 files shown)


wheel/src/api.rs
```diff
@@ -6,9 +6,7 @@ use super::run_program::{
     __pyo3_get_function_deserialize_and_run_program2, __pyo3_get_function_run_chia_program,
     __pyo3_get_function_serialized_length,
 };
-use clvmr::chia_dialect::{LIMIT_HEAP, NO_NEG_DIV, NO_UNKNOWN_OPS, LIMIT_STACK};
-
-pub const MEMPOOL_MODE: u32 = NO_NEG_DIV | NO_UNKNOWN_OPS | LIMIT_HEAP | LIMIT_STACK;
```
arvidn, line 11: this definition is currently duplicated here: https://github.com/Chia-Network/chia_rs/blob/main/wheel/src/api.rs#L40

richardkiss (approved):

PR #238, opened 2022-12-21
add a build feature to enable pre-eval and post-eval functionality

arvidn:
the pre- and post-eval functionality is not used in block validation, but *is* used during compilation and debugging.
This allows block validation to not include this logic.

This change would (probably) require `clvm_tools_rs` to depend on this with the `pre-eval` feature enabled.

Note that the wheel in `clvm_rs` does not support the pre-eval function, so it doesn't need it enabled.

### DIFF
--- a/src/run_program.rs
+++ b/src/run_program.rs
@@ -25,6 +25,8 @@ enum Operation {
     Apply,
     Cons,
     SwapEval,
+
+    #[cfg(feature = "pre-eval")]
     PostEval,
 }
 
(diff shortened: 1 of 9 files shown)


src/run_program.rs
```diff
@@ -141,18 +146,20 @@ impl<'a, D: Dialect> RunProgramContext<'a, D> {
         Ok(())
     }
 
-    fn new(allocator: &'a mut Allocator, dialect: &'a D, pre_eval: Option<PreEval>) -> Self {
+    fn new(allocator: &'a mut Allocator, dialect: &'a D, _pre_eval: Option<PreEval>) -> Self {
```
richardkiss, line 149: You might want to simply do an `assert!(pre_eval.is_none());` in the case the feature isn't enabled. Or maybe even a panic that says the `pre_eval` feature isn't enabled. Or maybe `new_with_pre_eval` that only exists if the feature is enabled.

richardkiss (changes requested):
Looks good except one minor nit about `new`.

src/run_program.rs
arvidn: good point. then it won't be unused as well

prozacchiwawa: seconded, but i'd say make this entry point conditional on pre_eval as a feature and make a new one present in both that doesn't accept pre_eval.
that way code that expects this to work won't fail silently.

arvidn: I made this change. please see the second commit

prozacchiwawa (approved):

richardkiss (approved):
Seems reasonable. As long as it passes the tests (especially the pre-eval ones), I'm okay with it.

PR #239, opened 2022-12-22
introduce a checkpoint feature on the allocator

arvidn:
to restore to an earlier state.

The allows pre-loading an allocator with programs and data that will be reused in multiple executions. For example, we could load a program once and run it multiple times, restoring the checkpoint in between each call.

A specific example is when running block generators. Rather than parsing and loading the generator ROM into the allocator for each block, we can parse it into the allocator once and run multiple blocks (one at a time) by restoring to the checkpoint right after loading the ROM.

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -32,6 +32,15 @@ pub struct IntPair {
     rest: NodePtr,
 }
 
+// this represents a specific (former) state of an allocator. This can be used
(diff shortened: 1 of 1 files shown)


richardkiss (approved):
Not convinced it'll be useful, but seems straightforward enough to be low risk.

PR #241, opened 2023-01-03
Bump version

richardkiss:


arvidn (approved):
`brun` still use the wheel from this repo btw.

PR #246, opened 2023-01-19
New `run` function that doesn't import python code.

richardkiss:


arvidn:
to actually replace the python `clvm` module, we'd need the `SExp` wrapper too, with `as_python()` and `as_iter()`. Also, I don't think we have counterparts to `int_from_bytes()` and `int_to_bytes()`

wheel/src/run_program.rs
```diff
@@ -52,12 +47,13 @@ pub fn deserialize_and_run_program2(
...
-        Allocator::new_limited(500000000, 62500000, 62500000)
-    } else {
-        Allocator::new()
-    };
+) -> PyResult<(PyObject, LazyNode)> {
+  let mut allocator = if flags & LIMIT_HEAP != 0 {
+      Allocator::new_limited(500000000, 62500000, 62500000)
+  } else {
+      Allocator::new()
+  };
```
arvidn, line 55: something happened to the indentation here

wheel/src/api.rs
```diff
@@ -1,25 +1,51 @@
...
+}
+
+#[pyfunction]
+pub fn run_serialized_program(
+    py: Python,
+    program: &[u8],
+    args: &[u8],
+    max_cost: Cost,
+    flags: u32,
+) -> PyResult<(PyObject, LazyNode)> {
```
arvidn, line 25: it would be nice to have CI actually exercise the success and failure return paths from this function too

wheel/src/run_program.rs
```diff
@@ -52,12 +47,13 @@ pub fn deserialize_and_run_program2(
...
+    let mut allocator = Allocator::new();
```
arvidn, line 56: this looks like debugging left-over

wheel/src/adapt_response.rs
```diff
@@ -1,39 +1,26 @@
...
-
-pub fn adapt_response_to_py(
+pub fn adapt_response(
     py: Python,
     allocator: Allocator,
-    r: Response,
-) -> PyResult<(u64, LazyNode)> {
-    match r {
+    response: Response,
+) -> PyResult<(PyObject, LazyNode)> {
```
arvidn, line 13: why is this returning a `PyObject` now, instead of `u64`?

wheel/src/adapt_response.rs
```diff
@@ -1,39 +1,26 @@
...
         Ok(reduction) => {
             let val = LazyNode::new(Rc::new(allocator), reduction.1);
-            Ok((reduction.0, val))
+            let rv: PyObject = reduction.0.into_py(py);
+            Ok((rv, val))
+        }
+        Err(eval_err) => {
+            let rv: PyObject = eval_err.1.into_py(py);
+            let val = LazyNode::new(Rc::new(allocator), eval_err.0);
+            Ok((rv, val))
```
arvidn, line 23: so, this is not returning an `Err()`, does that mean that the `deserialize_and_run_program()` won't throw an exception on errors, but return an error?
That seems like a significant change in the API, especially since there are still failures that would cause it to fail, just not `EvalErr`.

richardkiss:
> to actually replace the python `clvm` module, we'd need the `SExp` wrapper too, with `as_python()` and `as_iter()`. Also, I don't think we have counterparts to `int_from_bytes()` and `int_to_bytes()`

Yes, there are quite a few things missing still. This is just a first step. The next PR after this will add a lot of python code that provides the real public interface, making this a hybrid wheel of python plus native code.

If you'd rather just see a massive PR with all the changes at once, I can do that. Let me know what you prefer.

wheel/src/adapt_response.rs
richardkiss: The idea is to return an `int` on success or a `str` (containing a human-readable message) on failure. This eliminates the dependency on `EvalError` (which is implemented in python).

arvidn:
> Yes, there are quite a few things missing still. This is just a first step. The next PR after this will add a lot of python code that provides the real public interface, making this a hybrid wheel of python plus native code.
> 
> If you'd rather just see a massive PR with all the changes at once, I can do that. Let me know what you prefer.

small steps are good

wheel/src/adapt_response.rs
arvidn: ok. that still leaves a second form of failure where the `PyResult` is an `Err`, right?

wheel/src/run_program.rs
richardkiss: Something weird is going on with `cargo fmt`. I'm investigating.

wheel/src/adapt_response.rs
richardkiss: That's the `str` case. The python code will do a switch on the type returned to determine success vs failure, and then turns it into an `EvalError` on failure. This api is not expected to be used without the python wrapper.

Having a hybrid wheel means stuff that doesn't need to be fast and can be written in python when it's easier to do so. This wrapping code (which would be ugly in rust) is an example.

wheel/src/adapt_response.rs
richardkiss: That's true. I can change the name if necessary. The thing is I believe no one is using the `clvm_rs` wheel any longer, so API changes seem pretty safe. At least, `chia-blockchain` is not using it.

richardkiss:
I've converted to draft for now so I can remove obsolete API.

wheel/src/adapt_response.rs
richardkiss: The tests and benchmarks were failing, and to fix them, I had to write the boilerplate I mentioned above, so you can now see first hand how it works.

wheel/src/adapt_response.rs
richardkiss: I've removed old api and created a brand new name for use going forward: `run_serialized_chia_program`.

richardkiss:
> Do you know if it's possible to define exception types in pyo3? The new interface seems quite awkward, by not always reporting errors via exceptions. The sad-path ends up using a lot more space in the source. It seems the ideal solution would be to return an `Err()` into a `PyResult` to indicate errors, we would just need a pyo3-based version of `EvalErr`, that can be "thrown" by returning an `Err()` on the rust side.

Yes, but that's exactly what this change is trying to avoid. I'd like the rust to not depend on the python, and the python not necessarily require the rust, which means not defining API-visible exceptions (like `EvalError`) in rust.

I suppose I could try returning a built-in exception type like `ValueError`, and then convert it into an `EvalError` in python. What do you think of that?

arvidn:
we should give the `chia_rs` binding the same treatment

PR #247, opened 2023-01-20
add parameter to set upper limit on output from serialize function

arvidn:


richardkiss (changes requested):
I think https://doc.rust-lang.org/std/primitive.u8.html gives a lot simpler way to write this patch (or, if you want to do it generally, create a structure that implements `io::Write` wrapper and limits the amount written).

arvidn:
ah, that's a good idea

richardkiss:
> I think https://doc.rust-lang.org/std/primitive.u8.html gives a lot simpler way to write this patch (or, if you want to do it generally, create a structure that implements `io::Write` wrapper and limits the amount written).

The link above is not what I thought. To be clear, `[u8]` implements `Write`, so that may be all you need. https://doc.rust-lang.org/std/io/trait.Write.html#impl-Write-for-%26mut%20%5Bu8%5D

src/serde/ser.rs
```diff
@@ -1,12 +1,75 @@
...
+
+    pub fn into_inner(self) -> W {
+        self.inner
+    }
+}
+
+impl<W: io::Write> Write for LimitedWriter<W> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        if (self.limit - buf.len() as i64) < 0 {
+            return Err(ErrorKind::OutOfMemory.into());
```
richardkiss, line 32: Consider

```suggestion
            return Err(ErrorKind:: StorageFull.into());
```

src/serde/ser.rs
```diff
@@ -1,12 +1,75 @@
...
+            let written = self.write(b)?;
+            ret += written;
+            if written < b.len() {
+                break;
+            }
+        }
+        Ok(ret)
+    }
+    fn write_all(&mut self, buf: &[u8]) -> io::Result<()> {
+        self.limit -= buf.len() as i64;
```
richardkiss, line 54: I may be wrong, but I don't think you have to implement anything besides `write` & `flush`. The defaults should probably be just fine.

src/serde/ser.rs
```diff
@@ -30,9 +93,48 @@ pub fn node_to_stream<W: io::Write>(node: &Node, f: &mut W) -> io::Result<()> {
 }
 
 pub fn node_to_bytes(node: &Node) -> io::Result<Vec<u8>> {
```
richardkiss, line 95: Maybe can we create a new function that takes a `max_size` parameter? And then we can move this `2_000_000` number higher up in the call stack.

richardkiss (changes requested):
A couple ideas to make things more complex in places and simpler in others.

src/serde/ser.rs
arvidn: I'm not sure about this. Today there's no use case for receiving an error *and* use the partial output for something. I think making this change (and adding this complexity) is definitely fine to do, as soon as we have a use case for it. (I subscribe to the "you ain't gonna need it" principle)

src/serde/ser.rs
arvidn: oh. I didn't realize that. I'll take a look!

src/serde/ser.rs
arvidn: `StorageFull` is not part of the stable API yet (as far as I understand). It requires building with the `io_error_more` feature. This makes me hesitate to use it for now.

src/serde/ser.rs
```diff
@@ -1,12 +1,43 @@
...
 
 use super::write_atom::write_atom;
 use crate::allocator::{NodePtr, SExp};
 use crate::node::Node;
 
 const CONS_BOX_MARKER: u8 = 0xff;
 
+pub struct LimitedWriter<W: io::Write> {
+    inner: W,
+    limit: i64,
```
richardkiss, line 14:
```suggestion
    limit: usize,
```
This will propagate a few other simple (and annoying) changes.

src/serde/ser.rs
richardkiss: Ah okay. I hadn't tried it.

I was looking around for more options, when I saw `WriteZero` and realized, it's probably enough for `write` to simply return `0` each time and let `write_fully` return the `WriteZero` error. I'll try it locally.

richardkiss: Yep, that works! In conjunction with the `min` code path, we get everything we need.

richardkiss (approved):
Looks good. Not crazy about having the 2_000_000 at this level but it'll do for now.

PR #250, opened 2023-01-28
rework workflows for consistency with other repos

altendky:


altendky:
So aside from the general tidying and such that I need to do...  @arvidn and @richardkiss (or even @cmmarslender), what do you think of sticking with ABI3 wheels vs. building separately for each Python version?  My main interest in considering dropping ABI3 is just for consistency across our workflow definitions.  There's enough complexity and variation between projects without having the main matrix also being different.  Though, I also feel a bit silly with that.  But I'm not sure that ABI3 really brings us a lot of value particularly.  Mostly that the wheels would be forward compatible with future Python versions, what else?  But I don't think that publishing a wheel is a big cost for us and we really ought to test the new Python version first anyways.

So, how do you all feel about this?

richardkiss:
I strongly believe that if we can use ABI3, we should do so. Why would we want to produce more artifact outputs when we can get away with producing fewer?

altendky:
> My main interest in considering dropping ABI3 is just for consistency across our workflow definitions. There's enough complexity and variation between projects without having the main matrix also being different. Though, I also feel a bit silly with that. But I'm not sure that ABI3 really brings us a lot of value particularly. Mostly that the wheels would be forward compatible with future Python versions, what else?

PR #252, opened 2023-02-04
remove NO_NEG_DIV flag and hard-code it in the division operator

arvidn:
now that the soft-fork has activated.

This patch removes all logic around having two separate `op_div` functions and makes a single one with the `NO_NEG_DIV` behavior. The test cases for the `/`-operator are updated to expect the no-negative-div behavior.

Since the soft-fork to disallow division with negative numbers activated a long time ago, there is no reason to keep this flag around anymore. It only creates a risk of forgetting to pass it in.

### DIFF
--- a/src/chia_dialect.rs
+++ b/src/chia_dialect.rs
@@ -4,15 +4,11 @@ use crate::cost::Cost;
 use crate::dialect::Dialect;
 use crate::err_utils::err;
 use crate::more_ops::{
-    op_add, op_all, op_any, op_ash, op_concat, op_div, op_div_deprecated, op_divmod, op_gr,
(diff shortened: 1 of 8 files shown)


ChiaMineJP (commented):
I searched for `NO_NEG_DIV` dependencies as below
image

We might need to dispatch PRs for `chia_rs` and `clvm_tools_rs` at the same time as this PR.

arvidn:
we pin our dependencies, so the fixup has to happen when we bump the dependency. This can be done atomically in the same commit (or at least PR)

richardkiss (approved):
Looks good.

PR #254, opened 2023-02-08
Add `skip_clvm_object` api.

richardkiss:


wheel/src/api.rs
```diff
@@ -10,11 +10,27 @@ use clvmr::{LIMIT_HEAP, LIMIT_STACK, MEMPOOL_MODE, NO_UNKNOWN_OPS};
...
+}
+
 #[pyfunction]
 pub fn serialized_length(program: &[u8]) -> PyResult<u64> {
     Ok(serialized_length_from_bytes(program)?)
 }
 
+#[pyfunction]
+pub fn skip_clvm_object(obj: &PyAny) -> PyResult<()> {
+    Ok(parse_through_clvm_object(&mut ReadPyAny(obj.getattr("read")?))?)
```
richardkiss, line 31: This repeatedly calls `.read` on the python object passed.

src/serde/tools.rs
```diff
@@ -239,16 +224,16 @@ fn test_serialized_length_from_bytes() {
     );
 
     let e = serialized_length_from_bytes(&[0x8f, 0xff]).unwrap_err();
-    assert_eq!(e.kind(), bad_encoding().kind());
-    assert_eq!(e.to_string(), "bad encoding");
+    assert_eq!(e.kind(), io::ErrorKind::UnexpectedEof);
+    assert_eq!(e.to_string(), "copy terminated early");
```
arvidn, line 228: is the standard library only using `UnexpectedEof` for copy operations? Is that why the error message is a bit awkward?

src/serde/utils.rs
```diff
@@ -0,0 +1,24 @@
...
+        Err(Error::new(
+            std::io::ErrorKind::UnexpectedEof,
+            "copy terminated early",
+        ))
+    } else {
+        Ok(())
+    }
+}
+
+pub fn skip_bytes<R: Read>(f: &mut R, skip_size: u64) -> io::Result<()> {
```
arvidn, line 22: This seems like a very elaborate (and costly) way of saying
```
f.seek(SeekFrom::Current(skip_size));
```

I think it would be better to just use seek than to actually copy into a sink. If you need this copy function in a future PR, I think you should introduce it there, and with unit tests.

wheel/src/api.rs
```diff
@@ -5,16 +5,34 @@ use clvmr::chia_dialect::ChiaDialect;
...
 use clvmr::run_program::run_program;
-use clvmr::serde::{node_from_bytes, serialized_length_from_bytes};
+use clvmr::serde::{node_from_bytes, parse_through_clvm_object, serialized_length_from_bytes};
 use clvmr::{LIMIT_HEAP, LIMIT_STACK, MEMPOOL_MODE, NO_UNKNOWN_OPS};
 use pyo3::prelude::*;
 use pyo3::wrap_pyfunction;
 
+struct ReadPyAny<'py>(&'py PyAny);
+
+impl<'py> std::io::Read for ReadPyAny<'py> {
```
arvidn, line 15: I assume this python interface doesn't support seeking forward. I think the work-around to just read the bytes that you want to skip should be moved here. `ReadPyAny` should implement `Seek` and do the trick.

(which still doesn't need a copy function btw, just to read and discard bytes)

wheel/src/api.rs
```diff
@@ -5,16 +5,34 @@ use clvmr::chia_dialect::ChiaDialect;
...
+    }
+}
+
 #[pyfunction]
 pub fn serialized_length(program: &[u8]) -> PyResult<u64> {
     Ok(serialized_length_from_bytes(program)?)
 }
 
+#[pyfunction]
+pub fn skip_clvm_object(obj: &PyAny) -> PyResult<()> {
```
arvidn, line 30: `obj` is a python input stream, right? I think how this function is meant to be used warrants a comment. and perhaps `obj` could have a better name (unless I misunderstand it)

wheel/src/api.rs
```diff
@@ -43,6 +61,7 @@ pub fn run_serialized_chia_program(
 fn clvm_rs(_py: Python, m: &PyModule) -> PyResult<()> {
     m.add_function(wrap_pyfunction!(run_serialized_chia_program, m)?)?;
     m.add_function(wrap_pyfunction!(serialized_length, m)?)?;
+    m.add_function(wrap_pyfunction!(skip_clvm_object, m)?)?;
```
arvidn, line 64: don't forget to add this function to `wheel/generate_type_stubs.py` and rerun it as well

arvidn (changes requested):

src/serde/utils.rs
richardkiss: The `Read` trait does not have `.seek`.

wheel/src/api.rs
richardkiss: True, I could write my own loop to read bytes, but `copy` https://doc.rust-lang.org/std/io/fn.copy.html is parameterized by type, so the `Sink` writes should go away.

wheel/src/api.rs
richardkiss: I think that's only in `chia_rs`. The python API here is small enough to manage by hand.

src/serde/utils.rs
arvidn: right, you could make `R` `Read + Seek`

wheel/src/api.rs
arvidn: you already did write your own loop reading bytes. you just also write them to another stream. what I mean is that if you're going to add this feature you should make sure to add tests for it as well. But given that we don't need it (the copy), it would make more sense to do none of them.

wheel/src/api.rs
arvidn: oh, right

src/serde/tools.rs
richardkiss: It's because of the `copy` call. I agree it's a bit confusing. Maybe I should use `.map` to swap out the error message.

wheel/src/api.rs
richardkiss: Oh yeah, I suppose I could also pass a `Write` to `.skip_clvm_object`! That would be a lot more direct (and mean we don't have to call out to python).

richardkiss:
Okay, what I really want to do here is get a `r: Read` and pull the bytes out and copy elsewhere, so I'll try again with a `Read` and a `Write`.

src/serde/tools.rs
```diff
@@ -15,43 +16,27 @@ enum ParseOp {
...
-                    //ops.push(ParseOp::Cons);
-                    ops.push(ParseOp::SExp);
-                    ops.push(ParseOp::SExp);
-                } else if b[0] == 0x80 || b[0] <= MAX_SINGLE_BYTE {
-                    // This one byte we just read was the whole atom.
-                    // or the
-                    // special case of NIL
-                } else {
-                    let blob_size = decode_size(&mut f, b[0])?;
-                    f.seek(SeekFrom::Current(blob_size as i64))?;
```
arvidn, line 42: currently you're replacing this `seek()` call with a loop of `read()`. I don't see a good reason to do that in the case where we have a `Cursor` (which does implement `Seek`). If you want to be able to pass a python stream that doesn't support seek, I think it would be best to work around that at the python boundary.

src/serde/tools.rs
```diff
@@ -15,43 +16,27 @@ enum ParseOp {
 
 pub fn serialized_length_from_bytes(b: &[u8]) -> io::Result<u64> {
     let mut f = Cursor::new(b);
-    let mut ops = vec![ParseOp::SExp];
+    parse_through_clvm_object(&mut f).map_err(|_e| bad_encoding())?;
+    Ok(f.position())
+}
+
+pub fn parse_through_clvm_object<R: Read>(f: &mut R) -> io::Result<()> {
```
arvidn, line 23: I'm suggesting making this: `R: Read + Seek`

richardkiss:
Duplicate.

PR #259, opened 2023-03-04
Add python implementation of `Program` and helper functions.

richardkiss:


wheel/python/clvm_rs/program.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def tree_hash(self) -> bytes:
+        # we operate on the unwrapped version to prevent the re-wrapping that
+        # happens on each invocation of `Program.pair` whenever possible
+        if self._cached_sha256_treehash is None:
+            self._cached_sha256_treehash = sha256_treehash(self._unwrapped)
+        return self._cached_sha256_treehash
+
+    def run_with_cost(
+        self, args, max_cost: int = MAX_COST, flags: int = 0
```
trepca, line 257: this is a bit unrelated, but wondering how come we chose to default to MAX_COST and not BLOCK_COST? Assuming that most people will write clvm programs to run on chain not as a scripting language.

wheel/python/clvm_rs/program.py
```diff
@@ -0,0 +1,346 @@
...
+        v = self.as_int()
+        if v is None:
+            raise ValueError("can't cast pair to int")
+        return v
+
+    def __hash__(self):
+        return id(self)
+
+    @classmethod
+    def int_from_bytes(cls, b: bytes) -> int:
```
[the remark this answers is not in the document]
richardkiss: The idea is to present all of `clvm_rs` api under a single class to minimize imports and act as a public-facing interface. Python has a bit of a problem when it comes to keeping private API private.

wheel/python/clvm_rs/program.py
richardkiss: This is the classic problem of how when you go down a stack of calls, the parameters and call names go from more specific to more generic. IMO, this level should know as little as possible about consensus, and encoding a consensus constant at this level might temp users to depend on it when they really probably shouldn't.

wheel/python/clvm_rs/program.py
```diff
@@ -0,0 +1,346 @@
...
+        )
+        obj = cls.wrap(tree)
+        new_cursor = len(bytes(tree)) + cursor
+        return obj, new_cursor
+
+    @classmethod
+    def fromhex(cls, hexstr: str) -> Program:
+        return cls.from_bytes(bytes.fromhex(hexstr))
+
+    def __bytes__(self) -> bytes:
```
Quexington, line 59: It would be awesome to have a `bytes`-like method that returns the bytes representation of something minus the size prefix for use cases like `puzzle_hash: bytes = prog.atom_bytes` for example instead of `puzzle_hash: bytes = bytes(prog)[1:]`

wheel/python/clvm_rs/program.py
```diff
@@ -0,0 +1,346 @@
...
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}({str(self)})"
+
+    def __eq__(self, other) -> bool:
+        try:
+            other_obj = self.to(other)
+        except ValueError:
+            # cast failure
+            return False
+        return self.tree_hash() == other_obj.tree_hash()
```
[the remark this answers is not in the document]
richardkiss: Yes, it would be more efficient to recursively compare leaf nodes, and that code has existed.

However, it's pure python, and tree hashing is generally done in fast native code, so it may end up being faster to do the hash even in the case of it not yet being cached.

I don't expect this to be used very much outside of tests (I believe wallets would typically compare hashes), so porting to native code is probably not warranted.

wheel/python/clvm_rs/program.py
richardkiss: That's `prog.atom`.

richardkiss: Do not use `bytes(prog)[1:]` as that will not be the same for strings with length >= 0x3f.

wheel/python/clvm_rs/program.py
```diff
@@ -0,0 +1,346 @@
...
+        The resulting argument list is interpreted with apply (2)
+
+        (a (q . self) rest)
+
+        Resulting in a function which places its own arguments after those
+        curried in in the form of a proper list.
+        """
+        return self.to(self.curry_treehasher.curry(self, *args))
+
+    def uncurry(self) -> Tuple[Program, Optional[List[Program]]]:
```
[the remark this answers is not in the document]
richardkiss: I don't have a good understanding of how repeated attempts at uncurrying something would be done. Would it be to determine what type of contract a puzzle is?

Seems to me that there should be multiple layers, one which uncurries the puzzle to determine the `mod`, then another that dispatches to the appropriate sub-wallet handler based on that mod, passing in the curried parameters at that time to obviate the need for a second uncurrying.

wheel/python/clvm_rs/program.py
```diff
@@ -0,0 +1,346 @@
...
+        new_cursor = len(bytes(tree)) + cursor
+        return obj, new_cursor
+
+    @classmethod
+    def fromhex(cls, hexstr: str) -> Program:
+        return cls.from_bytes(bytes.fromhex(hexstr))
+
+    def __bytes__(self) -> bytes:
+        if self._cached_serialization is None:
+            self._cached_serialization = sexp_to_bytes(self)
```
arvidn, line 61: I think I'm misunderstanding something. Aren't the serialized form available in `self.blob`? You just need to pick the right slice from it. This looks like it's actually producing a new serialization.

wheel/python/clvm_rs/program.py
arvidn: I think it's a security risk to have this `MAX_COST` as a default parameter. I can't think of any situation where we *want* to use this as the cost limit

wheel/python/clvm_rs/program.py
```diff
@@ -0,0 +1,346 @@
...
+        try:
+            cost, lazy_node = run_serialized_chia_program(
+                prog_bytes, args_bytes, max_cost, flags
+            )
+            r = self.wrap(lazy_node)
+        except ValueError as ve:
+            raise EvalError(ve.args[0], self.wrap(ve.args[1]))
+        return cost, r
+
+    def run(self, args) -> "Program":
```
arvidn, line 270: I think `run_with_cost()` should be renamed to `run()`. We don't want the shorter and better sounding name to be the dangerous one that should never be used. Since I don't think it's a good idea to provide a way to run programs without explicitly specifying a cost limit

wheel/python/clvm_rs/program.py
```diff
@@ -0,0 +1,346 @@
...
+        cost, r = self.run_with_cost(args, MAX_COST)
+        return r
+
+    def curry(self, *args) -> "Program":
+        """
+        Replicates the curry function from clvm_tools, taking advantage of *args
+        being a list.  We iterate through args in reverse building the code to
+        create a clvm list.
```
arvidn, line 278: this comment looks like it describes an earlier implementation. is `args` really required to be a list here? (if so, it should be typed as such)

wheel/python/clvm_rs/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .eval_error import EvalError  # noqa: F401
+from .program import Program  # noqa: F401
```
arvidn, line 2: is the point of this to re-export these names? If so, I believe the way do to that (to not get the warnings) is to add:

```
__all__ = ["Program", "EvalError"]
```

wheel/python/clvm_rs/at.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import annotations
+from typing import Optional, cast
+
+from .clvm_storage import CLVMStorage
+
+
+def at(obj: CLVMStorage, position: str) -> Optional[CLVMStorage]:
```
arvidn, line 7: there should be unit tests for this function.
Also, I don't see any circumstance where this function returns `None`. Should it not just return `CLVMStorage`?

wheel/python/clvm_rs/at.py
```diff
@@ -0,0 +1,40 @@
...
+
+    Returns `None` if an atom is hit at some intermediate node.
+
+    ```
+    p1 = Program.to(10)
+    assert None == at(p1, "rr")
+    ```
+
+    """
+    v: Optional[CLVMStorage] = obj
```
arvidn, line 26: why is `v` an `Optional`? it seems like it's always set.

wheel/python/clvm_rs/at.py
```diff
@@ -0,0 +1,40 @@
...
+    for c in position.lower():
+        v = cast(CLVMStorage, v)
```
arvidn, line 28: is this just to cast-away the optionality? It seems redundant

wheel/python/clvm_rs/casts.py
```diff
@@ -0,0 +1,170 @@
...
+    if size == 0:
+        return 0
+    return int.from_bytes(blob, "big", signed=True)
+
+
+def int_to_bytes(v) -> bytes:
+    """
+    Convert a python int to a blob that encodes as this integer in clvm.
+    """
+    byte_count = (v.bit_length() + 8) >> 3
```
arvidn, line 65: is there really a good reason to use `>> 3` instead of `// 8`?

wheel/python/clvm_rs/casts.py
```diff
@@ -0,0 +1,170 @@
...
+    if v == 0:
+        return b""
```
arvidn, line 67: you could probably move this test and early-exit to before you compute `byte_count`

wheel/python/clvm_rs/casts.py
```diff
@@ -0,0 +1,170 @@
...
+        return memview_or_bytes_py38_or_later
+    except TypeError:
+        pass
+    return memview_or_bytes_py37
+
+
+memview_or_bytes = generate_working_memview_or_bytes()
+
+
+def int_from_bytes(blob):
```
arvidn, line 51: I can't find unit tests for this. Also, you're missing type annotations.

wheel/python/clvm_rs/casts.py
```diff
@@ -0,0 +1,170 @@
...
+    """
+    Convert a bytes blob encoded as a clvm int to a python int.
+    """
+    size = len(blob)
+    if size == 0:
+        return 0
+    return int.from_bytes(blob, "big", signed=True)
+
+
+def int_to_bytes(v) -> bytes:
```
arvidn, line 61: I can't find unit tests for this function, and `v` is missing type annotation

wheel/python/clvm_rs/casts.py
```diff
@@ -0,0 +1,170 @@
...
+        return v.encode()
+    if isinstance(v, int):
+        return int_to_bytes(v)
+    if memview_or_bytes(v):
+        return bytes(v)
+
+    raise ValueError("can't cast %s (%s) to bytes" % (type(v), v))
+
+
+def to_clvm_object(
```
arvidn, line 93: I think you should have a unit test for this function. It should also have a return type annotation.
I imagine this is a dangerous call to make on untrusted input. It might make sense to document that.

arvidn (commented):
I'm only 9 out of 25 files in. I'll just commit the comments I have so far.
I think it would be good to set up `mypy` with the strictest settings in CI.

wheel/python/clvm_rs/program.py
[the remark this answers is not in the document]
Quexington: Something we do a lot is `if some_prog == SOME_MOD:` and I would expect we would not have need for either of those hashes so we would end up doing two tree hashes where before we were doing 0.

wheel/python/clvm_rs/program.py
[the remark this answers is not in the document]
Quexington: > Seems to me that there should be multiple layers, one which uncurries the puzzle to determine the mod, then another that dispatches to the appropriate sub-wallet handler based on that mod, passing in the curried parameters at that time to obviate the need for a second uncurrying

Right yeah, this is what we had to build to solve the problem so I'm wondering if we could eliminate the need for this architecture (and eliminate the possibility of omitting it and writing inefficient code because you don't know that `uncurry` is inefficient) by caching the values

wheel/python/clvm_rs/program.py
```diff
@@ -0,0 +1,346 @@
...
+    @classmethod
+    def new_pair(cls, left: CLVMStorage, right: CLVMStorage) -> Program:
+        o = cls()
+        o.atom = None
+        o._pair = None
+        o._unwrapped_pair = (left, right)
+        return o
+
+    @classmethod
+    def null(cls) -> Program:
```
Quexington, line 136: What do you think about a similar program that uses `1`?  Useful for a lot of testing, and redefining at the top of every file is annoying.

wheel/python/clvm_rs/at.py
richardkiss: It's just to make `mypy` happy.

wheel/python/clvm_rs/at.py
```diff
@@ -0,0 +1,40 @@
...
+        pair = v.pair
+        if pair is None:
+            return None
```
richardkiss, line 31: This here is the `None` case (ie. we're trying to drill down through an atom).

wheel/python/clvm_rs/at.py
richardkiss: Google "mypy using `cast`", for example https://adamj.eu/tech/2021/07/06/python-type-hints-how-to-use-typing-cast/ or https://stackoverflow.com/questions/55461519/python-type-hints-better-syntax-for-cast

richardkiss: Actually, you're right. This was required until I made some other change, so I'll get rid of it.

wheel/python/clvm_rs/at.py
richardkiss: The `None` is below, line 31.

wheel/python/clvm_rs/at.py
richardkiss: You're right, removing.

wheel/python/clvm_rs/casts.py
richardkiss: It's tested indirectly through `test_int_round_trip` and `Program.int_from_bytes` in wheel/python/tests/test_program.py. I've actually been using `coverage` (manually) to maximize test coverage, and am at nearly 100%.

I would be interested in getting feedback on how to automate coverage reporting and track it in CI. Not sure what best practices are for that right now.

wheel/python/clvm_rs/casts.py
richardkiss: Also in `test_program.py:test_int_round_trip` via `Program.int_to_bytes`.

wheel/python/clvm_rs/casts.py
richardkiss: I assumed it would be faster, or maybe compile to the same thing. Turns out I was wrong.

```
Disassembly of <code object div8 at 0x100432ef0, file "shift_vs_div.py", line 1>:
  2           0 LOAD_FAST                0 (n)
              2 LOAD_CONST               1 (8)
              4 BINARY_FLOOR_DIVIDE
              6 RETURN_VALUE

Disassembly of <code object shift3 at 0x100432fa0, file "shift_vs_div.py", line 5>:
  6           0 LOAD_FAST                0 (n)
              2 LOAD_CONST               1 (3)
              4 BINARY_RSHIFT
              6 RETURN_VALUE
(venv) kiss@newton add-python % p -m timeit -s 'from shift_vs_div import div8, shift3' 'shift3(478491)'
5000000 loops, best of 5: 41.6 nsec per loop
(venv) kiss@newton add-python % p -m timeit -s 'from shift_vs_div import div8, shift3' 'div8(478491)'  
10000000 loops, best of 5: 38.3 nsec per loop
(venv) kiss@newton add-python % p -m timeit -s 'from shift_vs_div import div8, shift3' 'shift3(478491)'
5000000 loops, best of 5: 41.9 nsec per loop
```

So I'll switch to `// 8`.

wheel/python/clvm_rs/casts.py
richardkiss: The only dangerous case I can think of is a recursive list structure (as in `t = []; t.append(t); to_clvm_object(t)`). Any other cases?

I don't see this being used on untrusted input much, since untrusted input would generally be in clvm form already, rather than python objects.

wheel/python/clvm_rs/clvm_tree.py
```diff
@@ -0,0 +1,93 @@
...
+        self,
+        blob: Union[memoryview, bytes],
+        int_tuples: List[Tuple[int, int, int]],
+        tree_hashes: Optional[List[bytes]],
+        index: int,
+    ):
+        self.blob = blob
+        self.int_tuples = int_tuples
+        self.tree_hashes = tree_hashes
+        self.index = index
```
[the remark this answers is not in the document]
richardkiss: I'll add type annotation for the members.

I'm considering making it a `dataclass`. I'm not sure if there are any advantages as I wouldn't be using the automatically generated `__init__` nor `__str__` or `__repr__`. What advantages do you expect I would get?

wheel/python/clvm_rs/program.py
richardkiss: The `sexp_to_bytes` knows about the optional `_cached_serialization` memoizing. That's necessary because you might go `(q . MOD)` where the serialization for `MOD` is cached, so you want to check at each step down the tree and not just at the top.

I have a separate (but not checked in) benchmark tool which runs a bunch of cases and makes sure only the things that are supposed to be slow are slow, and everything else that should use memoized values (tree hashing, serialization, etc.) must run fast (less than .01 s). It's been useful in spotting performance bugs. I'm not sure how to check it in yet though, since benchmarking of this sort is so dependent on the platform it runs on (ie. how to generalize the concept of .01 s of time on my M2 Mac to "slow" on other less predictable platforms like in CI).

wheel/python/clvm_rs/program.py
richardkiss: I'd be open to that. Do you prefer `.true` or `.one`, or something else?

wheel/python/clvm_rs/program.py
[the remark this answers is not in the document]
richardkiss: Generally, `SOME_MOD` is going to come from a hex file, which means you'll use `Program.from_bytes`, which by default generates the tree hash for the entire hierarchy (and Coming Soon™ in rust, so it'll be very fast). Same with `some_prog`. Anything that comes from a deserialization rather than building by "hand" (with `Program.to` for example), you'll have to work hard to not pre-generate and cache the tree hashes, which will make these equality comparisons fast.

wheel/python/clvm_rs/program.py
richardkiss: I'm a bit flummoxed as to how typing works for `*args` in python. Let me look it up.

richardkiss: Okay, I want `*args: CastableType` here (as opposed to what I originally thought, which was `List[CastableType]`).

wheel/python/clvm_rs/program.py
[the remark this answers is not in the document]
richardkiss: How about this code flow: if you know a puzzle is a curried `MOD`, uncurry it to get the `MOD` and parameters. You use the `MOD` hash to figure out which "driver", then the curried parameters to figure out which instance of the driver is appropriate.

I think a list of `if MOD == a`, `if MOD == b`, etc. is slow. In general, a `switch/case` can often be replaced with a `dict` look-up.

wheel/python/clvm_rs/__init__.py
richardkiss: Sweet trick!

wheel/python/clvm_rs/at.py
richardkiss: Tests are in `wheel/python/tests/test_program.py` https://github.com/Chia-Network/clvm_rs/blob/0ddb2668434de8a19222eed56fef060f9e7ca44b/wheel/python/tests/test_program.py#L14

For some reason, github decided there were too many changes in this (new) file and elides it from the default view of this patch.

wheel/python/clvm_rs/clvm_storage.py
```diff
@@ -0,0 +1,26 @@
...
+    from typing import Protocol
+except ImportError:
+    Protocol = cast(_SpecialForm, object)
+
+
+class CLVMStorage(Protocol):
+    atom: Optional[bytes]
+
+    @property
+    def pair(self) -> Optional[Tuple["CLVMStorage", "CLVMStorage"]]:
```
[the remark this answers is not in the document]
richardkiss: I'm on the fence about this, since "the future" on this was supposed to happen in py310, but it was delayed once, so I'm not 100% convinced it's going to happen. In any case, both syntaxes will likely have to be supported for some time, so we can transition once py311 is deprecated (which is a long time from now).

wheel/python/clvm_rs/program.py
richardkiss: Point taken. I need to think more about how to handle this. Perhaps a `Program` class-level `DEFAULT_MAX_COST` that can be set by applications at start-up time. I'm a little hesitant to set policy on this that knows about Chia consensus though.

wheel/python/clvm_rs/casts.py
richardkiss: Done.

wheel/python/clvm_rs/casts.py
richardkiss: Unit tests are in https://github.com/Chia-Network/clvm_rs/blob/0ddb2668434de8a19222eed56fef060f9e7ca44b/wheel/python/tests/test_as_python.py (I probably should rename this file).

wheel/python/clvm_rs/clvm_tree.py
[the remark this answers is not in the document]
richardkiss: The source is here https://peps.python.org/pep-0557/ Doesn't seem like a big win in this case (limited utility for `__eq__` and `__lt__` etc. for this class).

wheel/python/clvm_rs/program.py
richardkiss: I did this with a `UNSAFE_MAX_COST` value, which must be set prior to using `.run`.

wheel/python/clvm_rs/program.py
richardkiss: I've done this a bit differently, but I believe I've addressed it.

wheel/python/clvm_rs/program.py
richardkiss: I went with `.one`.

wheel/python/clvm_rs/program.py
ChiaMineJP, line 1: Sorry if this's a silly question.
Is this `Program` class trying to replace:
https://github.com/Chia-Network/chia-blockchain/blob/main/chia/types/blockchain_format/program.py ?

wheel/python/clvm_rs/casts.py
arvidn: I used to assume that the wallet would only manipulate clvm structures and programs that it had produced itself, and therefore was safe. I don't think that's the case anymore and I think it's risky to assume that code that is only called on trusted structures *today* will not be used on untrusted structures in the future. I believe there are plenty of examples of this happening in the wallet.

wheel/python/clvm_rs/casts.py
arvidn: I'm surprised the compiler even produces different bytecode

wheel/python/clvm_rs/tree_hash.py
```diff
@@ -0,0 +1,107 @@
...
+        obj_stack: List[CLVMStorage] = [clvm_storage]
+        op_stack: List[OP_STACK_F] = [handle_obj]
+        hash_stack: List[bytes] = []
+        while len(op_stack) > 0:
+            op: OP_STACK_F = op_stack.pop()
+            op(obj_stack, hash_stack, op_stack)
+        return hash_stack[0]
+
+
+CHIA_TREEHASHER = Treehasher(bytes.fromhex("01"), bytes.fromhex("02"))
```
arvidn, line 103: could you not pull these values from `CHIA_DIALECT`, to make it a bit clearer what they mean?

wheel/python/clvm_rs/tree_hash.py
```diff
@@ -0,0 +1,107 @@
...
+from hashlib import sha256
+from typing import Callable, List, Tuple, cast
+
+from .clvm_storage import CLVMStorage
+
+
+OP_STACK_F = Callable[[List[CLVMStorage], List[bytes], List["OP_STACK_F"]], None]
+
+
+class Treehasher:
```
arvidn, line 18: it seems like a bit of a detour to create a class to record the tree-hash format, just to instantiate a singleton and pull out the member functions. Why not just write the free functions directly?
You don't reuse this class anywhere, and we only have one form of tree-hash (and I think we would need to have a really good reason to add another kind, with values other than `1` and `2`)

wheel/python/clvm_rs/tree_hash.py
```diff
@@ -0,0 +1,107 @@
...
+    way so that extremely large objects don't blow out the python stack.
+
+    We also force a `_cached_sha256_treehash` into the hashed sub-objects
+    whenever possible so that taking the hash of the same sub-tree is
+    more efficient in future.
+    """
+
+    atom_prefix: bytes
+    pair_prefix: bytes
+    cache_hits: int
```
arvidn, line 30: I don't see any use of `cache_hits` (but maybe it's in a collapsed diff)

wheel/python/clvm_rs/replace.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import annotations
+from typing import Dict
+
+from .casts import CastableType
+from .clvm_storage import CLVMStorage
+
+
+def replace(program: CLVMStorage, **kwargs: CastableType) -> CastableType:
```
arvidn, line 8: I take it you have some future need for this in mind, yeah?

arvidn (commented):
is `Program` a drop-in replacement of `chia.types.blockchain_format.Program` now?
if so, I would recommend you profile the wallet, with this version of `Program` while syncing the toilet wallet.

wheel/python/clvm_rs/casts.py
richardkiss: The object that comes back from `.run` conforms to `CLVMStorage`. This function should notice and return the original object with no changes.

wheel/python/clvm_rs/program.py
richardkiss: Yes, eventually. There are some minor differences ("improvements") to the API which means there will be a transition phase.

wheel/python/clvm_rs/program.py
```diff
@@ -0,0 +1,350 @@
...
+
+
+class Program(CLVMStorage):
+    """
+    A wrapper around `CLVMStorage` providing many convenience functions.
+    """
+
+    UNSAFE_MAX_COST: Optional[int] = None
+
+    curry_treehasher: CurryTreehasher = CurryTreehasher(CHIA_DIALECT)
```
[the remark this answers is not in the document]
richardkiss: These are class members. In theory, `Program` objects can override them. But they are not intended to.

wheel/python/clvm_rs/program.py
```diff
@@ -0,0 +1,350 @@
...
+        return obj, new_cursor
+
+    @classmethod
+    def fromhex(cls, hexstr: str) -> Program:
+        return cls.from_bytes(bytes.fromhex(hexstr))
+
+    def __bytes__(self) -> bytes:
+        if self._cached_serialization is None:
+            self._cached_serialization = sexp_to_bytes(self)
+        if not isinstance(self._cached_serialization, bytes):
```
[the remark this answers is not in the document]
richardkiss: I'm a bit confused by this question, as that line of code appears verbatim in the above.

wheel/python/clvm_rs/clvm_storage.py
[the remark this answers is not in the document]
richardkiss: Possibly historically, but it seems to find type bugs just fine. It supports recursive types now too, which is a huge relief, since there's not really a way to define clvm types without it.

wheel/python/clvm_rs/casts.py
richardkiss: Turns out there is a subtle difference if the dividend is a float.

```
>>> 10.0 // 8
1.0
>>> 10.0 >> 3
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for >>: 'float' and 'int'
```

richardkiss: Maybe not so subtle.

wheel/python/clvm_rs/casts.py
richardkiss: > I just realized this is converting in the opposite direction though, so probably lower risk. As long as people don't get a lazily convertible output from a malicious program, and then convert it back to clvm via this function.

I removed `.as_python` from the API for just this reason, and now point people to safer alternatives like `.as_iter`, `.at` and `.at_many`.

wheel/python/clvm_rs/curry_and_treehash.py
```diff
@@ -0,0 +1,156 @@
...
+from .casts import CastableType
+from .chia_dialect import Dialect
+from .clvm_storage import CLVMStorage
+from .tree_hash import shatree_pair, shatree_atom
+
+
+ONE = bytes.fromhex("01")
+
+
+class CurryTreehasher:
```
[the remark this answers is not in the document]
richardkiss: It's an attempt to separate the algorithms from the magic constants. The magic constants won't change, but it provides for them in a "don't repeat yourself" kind of way while using as few global imports as possible.

There isn't much difference between using a class and a module in this way, and I doubt one is significantly faster than the other. This makes it easier to monkey-patch. I agree it's unlikely anyone would ever want to change the constants, but I still believe it's a better pattern.

wheel/python/clvm_rs/tree_hash.py
richardkiss: These are the magic `1` and `2` constants of `tree_hash`, which doesn't have to do with the chia dialect. They have to do with the sha256-based tree hash standard.

wheel/python/clvm_rs/tree_hash.py
richardkiss: See above. It's motivated by minimizing the number of things `Program` needs to use, and tie all those magic values into one class (not too different from a module except an instance is instantiated).

wheel/python/clvm_rs/tree_hash.py
richardkiss: It was used during development. It might be good for tests or benchmarks to ensure that fast code paths are taken when expected. I can remove it, but I'd expect it not to have much performance impact.

wheel/python/clvm_rs/replace.py
richardkiss: This is the implementation for `Program.replace`. The idea was to make `Program` an amalgamation of various disparate api.

This `replace` works on any `CLVMStorage` type. The one in `Program` casts to `Program`, as you'd expect. This follows the philosophy that `Program` can be just a thin wrapper around some other `CLVMStorage` type (like rust's `LazyNode`) and provide it with all the same conveniences.

wheel/python/clvm_rs/chia_dialect.py
```diff
@@ -0,0 +1,40 @@
...
+        "+ - * / divmod > ash lsh "
+        # opcodes on atoms as vectors of bools 0x18-0x1c
+        "logand logior logxor lognot . "
+        # opcodes for bls 1381 0x1d-0x1f
+        "point_add pubkey_for_exp . "
+        # bool opcodes 0x20-0x23
+        "not any all . "
+        # misc 0x24
+        "softfork "
+    ).split(),
```
ChiaMineJP, line 33: I recall that `index` - `keyword` mapping like this is defined in `src/chia_dialect.rs`.
For maintenanceability, the mapping should be defined in one place.

wheel/python/tests/test_program.py
```diff
@@ -0,0 +1,495 @@
...
+    pair = tree.pair
+    assert pair is not None
+    ret = "("
+    for i in pair:
+        ret += print_tree(i)
+    ret += ")"
+    return ret
+
+
+class ProgramTest(TestCase):
```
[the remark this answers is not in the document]
richardkiss: I still have a soft spot for `unittest` since it's built-in and doesn't require `pytest`, although I suppose you're right that there's not really much benefit here, and it could possibly increase confusion to someone who hasn't lived through the `unittest` -> `pytest` ecosystem shift.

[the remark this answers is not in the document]
richardkiss: Turns out that making this change would require rewriting both `self.assertEqual` and `self.assertRaises`, which I'd prefer not to do because I'm worried there might be something subtle going on in them. Probably not, but I think I'm going to just leave this as-is for now.

This is inherited code stolen from `chia-blockchain` or `clvm`, so it could be quite old by this point. I think it's best not to poke the beast too hard here.

wheel/python/clvm_rs/chia_dialect.py
richardkiss: I agree that this violates "DRY". The problem is that in `src/chia_dialect.rs` all the integer to operator look-ups are done through symbols, which don't live past the compilation phase, so there's no way to actually export those symbols to python without some serious changes. Originally there was an `operator_lookup` table that was passed into the python version of clvm, and for a while there was even a version that got passed into rust, but Arvid was not a big fan and eventually replaced it with this static table.

Putting it back in would be a lot of work with not a clear gain, especially since `clvm_tools` is deprecated (and it was the main user of `operator_lookup`, adding `com` and `opt` among others). So I think this is the unfortunate compromise we're stuck with for now.

The main reason for exporting these is for use in a `disassemble` function (which is not included in this repo).

wheel/python/clvm_rs/program.py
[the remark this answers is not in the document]
arvidn: in what case is `self._cached_serialization` anything but `bytes`?
the type of `_cached_serialization` is `Optional[bytes]`. The only two types it can have is `None` or `bytes`. So, only if `sexp_to_bytes()` return `None` would you enter this if-block. But a more intuitive way of saying that would be `if self._cached_serialization is not None`

wheel/python/clvm_rs/program.py
```diff
@@ -0,0 +1,350 @@
...
+    UNSAFE_MAX_COST: Optional[int] = None
+
+    curry_treehasher: CurryTreehasher = CurryTreehasher(CHIA_DIALECT)
+    _cached_serialization: Optional[bytes]
+
+    # serialization/deserialization
+
+    @classmethod
+    def parse(cls, f: BinaryIO) -> Program:
+        return sexp_from_stream(f, cls.new_pair, cls.new_atom)
```
arvidn, line 32: `parse()`, `from_bytes()` and `stream()` resemble the protocol we use for `Streamable` types. It doesn't seem appropriate to use this type in a streamable structure though, since this function, `parse()` looks quite slow.
maybe it could be made faster by making a copy of the bytes and then use the clvm tree method of parsing.
I suppose there would have to be some glue in `Streamable`, which could call `from_bytes_with_cursor()` instead.

wheel/python/clvm_rs/clvm_tree.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # which of the clvm objects in the tree this instance represents
+    index: int  # index into `int_tuples`
+
+    # the sha256 tree hashes that are optionally created on deserialization
+    tree_hashes: Optional[List[bytes]]
+
+    @classmethod
+    def from_bytes(cls, blob: bytes, calculate_tree_hash: bool = True) -> "CLVMTree":
+        int_tuples, tree_hashes = deserialize_as_tuples(
```
arvidn, line 65: you parse the whole tree structure eagerly here. In case the program is just run, this is a waste of time. It would seem more appropriate to do this lazily, don't you think?

You also compute the tree hashes eagerly by default, that also seems unintuitive to me. Surely that can't be a win on average, is it?

wheel/python/clvm_rs/program.py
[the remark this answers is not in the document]
richardkiss: The original idea is that in some cases it may be `memoryview` because that has the nice property that slices are cheap (whereas `bytes` slices do a copy). It's been a while since I looked at this so I'm not sure offhand if there are benefits to this approach and in what cases they occur.

wheel/python/clvm_rs/program.py
richardkiss: I agree this is still slow. I have a version of this `parse` function in a separate branch that addresses this issue. It wraps the `BinaryIO` to `tee` the blob to a `io.BytesIO` as things are pulled out of it, and then quickly skips over all the bytes using a simple `0xff` counter algorithm like that in `SerializedProgram`. Now we've turned the streamed program into a blob, and we can parse it using `from_bytes`. There are python and rust implementations, although both need to repeatedly call `f.read` from python to get bytes out.

I didn't include it in this patch because I'm not completely happy with it yet, and this patch is already large and complex enough. Baby steps.

wheel/python/clvm_rs/clvm_tree.py
```diff
@@ -0,0 +1,114 @@
...
+        start, end, atom_offset = self.int_tuples[self.index]
+        if self.blob[start] == 0xFF:
+            self.atom = None
+        else:
+            self.atom = bytes(self.blob[start + atom_offset:end])
+            self._pair = None
+
+    @property
+    def pair(self) -> Optional[Tuple["CLVMStorage", "CLVMStorage"]]:
+        if not hasattr(self, "_pair"):
```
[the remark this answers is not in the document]
richardkiss: The idea is `._pair` is lazily created on first call to `.pair` since generating it at `__init__` time would cause the entire tree to be instantiated. So what seems like just a little bit of laziness is actually a whole lot of laziness.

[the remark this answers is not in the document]
richardkiss: I'll add a comment.

wheel/python/clvm_rs/clvm_tree.py
richardkiss: My tests suggest the parsing will be fast enough to not be a big deal in most cases. Calculating the hash lazily is much slower because it has to be done in python (so we can check subtrees for cached hashes). Even if ported to rust, it would still have to context switch repeatedly into python to call `.atom` and `.pair` down the tree. At parse-time, we can stay in rust the entire time.

We could make `.parse` have a different default for `calculate_tree_hash` if this is implicitly called during consensus. Or have you completed your port to rust of deserializing these blobs?

For wallet use, deserialization will mostly be done on puzzle template `MOD` files, once per mod at start-up, and having the hashes is useful for doing fast zero-knowledge currying.

Opting out when performance is crucial is still possible, but I don't know if any cases of this still exist.

wheel/python/clvm_rs/program.py
```diff
@@ -0,0 +1,350 @@
...
+    @classmethod
+    def from_bytes(cls, blob: bytes, calculate_tree_hash: bool = True) -> Program:
+        obj, cursor = cls.from_bytes_with_cursor(
+            blob, 0, calculate_tree_hash=calculate_tree_hash
+        )
+        return obj
+
+    @classmethod
+    def from_bytes_with_cursor(
+        cls, blob: bytes, cursor: int, calculate_tree_hash: bool = True
```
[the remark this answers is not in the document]
richardkiss: See above.

wheel/python/clvm_rs/program.py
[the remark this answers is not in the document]
arvidn: I see. so `_cached_serialization` should probably have the type: `Optional[Union[bytes, memoryview]]` then. and I would expect `sexp_to_bytes()` would return `Union[bytes, memoryview]`, or something like that

wheel/python/clvm_rs/clvm_tree.py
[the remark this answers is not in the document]
arvidn: ok, so once `_pair` is created, it may be created as `None`, and to distinguish a `None` pair and one that hasn't been computed yet, you us `hasattr()`. In other languages I think this would be expressed as an `Optional[Optional[...]]`. but perhaps that's not possible in python.

wheel/python/clvm_rs/clvm_tree.py
arvidn: > My tests suggest the parsing will be fast enough to not be a big deal in most cases.

From my experience of fixing performance issues in the full node and the wallet, deserializing objects is one of our main bottlenecks.

> Even if ported to rust, it would still have to context switch repeatedly into python to call .atom and .pair down the tree

I'm sure there are ways to avoid that, and stay in rust even when calculating it lazily.

> Or have you completed your port to rust of deserializing these blobs?

Everything has a rust counterpart, and it can be serialized and deserialized. The two main issues are:
1. the rust types are just the data, not all the functionality that's currently implemented as member functions. `Program` is also just a blob of bytes.
2. BLS types, G1Element, G2Element etc. are just byte blobs in rust, and it's quite expensive to parse them into the internal representation of relic, so that can't really be done on the fly. So to use the rust types in python requires a lot of changes to the class interfaces

> For wallet use, deserialization will mostly be done on puzzle template MOD files

Since those only happens once, on startup, I would think incoming Spends would be a much more common case for parsing puzzles and solutions.

> I just realized, the only implementation you see here is python. I have a rust version of the tree parsing that's not included in this patch.

Is there a good reason to have a python implementation at all? If there is a fast version, having a slow version seems risky, as someone might use it by accident.

richardkiss: I wrote the python implementation first as a reference implementation. It's much easier to read and to understand. It's pretty hard to use it by accident.

If for some reason someone in future wants to run in pure python, this gives them a way to do so. I see rust as an accelerator, and whenever we can make it an optional feature, I think it makes sense to do so. I bet there is some marginal platform out there that has better support for python than rust.

richardkiss: I can do further work down the line to separate out parsing and hashing, and in fact, I think it may make the code a lot simpler. I don't know if there will be a performance penalty, as we might have to load each atom into cache twice instead of once.

That said, do understand that merging this will have no impact on `chia-blockchain` until we actually do the next step of trying to replace `clvm` with `clvm_rs`. It might be more enlightening to run benchmarks on that PR and try to improve things then.

arvidn (approved):
It's a real risk that we might end up using the python implementation by accident. We don't have a lot of performance benchmark coverage in CI, so it can easily go unnoticed. We had this problem with clvm_tools using the python implementation. So I'm skeptical of the value of having it, since we most likely don't want to use it.

My hunch is that the interface between python and rust is cut at a place that inherently will make it hard to maximize performance.

PR #260, opened 2023-03-04
Update dependencies

richardkiss:


arvidn (approved):
I think the `fuzz` crate may need updates too. at least its `.lock` file will be updated I believe

richardkiss:
> I think the `fuzz` crate may need updates too. at least its `.lock` file will be updated I believe

Fuzzers don't have a lock file. I did update its `Cargo.toml` and the `wasm` stuff too though.

arvidn:
oh, right. there's only one lock file in the root

arvidn (approved):

PR #269, opened 2023-04-19
optimize parsing the cost argument to softfork

arvidn:
We know it may not exceed u64::MAX, so no need to allocate a BigInt.

I also added a dev-dependency on `rstest` for parameterizing the test for the new function, `uint_atom()`.

### DIFF
--- a/src/more_ops.rs
+++ b/src/more_ops.rs
@@ -1,7 +1,6 @@
 use bls12_381::{G1Affine, G1Projective, Scalar};
 use num_bigint::{BigUint, Sign};
 use num_integer::Integer;
-use std::convert::TryFrom;
 use std::ops::BitAndAssign;
 use std::ops::BitOrAssign;
 use std::ops::BitXorAssign;
@@ -14,7 +13,7 @@ use crate::err_utils::err;
 use crate::node::Node;
 use crate::number::{number_from_u8, ptr_from_number, Number};
 use crate::op_utils::{
-    arg_count, atom, check_arg_count, i32_atom, int_atom, two_ints, u32_from_u8,
+    arg_count, atom, check_arg_count, i32_atom, int_atom, two_ints, u32_from_u8, uint_atom,
 };
 use crate::reduction::{Reduction, Response};
 use crate::sha2::{Digest, Sha256};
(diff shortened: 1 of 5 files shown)


src/op_utils.rs
```diff
@@ -73,6 +73,128 @@ pub fn int_atom<'a>(args: Node<'a>, op_name: &str) -> Result<&'a [u8], EvalErr>
...
+    if bytes.is_empty() {
+        return Ok(0);
+    }
+
+    if (bytes[0] & 0x80) != 0 {
+        return args.err(&format!("{op_name} requires positive int arg"));
+    }
+
+    // strip leading zeros
+    let mut buf: &[u8] = bytes;
```
xdustinface, line 93: It seems confusing to require parsing `SIZE` bytes in this function but first drop potentially an infinite number of zero bytes here and then require the rest to be `SIZE` length but i guess thats some part of the protocol which im missing?

xdustinface (changes requested):
Given you test only parsing `1` and with leading zeros and max size with `ff` i think you don't really test the integer gets built in the correct byte order somewhere?

src/op_utils.rs
arvidn: the `SIZE` parameter specifies the limit of the variable the number is being parsed into. 8 bytes for `u64`, 4 bytes for `u32`. Leading zeroes do not contribute to the magnitude of the value, so they are ignored. This is part of how we interpret byte buffers as integers in clvm

arvidn:
updated with more test cases

xdustinface (approved):

src/more_ops.rs
```diff
@@ -791,16 +790,14 @@ pub fn op_softfork(a: &mut Allocator, input: NodePtr, max_cost: Cost) -> Respons
...
-            let n: Number = number_from_u8(int_atom(p1, "softfork")?);
-            if n.sign() == Sign::Plus {
-                if n > Number::from(max_cost) {
-                    return err(a.null(), "cost exceeded");
-                }
-                let cost: Cost = TryFrom::try_from(&n).unwrap();
-                Ok(Reduction(cost, args.null().node))
-            } else {
-                args.err("cost must be > 0")
+            let cost = uint_atom::<8>(&p1, "softfork")?;
```
richardkiss, line 793: Can we maybe call this `u64_atom`?

richardkiss (changes requested):
Can we call the new function `u64_atom` (since `uint` isn't really a thing in rust)?

richardkiss:
Wait, I realize now `uint_atom` has a size parameter. Maybe you just add `u64_atom` that calls `uint_atom::<8>`. But I will approve for now.

richardkiss (approved):

PR #271, opened 2023-04-20
fix lifetime annotation on int_atom() and atom()

arvidn:
to make the returned buffer depend on the allocator's lifetime, not the node's.

### DIFF
--- a/src/node.rs
+++ b/src/node.rs
@@ -19,7 +19,7 @@ impl<'a> Node<'a> {
         self.allocator.sexp(self.node)
     }
 
-    pub fn atom(&'a self) -> Option<&'a [u8]> {
+    pub fn atom(&self) -> Option<&'a [u8]> {
         match self.sexp() {
             SExp::Atom(_) => Some(self.allocator.atom(self.node)),
             _ => None,
--- a/src/op_utils.rs
+++ b/src/op_utils.rs
@@ -66,15 +66,14 @@ fn test_arg_count() {
     assert_eq!(arg_count(&count_3_args, 4), 3);
 }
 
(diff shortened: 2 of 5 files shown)


src/node.rs
```diff
@@ -19,7 +19,7 @@ impl<'a> Node<'a> {
         self.allocator.sexp(self.node)
     }
 
-    pub fn atom(&'a self) -> Option<&'a [u8]> {
```
arvidn, line 22: this was a mistake. The returned buffer depends on the allocator in `self` (which is what `'a` refers to in the outer scope). This introduces a new `'a`, referring to `self`'s lifetime, shadowing the correct one.

src/op_utils.rs
```diff
@@ -66,15 +66,14 @@ fn test_arg_count() {
...
+pub fn int_atom<'a>(args: Node<'a>, op_name: &str) -> Result<&'a [u8], EvalErr> {
     match args.atom() {
         Some(a) => Ok(a),
         _ => args.err(&format!("{op_name} requires int args")),
     }
 }
 
-// rename to atom()
-pub fn atom<'a>(args: &'a Node, op_name: &str) -> Result<&'a [u8], EvalErr> {
+pub fn atom<'a>(args: Node<'a>, op_name: &str) -> Result<&'a [u8], EvalErr> {
```
arvidn, line 76: these are the two main changes, the other changes fall out of these.
Note that the returned buffer depends on the allocator's lifetime (which is the lifetime parameter to `Node` itself) not on the reference to the node itself (which is what we used to do).

The same goes for `int_atom()` above.

These changes allow passing in temporary `Node` objects to these functions, and make code using them more compact, without unnecessary new variables.

richardkiss (approved):
"Borrow checker for the win"

PR #273, opened 2023-04-21
Softfork infrastructure and coinid operator

arvidn:
There are two parts to this PR, best reviewed seperately;
1. the infrastructure to support the `softfork` operator.
   * Moving the softfork logic into the interpreter, to allow giving it apply-semantics, and the unit tests are moved from `test_ops` into `run_program` (since the former doesn't run the interpreter)
   * Adding a stack of softfork guard frames, used to record the expected cost to check against when exiting a guard. When exiting a softfortk guard, the allocator is also reset to the state of when entering it, to reclaim memory
   * Parsing and checking its arguments, preserving the existing relaxed rules in consensus mode.
   * validating the cost argument (when execution completes)
   * The cost of the softfork operator is 140, slightly higher than apply
2. Add the `coinid` operator as part of softfork extension 0
   * Having a valid extension value enables testing the validation of the cost argument (so, more tests are added in this PR)


src/run_program.rs
```diff
@@ -57,6 +59,26 @@ impl Counters {
     }
 }
 
+// this represents the state we were in before entering a soft-fork guard. We
+// may need this to long-jump out of the guard, and also to validate the cost
+// when exiting the guard
+struct SoftforkGuard {
```
richardkiss, line 65: Do we still need this when there is no long-jump capabilities?

src/dialect.rs
```diff
@@ -2,10 +2,26 @@ use crate::allocator::{Allocator, NodePtr};
 use crate::cost::Cost;
 use crate::reduction::Response;
 
+#[repr(u32)]
+#[derive(Clone, Copy, Eq, PartialEq)]
+pub enum Extension {
```
richardkiss, line 7: I'm trying to understand why we need this new structure. I'd always figured the `Dialect` object would be the thing that changes based on extension. So you have the base dialect, which is as we see, then a dialect level 0 which adds `coinid` and the bls12 ops. Is there something that makes `Extension` necessary? Isn't it just sort of an add-on to the `Dialect`?

src/run_program.rs
```diff
@@ -66,6 +88,7 @@ struct RunProgramContext<'a, D> {
     val_stack: Vec<NodePtr>,
     env_stack: Vec<NodePtr>,
     op_stack: Vec<Operation>,
+    softfork_stack: Vec<SoftforkGuard>,
```
richardkiss, line 91: As per my comment above, why not just use a stack of dialects, and at the end of the softfork eval process, pop off the top dialect.

src/chia_dialect.rs
```diff
@@ -21,6 +21,10 @@ pub const LIMIT_HEAP: u32 = 0x0004;
 // When set, enforce a stack size limit for CLVM programs
 pub const LIMIT_STACK: u32 = 0x0008;
 
+// When set, we allow softfork with extension 0 (which includes coinid and the
+// BLS operators)
+pub const ENABLE_BLS_OPS: u32 = 0x0010;
```
richardkiss, line 26: Ah, I see, as we move through the blockchain consensus, we have to adjust how the softfork operator works.

I feel like this can be completely encapsulated in `Dialect`. For simplicity, suppose a soft fork kicks in that introduces a new dialect that supports `coinid`.

src/chia_dialect.rs
```diff
@@ -56,12 +60,15 @@ impl Dialect for ChiaDialect {
         o: NodePtr,
         argument_list: NodePtr,
         max_cost: Cost,
+        extensions: Extension,
```
richardkiss, line 63:
```suggestion
        extension: Extension,
```

src/chia_dialect.rs
```diff
@@ -117,11 +127,33 @@ impl Dialect for ChiaDialect {
         &[2]
     }
 
+    fn softfork_kw(&self) -> &[u8] {
+        &[36]
+    }
+
+    fn softfork_extension(&self, ext: u32) -> Extension {
```
richardkiss, line 134: What about passing in the `Extension` directly instead of using flags?

src/more_ops.rs
```diff
@@ -877,3 +864,45 @@ pub fn op_point_add(a: &mut Allocator, input: NodePtr, max_cost: Cost) -> Respon
...
+        return args.err("coinid: invalid parent coin id (must be 32 bytes)");
+    }
+    let args = args.rest()?;
+    let puzzle_hash = atom(args.first()?, "coinid")?;
+    if puzzle_hash.len() != 32 {
+        return args.err("coinid: invalid puzzle hash (must be 32 bytes)");
+    }
+    let args = args.rest()?;
+    let amount = atom(args.first()?, "coinid")?;
+    if !amount.is_empty() {
```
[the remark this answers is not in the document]
richardkiss: You might consider checking the assembly generated... it's possible the optimizer might realize you don't use the return value. Also, since it likely just goes into a register anyway, it's probably not particularly expensive.

(I may be being a bit naïvely optimistic about how good the optimizers are.)

src/run_program.rs
```diff
@@ -312,16 +364,100 @@ impl<'a, D: Dialect> RunProgramContext<'a, D> {
...
+            }
+            if expected_cost == 0 {
+                return err(self.allocator.null(), "cost must be > 0");
+            }
+
+            // we can't blindly propagate errors here, since we handle errors
+            // differently depending on whether we allow unknown ops or not
+            let (ext, prg, env) = match self.parse_softfork_arguments(&operand_list) {
+                Ok(ret_values) => ret_values,
+                Err(err) => {
```
richardkiss, line 381: Is this a potentially temporary workaround here?

src/run_program.rs
```diff
@@ -347,9 +493,11 @@ impl<'a, D: Dialect> RunProgramContext<'a, D> {
...
             };
             cost += match op {
-                Operation::Apply => {
-                    augment_cost_errors(self.apply_op(max_cost - cost), max_cost_ptr)?
-                }
+                Operation::Apply => augment_cost_errors(
+                    self.apply_op(cost, effective_max_cost - cost),
+                    max_cost_ptr,
+                )?,
+                Operation::Softfork => self.exit_softfork_guard(cost)?,
```
richardkiss, line 500: The names are a bit inconsistent here. Maybe rename the operator to match the name of the function it calls? `exit_guard` & `ExitGuard`, or `EndSoftfork` and `end_soft_fork`, or whatever.

src/test_ops.rs
```diff
@@ -878,6 +856,27 @@ pubkey_for_exp 0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00f0000
 pubkey_for_exp 0x8c1258acd66282b7ccc627f7f65e27faac425bfd0001a40100000000fffffffe => 0xb7f1d3a73197d7942695638c4fa9ac0fc3688c4f9774b905a14e3a3f171bac586c55e83ff97a1aeffb3af00adb22c6bb | 1327426
 pubkey_for_exp 0x8c1258acd66282b7ccc627f7f65e27faac425bfd0001a40100000000ffffffff => 0xc00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 | 1327426
 pubkey_for_exp 0x8c1258acd66282b7ccc627f7f65e27faac425bfd0001a4010000000000000000 => 0x847f5fcce0b9aa0f2bb3de6847337c9ed1bc2184a125c232721e1c81b0f0fee78506790a78c98abff2dd4b01a0756352 | 1327426
+
+; BLS extensions
```
richardkiss, line 860:
```suggestion
; BLS extensions could go here
```

richardkiss (commented):
I think `Extension` is a bit vague and could do with a different name. To me, it seems like it represents the maximum soft fork version number known to this current runtime. Maybe `AdditionalOpcodeSet` or something like that?

src/more_ops.rs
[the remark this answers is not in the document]
arvidn: I looked into splitting the validation and the parsing of the integer and realized that another important difference is that `coinid` requires the amount to be in canonical integer form. i.e. it does not allow any redundant leading zero. This is an important rule since it affects the hash.

The only shared check is really that the value is not negative. There are some utility functions to sanitize integers in `chia_rs`, for parsing conditions. Perhaps in the future (especially as we have more operators in here) it might make sense to move those into this repo and have some stricter conventions for validating arguments. I don't think there are enough places we do this in right now though.

Even your suggestion of calling `uint_atom::<8>()` just for the validation won't work, because redundant leading zeros would still have to be checked.

src/dialect.rs
arvidn: in a way `Dialect` might be a pretty good name for this enum, but it's taken

src/run_program.rs
```diff
@@ -57,6 +60,26 @@ impl Counters {
...
+    // current_cost + the first argument to the operator
+    expected_cost: Cost,
+
+    // When exiting a softfork guard, all values used inside it are zapped. This
+    // was the state of the allocator before entering. We restore to this state
+    // on exit.
+    allocator_state: Checkpoint,
+
+    // this specifies which new operators are available
+    extension: Operators,
```
richardkiss, line 77: Consider renaming to `operators`.

richardkiss (approved):
Seems good enough for now. If you want to keep working on it, that's fine too. Let me know if you need further approvals.

One comment I want to make is that `coinid` probably isn't really that useful of a thing to be behind the softfork guard... probably repurposing an unused operator is a better way to do it (although it would have to be `assert_coinid`). It'd take the three fields along with a hash, and it'd assert the field sizes and that the hash is correct. Since `coinid` is stuck behind the guard, either the whole puzzle has to be there or the hash has to be passed in so it can confirm it's right, effectively making it `assert_coinid`. We should probably discuss if there's any use to adding this operator as a guarded softfork op. For now, it's a decent test case.

PR #277, opened 2023-05-01
raise the abstraction level of Allocator to support Numbers directly

arvidn:
not just Atoms as byte buffers.

The idea is to open up the possibility for the allocator to store atoms in their "native" forms, and only lazily convert them to a requested form. Hopefully reducing the time spent serializing and deserializing numbers.

One common reason to ask for bytes for an atom is to account for the cost of operations, based on the number of bytes in the atom. To avoid forcing a serialization for this case, the Allocator also gains a function to request the length of an atom directly. This is presumably cheaper than serializing a number, just to count the number of bytes.

The utility function `int_atom()` was extended to return, not just the `Number` but also the length of the atom, for cost calculation purposes.

This change further sets a precedence to also support other types natively, such as G1, G2 and Gt points.

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -1,4 +1,5 @@
 use crate::err_utils::err;
+use crate::number::{node_from_number, number_from_u8, Number};
 use crate::reduction::EvalErr;
 
 pub type NodePtr = i32;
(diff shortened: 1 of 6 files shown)


AmineKhaldi (approved):
Looks great!

richardkiss (approved):

PR #278, opened 2023-05-01
move operator test cases out into separate text files

arvidn:
The test cases are move verbatim into the text files.

The main rationale for this is to make it easier to automate the creation of test cases. I'm specifically considering the existing (and forthcoming) BLS operations, where we need to ensure all operations produce identical results as `blspy`. I intend to generate test cases using `blspy` for this.

To ensure the test picks up the right files and runs the right tests, I ran:

```
cargo test -- --nocapture test_ops::test_ops
```

Which prints all tests being exercised.


src/test_ops.rs
```diff
@@ -1062,6 +197,10 @@ fn run_op_test(op: &Opf, args_str: &str, expected: &str, expected_cost: u64) {
 
 #[test]
 fn test_ops() {
+    use std::fs::read_to_string;
+
+    const TEST_FILES: &[&str; 3] = &["test-core-ops.txt", "test-more-ops.txt", "test-bls-ops.txt"];
```
richardkiss, line 202: Consider abstracting away this list into a root text file list. Also consider printing the number of files and the total number of tests run at the end of the output (like `pytest` does). It's a useful sanity check that we're not suddenly running way too few tests.

richardkiss (approved):
Looks good. Another useful thing to print is the

src/test_ops.rs
arvidn: `cargo` hides all output for successful runs. Do you mean the test would have a *minimum* number of test for each file, and assert that we don't run fewer?

richardkiss: Yeah, you'd still have to use the `--nocapture` trick to see the output. I'm just sayin' print out the total number of tests run.

The other change I didn't describe very well would be to have a file called `list_of_tests.txt` (or something like that) which has three lines for now: `test-core-ops.txt`, `text-more-ops.txt` and `test-bls-ops.txt`.

Or maybe even better, you just look for files that match `test-*txt`.

Using either of these methods, you could even pawn of the work of writing even more test cases to people who don't know rust.

arvidn: I'll address this in a follow-up PR

PR #279, opened 2023-05-05
More fuzzing

arvidn:
Please review commits individually.

In this patch set:

* the fuzzer for `run_program()` is extended to cover more configuration flags
* a new fuzzer directly targeting operator functions is added
* a tool to generate an initial corpus for `run_program()` is added, that knows about all operators and the shape of their operands
* the CI workflow is updated to run the tool generating the corpus, when fuzzing


Cargo.toml
```diff
@@ -1,3 +1,7 @@
+[workspace]
```
richardkiss, line 1: I feel like there was some reason we didn't do this before, but I don't know what it could be.

tools/src/bin/generate-fuzz-corpus.rs
```diff
@@ -0,0 +1,335 @@
...
+    // coinid
+    op(
+        48,
+        &[Type::Bytes32, Type::Bytes32, Type::Int64],
+        Type::Bytes32,
+    ),
+];
+
+const ZEROS: [u8; 288] = [
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
richardkiss, line 144:
```suggestion
    [0; 288];
```

richardkiss (approved):
Nice

Cargo.toml
arvidn: `maturin` pyo3 doesn't like to be built as part of a workspace, but adding `wheel` to the exclude key makes it work

tools/src/bin/generate-fuzz-corpus.rs
arvidn: I will update this in a future PR

PR #280, opened 2023-05-06
add option to make BLS extensions available outside the softfork operator

arvidn:
The BLS extensions only contain `coinid` for now. This is a hard fork and intended for when the 2.0 hard fork activates.

### DIFF
--- a/src/chia_dialect.rs
+++ b/src/chia_dialect.rs
@@ -25,6 +25,10 @@ pub const LIMIT_STACK: u32 = 0x0008;
 // BLS operators). This remains disabled until the soft-fork activates
 pub const ENABLE_BLS_OPS: u32 = 0x0010;
 
+// enables the BLS ops extensions *outside* the softfork guard. This is a
+// hard-fork and should only be enabled when it activates
+pub const ENABLE_BLS_OPS_OUTSIDE_GUARD: u32 = 0x0020;
+
 // The default mode when running grnerators in mempool-mode (i.e. the stricter
 // mode)
 pub const MEMPOOL_MODE: u32 = NO_UNKNOWN_OPS | LIMIT_HEAP | LIMIT_STACK;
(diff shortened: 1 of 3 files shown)


richardkiss:
It may be nice to allow this feature to be available from `brun` or `Program.run`. I think maybe that's why I had the idea of named or versioned "dialects" so that `brun` could accept a dialect as a parameter. I suppose passing in "flags" is possible, but having magic flags seems pretty gross from a command-line perspective because it's so opaque. Do you have any thoughts on this?

src/chia_dialect.rs
```diff
@@ -103,18 +107,26 @@ impl Dialect for ChiaDialect {
...
             // 35 ---
             // 36 = softfork
-            _ => match extension {
-                OperatorSet::BLS => match b[0] {
-                    48 => op_coinid,
-                    // TODO: add BLS operators here
-                    _ => {
-                        return unknown_operator(allocator, o, argument_list, self.flags, max_cost);
+            _ => {
+                if extension == OperatorSet::BLS || (self.flags & ENABLE_BLS_OPS_OUTSIDE_GUARD) != 0
```
richardkiss, line 111: I do worry a bit about complex logic like this sprouting in the dispatcher because a small mistake can result in historical behaviour changing. This might be why I'm so biased towards just having multiple dispatchers and choosing the correct one for the given dialect. There is a chance this one will end up not being permanent though, so it's probably good for now.

richardkiss (approved):
Seems reasonable.

src/chia_dialect.rs
arvidn: I was thinking of different ways to express this to be as simple and robust as possible. I couldn't find a clear-cut improvement. For example, one could imagine this flag just controls what the initial state of `extension` is, so, instead of being `None` it's `OperatorSet::BLS`. This would make all programs be run as if they were inside the BLS softfork. However, it doesn't really model that the default operator set in the future could be a combination of other sets.

Although, now that I say it out loud; we won't have another hard fork, so that's probably not a reasonable case to consider. I'll try to make this a bit better in a future patch

arvidn:
I imagine we could add support for more modes the same way we have a `--mempool` flag right now.

PR #283, opened 2023-05-11
Set specifically the macos deployment target to 10.14

emlowe:
Specifically setting the deployment target prevents maturin from doing something strange like publishing a wheel that claims to work on macOS 10.7 when it almost certainly doesn't

Set the target to 10.14 to match the setting on all other wheels

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -46,6 +46,8 @@ jobs:
 
     - name: Build MacOs with maturin on Python ${{ matrix.python }}
       if: startsWith(matrix.os, 'macos')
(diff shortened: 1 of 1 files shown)


arvidn (approved):

arvidn:
as far as I know, only our old python `clvm_tools` uses this wheel. Did we experience any issues on MacOS 10.7?

altendky:
Presumably we did not because presumably we never ran anything there because it's hard enough (and dangerous enough) to run even just 10.14, AFAIK.  `:]`

cmmarslender (approved):

PR #285, opened 2023-05-12
correct the cost of the coinid operator to be 800

arvidn:
which is a discount of 153 compared to `sha256`.

In the original patch I had forgotten to charge cost for the heap allocation of the return value. This patch corrects that and adds a discount, to incentivize using this operator instead of a naked sha256.


Quexington (approved):
💲 💲 💲

PR #286, opened 2023-05-12
Opt for source-based code coverage

AmineKhaldi:


arvidn (approved):
I think it would be worth a try to also invoke `maturin` with `-Cinstrument-coverage` and run `pytest`, to see if that reports coverage of the python bindings as well.

PR #287, opened 2023-05-16
BLS operator signatures

arvidn:
This changes the signatures of `bls_verify` from `signature (G1 . msg) (G1 . msg) ...` to `signature G1 msg G1 msg ...`.
and the signature of `bls_pairing_identity` from `(G1 . G2) (G1 . G2) ...` to `G1 G2 G1 G2 ...`


cameroncooper (approved):

richardkiss (approved):
Looks good.

PR #288, opened 2023-05-16
Measure clvm cost

arvidn:
These commits are best reviewed one at a time.

This PR:
* adds a benchmark tool to aid in determining reasonable costs for the new BLS operators.
* records the measurements in `benchmarks.txt`
* updates the cost constants
* updates the operator unit tests to reflect the new costs
* updates the BLS test case generation script to reflect the new costs

As you can see from the `benchmarks.txt`, there's quite a lot of variance between runs, so the precision of these costs shouldn't be overestimated. I think as long as the cost is somewhat reasonable, we have quite a lot of wiggle room. I made some of the higher cost constants somewhat even numbers for instance.

Something else noteworthy is that when calibrating the timings and cost against `point_add`, the base-cost and the cost-per-arg were quite far apart. I used different scale factors for those. It's not clear to me why the previous way of measuring this would come up with different numbers, but it was quite a bit different machine (Intel Mac Laptop) compared to what I ran these one, AMD Ubuntu desktop.

I'm interested in any opinions on shifting the costs a bit. For example, richard suggested that `g1_negate` and `g2_negate` should be cheaper.


src/bls_ops.rs
```diff
@@ -11,26 +11,36 @@ use bls12_381::{multi_miller_loop, G1Affine, G1Projective, G2Affine, G2Prepared,
...
+// the same cost as point_add (aka g1_add)
+const BLS_G1_SUBTRACT_BASE_COST: Cost = 101094;
+const BLS_G1_SUBTRACT_COST_PER_ARG: Cost = 1343980;
+
+const BLS_G1_MULTIPLY_BASE_COST: Cost = 1411000;
+const BLS_G1_MULTIPLY_COST_PER_BYTE: Cost = 10;
+
+const BLS_G1_NEGATE_BASE_COST: Cost = 839000;
+
+// g1_add and g2_subtract have the same cost
```
cameroncooper, line 23: do you mean g2_add?

arvidn: I believe I did, yes

richardkiss:
> I'm interested in any opinions on shifting the costs a bit. For example, richard suggested that `g1_negate` and `g2_negate` should be cheaper.

This is based on a new implementation that just does bit-twiddling (and doesn't bother verifying for validity). Validity verification is done many other places, and if `P` is valid, the bit-twiddled version of `P` is also valid. My concern is that developers start doing the bit twiddling themselves with `logxor` to save cost.

richardkiss (approved):
Seems fine. I'm still of the opinion about simplifying and making much cheaper the `*_negate` functions.

arvidn:
yeah, it would be bad if people would start using xor instead of `g1_negate` and `g2_negate`, because that would prevent the future optimization of not converting everything to bytes eagerly.

On the other hand, it's hard to predict how expensive it would be to run with that optimization without implementing it. I'm a bit hesitant to not validate g1 points. We do that for `point_add` and I think it's a safety issue. in other words, it seems risky to widen the contract for `g1_negate` to accept invalid points.

That said, maybe we can just deduct some cost from it as a preparation for optimizing this, and disincentivizing circumventing the checks with xor.

arvidn:
@richardkiss cameron pointed out that the cost of `AGG_SIG_*` is a lot less than `bls_verify`. I'm thinking that maybe we can lower some costs in anticipation of future optimizations.

`AGG_SIG_*` is really the counterpart to cost-per-arg, since the block is one big aggregate signature. Even then, the cost of the operator is essentially 5 times as expensive.

I think it would be reasonable to cut all costs in half (except the per-byte).

Regarding the negate operators. I think the main risk with leaving them as expensive is that it creates an incentive to use xor. Using XOR will defeat the future optimization where we can skip the subgroup check, since we preserve the values. However, it's possible the negation on the native G1 and G2 point is still more expensive than the XOR.

I think the risk of actually implementing negation as XOR is that it creates a (possibly risky) loop-hole to get invalid points.

Do you have any concrete suggestions on what changes you think we should make to this PR?

richardkiss:
I think we should add `*_negate_unsafe` and make it cheaper than `logxor` so there's no risk of people code-golfing in `logxor`. Then we can soft fork one out later if necessary once the concern about points outside the subgroup is better understood. I also understand if you don't want to do this due to time constraints.

I'm fine with reducing all costs by 50%. It's at worst just a mild DOS... "slow" blocks, which we'll likely be able to mitigate with optimizations.

arvidn:
@richardkiss @cameroncooper Please take a look at the most recent commit. I made these changes:

* added test to establish cost for XORing 48 and 96 byte buffers, and made g1_negate and g2_negate match. [here](https://github.com/Chia-Network/clvm_rs/pull/288/commits/f36a46101c03f2982444c9402d0d19fccc1dd6e9#diff-a7caec4c99a34068bd1628d5475a719169a1f173123e3b730d3937e85169e7f5R68-R72)
* implemented `g1_negate` as XOR, without validating the point. [here](https://github.com/Chia-Network/clvm_rs/pull/288/commits/f36a46101c03f2982444c9402d0d19fccc1dd6e9#diff-7836437a98e99f1bb6e7ac825dea0d56ca9ec9e8a38c6220dc34882a726e74c3R115-R127).
* implemented `g2_negate` as XOR without validating the point. [here](https://github.com/Chia-Network/clvm_rs/pull/288/commits/f36a46101c03f2982444c9402d0d19fccc1dd6e9#diff-7836437a98e99f1bb6e7ac825dea0d56ca9ec9e8a38c6220dc34882a726e74c3R185-R197).
* Updated most costs by cutting them in half ([here](https://github.com/Chia-Network/clvm_rs/pull/288/commits/f36a46101c03f2982444c9402d0d19fccc1dd6e9#diff-7836437a98e99f1bb6e7ac825dea0d56ca9ec9e8a38c6220dc34882a726e74c3R18-R47)), except:
   * per-byte-costs, which remained the same
   * the pairing per-arg costs. I made those match the cost of `AGG_SIG_*` conditions (1200000). Using the operators is still a bit more expensive since there's a base cost, and to verify a signature you need two arguments (message, pk) and (signature, generator). This is reasonable because in `AGG_SIG_*`, the second arg is amortized over all puzzles in the block. 
   * the `g1_negate` and `g2_negate` match the cost of the corresponding XOR. Using these dedicated operators will still be cheaper than XOR because you don't need to include a large constant ot XOR with.

src/bls_ops.rs
```diff
@@ -172,14 +182,19 @@ pub fn op_bls_g2_multiply(a: &mut Allocator, input: NodePtr, max_cost: Cost) ->
...
-    let total = -point;
-    new_atom_and_cost(a, cost, &total.to_compressed())
+
+    // we don't validate the point. We may want to soft fork-in validating the
+    // point once the allocator preserves native representation of points
+    let blob = atom(args.first()?, "G2 atom")?;
+    if blob.len() != 96 {
+        return args.first()?.err("atom is not G2 size, 96 bytes");
+    }
+    let mut blob: [u8; 96] = blob.try_into().unwrap();
```
richardkiss, line 195: Not a big deal, but you could just make the error case of `try_into` lead into the error case in 193

richardkiss (dismissed):

src/bls_ops.rs
arvidn: hm. I didn't actually try, I just assumed I would get in trouble for the error types being incompatible

richardkiss (approved):
Looks good.

PR #292, opened 2023-05-24
added secp256k1_verify and secp256r1_verify ops

richardkiss:


src/secp_ops.rs
```diff
@@ -0,0 +1,73 @@
...
+pub fn op_secp256p1_verify(a: &mut Allocator, input: NodePtr, max_cost: Cost) -> Response {
+    let args = Node::new(a, input);
+    check_arg_count(&args, 3, "secp256p1_verify")?;
+
+    let cost = SECP256P1_VERIFY_COST;
+    check_cost(a, cost, max_cost)?;
+
+    // first argument is sec1 encoded pubkey
+    let pubkey = atom(args.first()?, "secp256p1_verify pubkey")?;
+    let Ok(verifier) = P1VerifyingKey::from_sec1_bytes(pubkey) else { return args.err("secp256p1_verify pubkey is not valid") };
```
richardkiss, line 23: This line is really long and I can't get `cargo fmt` to fix it. I don't understand why.

This might be more rust-ic:
```suggestion
    let verifier = P1VerifyingKey::from_sec1_bytes(pubkey)
        .or_else(|_| args.err("secp256p1_verify pubkey is not valid"))?;
```

richardkiss: https://doc.rust-lang.org/beta/rust-by-example/flow_control/let_else.html#let-else

I'd recommend using `.or_else` as it's a bit more standard and lets us use `clippy`.

src/secp_ops.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    // second arg is message
+    let args = args.rest()?;
+    let msg = atom(args.first()?, "secp256p1_verify msg")?;
+
+    // third arg is der encoded sig
+    let args = args.rest()?;
+    let sig = atom(args.first()?, "secp256p1_verify sig")?;
+    let Ok(sig) = P1Signature::from_slice(sig) else { return args.err("secp256p1_verify sig is not valid") };
```
richardkiss, line 32:
```suggestion
    let sig =
        P1Signature::from_slice(sig).or_else(|_| args.err("secp256p1_verify sig is not valid"))?;
```

src/secp_ops.rs
```diff
@@ -0,0 +1,73 @@
...
+pub fn op_secp256k1_verify(a: &mut Allocator, input: NodePtr, max_cost: Cost) -> Response {
+    let args = Node::new(a, input);
+    check_arg_count(&args, 3, "secp256k1_verify")?;
+
+    let cost = SECP256K1_VERIFY_COST;
+    check_cost(a, cost, max_cost)?;
+
+    // first argument is sec1 encoded pubkey
+    let pubkey = atom(args.first()?, "secp256k1_verify pubkey")?;
+    let Ok(verifier) = K1VerifyingKey::from_sec1_bytes(pubkey) else { return args.err("secp256k1_verify pubkey is not valid") };
```
richardkiss, line 54: see above

src/secp_ops.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    // second arg is message
+    let args = args.rest()?;
+    let msg = atom(args.first()?, "secp256k1_verify msg")?;
+
+    // third arg is der encoded sig
+    let args = args.rest()?;
+    let sig = atom(args.first()?, "secp256k1_verify sig")?;
+    let Ok(sig) = K1Signature::from_slice(sig) else { return args.err("secp256k1_verify sig is not valid") };
```
richardkiss, line 63: see above

arvidn (commented):
It would be good to also:
* extend `src/bin/generate-fuzz-corpus.rs` to generate calls to the new operators and their signatures.
* extend `fuzz/fuzz_targets/operators.rs` to include the new operators
* similar to the BLS operators, I would think we want to generate test vectors using a second implementation, to ensure equivalence
* unless you actually want to make these part of the same extension as BLS, you would also need a flag to enable and disable these. `chia-blockchain` would then pass that flag in once the soft-fork activates. We usually use separate flags for separate features, even when they target the same soft-fork. this gives us some agility in case we need to disable one last minute.

PR #294, opened 2023-05-26
make G1 and G2 points first class objects in the Allocator

arvidn:
To prepare for a more efficient allocator, where the internal representation can be preserved across CLVM calls.

Is it correct to keep `G1Projective` and `G2Projective` as the internal representation?

note that `g1_negate` and `g2_negate` are not changed. They still operate on byte buffers. Maybe that should change.

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -1,6 +1,7 @@
 use crate::err_utils::err;
 use crate::number::{node_from_number, number_from_u8, Number};
 use crate::reduction::EvalErr;
+use bls12_381::{G1Affine, G1Projective, G2Affine, G2Projective};
 
 pub type NodePtr = i32;
 
@@ -152,6 +153,16 @@ impl Allocator {
         node_from_number(self, &v)
     }
 
+    pub fn new_g1(&mut self, g1: G1Projective) -> Result<NodePtr, EvalErr> {
+        let g1: G1Affine = g1.into();
+        self.new_atom(&g1.to_compressed())
+    }
+
+    pub fn new_g2(&mut self, g2: G2Projective) -> Result<NodePtr, EvalErr> {
+        let g2: G2Affine = g2.into();
+        self.new_atom(&g2.to_compressed())
+    }
+
     pub fn new_pair(&mut self, first: NodePtr, rest: NodePtr) -> Result<NodePtr, EvalErr> {
         let r = self.pair_vec.len() as i32;
         if self.pair_vec.len() == self.pair_limit {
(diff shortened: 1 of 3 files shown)


src/allocator.rs
```diff
@@ -152,6 +153,16 @@ impl Allocator {
         node_from_number(self, &v)
     }
 
+    pub fn new_g1(&mut self, g1: G1Projective) -> Result<NodePtr, EvalErr> {
+        let g1: G1Affine = g1.into();
```
richardkiss, line 157: It's going to be a lot faster if you simply leave it as projective.

src/allocator.rs
```diff
@@ -245,6 +256,44 @@ impl Allocator {
         number_from_u8(self.atom(node))
     }
 
+    pub fn g1(&self, node: NodePtr) -> Result<G1Projective, EvalErr> {
```
richardkiss, line 259: If we can get this change in before the hard fork, we can be a lot more accurate by charging for serialization/deserialization. Converting `g1` to/from `bytes` is very expensive, so a DOS attack would be to convert back and forth from `g1` to `bytes`. A genuine usage would just do a bunch of `g1` calculations with no intermediate conversions. It would be fantastic to be able to charge less for this. But it's obviously a hard fork.

All this same reasoning could be done with `Number`.

src/allocator.rs
```diff
@@ -152,6 +153,16 @@ impl Allocator {
         node_from_number(self, &v)
     }
 
+    pub fn new_g1(&mut self, g1: G1Projective) -> Result<NodePtr, EvalErr> {
+        let g1: G1Affine = g1.into();
+        self.new_atom(&g1.to_compressed())
```
richardkiss, line 158: Ideally, we do this lazily. It's possible no one ever asks for the bytes representation of this object.

src/allocator.rs
arvidn: great! I'll make that change

src/allocator.rs
arvidn: yes. My plan is to first raise the abstraction level of the Allocator (which this patch is part of). Then I update the Allocator to preserve the native representations of these types in memory.

This function is a temporary step along that path.

src/allocator.rs
arvidn: right. I'm a bit hesitant to expose so much implementation details in the consensus logic (i.e. cost model). You're right that there's a risk of not doing that. I'm thinking that one possibility would be to convert lazily, but also cache all representations of a value, so at least you can't convert the same value back-and-forth

src/allocator.rs
arvidn: oh, maybe I misunderstood you. Are you're referring to the serialization into bytes, or the API taking a `G1Projective` instead of `G2Projective`.

Would it make more sense for this API to take `G1Affine` or `G1Projective`? My assumption was that `G1Projective` is the most complete state, and most efficient to keep intermediate values in.

richardkiss: Yes, conversion from `G1Projective` (X, Y, Z) to `G1Affine` (x, y) is costly. The other way is nearly free.

Operating on `G1Projective` is fastest, although uses more memory. You generally want to convert to `G1Affine` before serializing or comparing as `G1Affine` serializes smaller and has canonical representations.

You can have two `G1Projective` points that look very different but are the same. You convert to `G1Affine` to canonicalize.

src/allocator.rs
richardkiss: Okay, so for now, I suppose using `G1Affine` is reasonable since all you're really doing is generating the atom blob. Once we cache the internal representations, we definitely should cache them as `G1Projective`. We only ever should convert to `G1Affine` if someone wants the atom blob representation.

You might even consider also consider caching `G1Affine` representations, as the cheapest op form is to take affine coordinates as inputs and generate projective coordinates as output (since z=1 always for affine, which is why it's elided). But this is a hyper-optimization and I'd say not really necessary.

src/allocator.rs
richardkiss: > You're right that there's a risk of not doing that.

It's not so much a risk (we'll still charge as IF the conversion is done every time) as a lack of reward for not trying to DOS. If we don't do this, then users doing legit computation that doesn't require conversion still have to pay the price they would if they were trying to DOS even though we have the capability to better notice this. It's just leaving potential agony cost on the table. But it's not the end of the world. It just makes actual DOS attempts more noticeable.

richardkiss (approved):
For what this is (an intermediate stepping-stone), this is fine.

src/allocator.rs
arvidn: I don't believe this PR is making anything worse than it already is btw. The intention is to move the conversions into the Allocator, to give the allocator control over whether and when conversions are made ,in a later patch. That later patch will be quite large and complex, so I'm trying to front-load as many changes as possible that are somewhat independent

PR #298, opened 2023-05-31
add zksnark test for the BLS operators

arvidn:
This is a proof-of-concept of a ZK-rollup proof implemented using the new `bls_pairing_identity` operator.
The JSON files were generated from snarkjs, there's a rust program that converts those keys into hex form and prints the CLVM program.

based on Cameron's patch.
* moves the code into `tools/src/bin` and the corresponding data files into `tools`.
* fixes `fmt` and `clippy` warnings
* moves the operator test itself into its own file


tools/src/bin/verify-zksnark.rs
```diff
@@ -0,0 +1,147 @@
...
+    let data: [u8; 192] = [fp_2, fp_1, fp_4, fp_3].concat().try_into().unwrap();
+    let p = G2Affine::from_uncompressed(&data);
+    p.unwrap()
+}
+
+pub fn main() {
+    println!("verifying zksnark");
+
+    // Read verification_key.json
+    let mut file = File::open("data/verification_key.json").unwrap();
```
xdustinface, line 73: Seems like there could be a helper for reading and parsing a file.

xdustinface (approved):
I can't really judge about the zksnark functionality at this point but the code looks good.

PR #301, opened 2023-06-01
make AtomBuf private, raise abstraction level of Allocator

arvidn:
Not exposing offsets into the internal heap affords the allocator more freedom in its internal data structure.
This is a step towards preserving internal representations of `Number`, G1 and G2 points.


richardkiss (approved):
Clever

PR #303, opened 2023-06-05
introduce secp operators

arvidn:
This is a softfork activated by `ENABLE_SECP_OPERATORS`. It's based on @cameroncooper 's patch which implements verification of secp256k1 and secp256r1 signatures.

The two new operators are:

| name | opcode | cost | description |
| --- | --- | --- | --- |
| `secp256k1_verify` | 0x0cf84f00 | 850000 | Verifies a signature given a public key, message digest and signature |
| `secp256r1_verify` | 0x1c3a8f00 | 1850000 | Verifies a signature given a public key, message digest and signature |

The message digest parameter must be the sha256 of the message to sign.

This is best reviewed one commit at a time.


arvidn:
@rostislav pointed out that there's a reason for k1 to be about twice as fast as r1. so maybe it would make sense to reflect that in the cost

richardkiss (approved):
Seems good

richardkiss (approved):
I haven't looked too closely since last time. But I thought I'd approve it anyway since my understanding is only the costs have changed.

PR #307, opened 2023-06-08
use NodePtr in serialization

arvidn:
This PR is best reviewed one commit at a time.

The main change is to make `node_to_bytes()` and `node_to_bytes_backrefs()` take an `&Allocator` and `NodePtr` directly, instead of a `&Node`.

The rationale is two-fold:

1. We want to move towards a future where we don't use `Node`. Wrapping an immutable reference to `Allocator` inside the `Node` object is problematic because it prevents simultaneous mutable references to the allocator. This is especially a problem as we move towards an allocator with lazy conversions, where more operations on it will require mutation.
2. It's a slight simplification of these functions. All they do with the `Node` is to unwrap it into the `&Allocator` and `NodePtr` first thing.

There are also two minor unrelated changes:
1. fixing a comment
2. adding some test cases for the `i` operator

### DIFF
--- a/fuzz/fuzz_targets/deserialize_br.rs
+++ b/fuzz/fuzz_targets/deserialize_br.rs
@@ -1,6 +1,5 @@
 #![no_main]
 use clvmr::allocator::Allocator;
-use clvmr::node::Node;
 use clvmr::serde::node_from_bytes_backrefs;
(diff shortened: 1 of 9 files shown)


AmineKhaldi (approved):
Looks good.

PR #309, opened 2023-06-08
require valid G1 and G2 points for negate operator

arvidn:
Even though the implementation is just flipping a bit, and the cost of the operation is very low, requiring a valid point really helps the implementation in the future, where we want to rely on the operator always getting a point in the internal representation (or failing)


op-tests/test-bls-ops.txt
```diff
@@ -67,8 +67,8 @@ g1_negate 0 => FAIL
 g1_negate 0xb7f1d3a73197d7942695638c4fa9ac0fc3688c4f9774b905a14e3a3f171bac586c55e83ff97a1aeffb3af00adb22c6bb 0xb7f1d3a73197d7942695638c4fa9ac0fc3688c4f9774b905a14e3a3f171bac586c55e83ff97a1aeffb3af00adb22c6bb => FAIL
 g1_negate => FAIL
 g1_negate ( 1 2 3 ) => FAIL
-; The input doesn't have to be a valid G1 point
-g1_negate 0xb7f1d3a7319092346345638c4fa9ac0fc3688c4f9774b905a14e3a3f171bac586c55e83ff97a1aeffb3af00adb22c6bb => 0x97f1d3a7319092346345638c4fa9ac0fc3688c4f9774b905a14e3a3f171bac586c55e83ff97a1aeffb3af00adb22c6bb | 1396
+; The input has to be a valid G1 point
```
richardkiss, line 70: How did you find this? Trial & error?

richardkiss (approved):

op-tests/test-bls-ops.txt
arvidn: you mean how I found a 48 byte value that's not a valid G1 point?
I believe I took a valid one and corrupted it some. I would expect *most* 48-byte values are not valid G1 points, at least I can't recall ever producing one by chance.

PR #311, opened 2023-06-10
ask the Allocator if two atoms are equal

arvidn:
This allows the allocator to not (necessarily) convert atoms to bytes, in the future. It's a small step closer to lazy conversion of numbers, G1 and G2 points to and from bytes.

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -232,6 +232,10 @@ impl Allocator {
         Ok(-(self.atom_vec.len() as i32))
     }
 
+    pub fn atom_eq(&self, lhs: NodePtr, rhs: NodePtr) -> bool {
+        self.atom(lhs) == self.atom(rhs)
+    }
+
     pub fn atom(&self, node: NodePtr) -> &[u8] {
         assert!(node < 0, "expected atom, got pair");
         let atom = self.atom_vec[(-node - 1) as usize];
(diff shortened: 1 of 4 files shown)


src/test_ops.rs
```diff
@@ -366,21 +357,6 @@ struct EvalFTracker {
     pub outcome: Option<NodePtr>,
 }
 
-#[cfg(feature = "pre-eval")]
-fn equal_sexp(allocator: &Allocator, s1: NodePtr, s2: NodePtr) -> bool {
```
arvidn, line 370: it turned out this function was duplicated as `node_eq`

src/tests.rs
```diff
@@ -1,15 +1,9 @@
...
 use super::serde::node_to_bytes;
 
 #[cfg(test)]
-fn node_eq(a: &Allocator, lhs: NodePtr, rhs: NodePtr) -> bool {
-    match (a.sexp(lhs), a.sexp(rhs)) {
-        (SExp::Pair(l0, l1), SExp::Pair(r0, r1)) => node_eq(a, l0, r0) && node_eq(a, l1, r1),
-        (SExp::Atom(), SExp::Atom()) => a.atom(lhs) == a.atom(rhs),
-        _ => false,
-    }
-}
```
arvidn, line 12: another copy of it here

src/allocator.rs
```diff
@@ -232,6 +232,10 @@ impl Allocator {
         Ok(-(self.atom_vec.len() as i32))
     }
 
+    pub fn atom_eq(&self, lhs: NodePtr, rhs: NodePtr) -> bool {
+        self.atom(lhs) == self.atom(rhs)
```
arvidn, line 236: this is the new function. The idea is to give the allocator an opportunity (in the future) lazily convert atoms to bytes. This function is another opportunity to keep atoms in their native/unserialized form

PR #312, opened 2023-06-14
introduce `modpow` and `mod` operators

arvidn:
These changes are best reviewed one commit at a time.

This introduces two new operators, under the same `softfork` extension as the BLS operators. Similar to those, these will also become available outside of the `softfork` operator after the hard-fork.

| name | opcode| arguments | description | cost |
| --- | --- | --- | --- | --- |
| modpow | 60 | base exponent modulus | computes: `(base^exponent) % modulus`. exponent may not be negative. modulus may not be 0. | base cost: 17000, cost per byte (base): 38. cost per square byte (exponent): 3. cost per square byte (modulus): 21. |
| % | 61 | numerator denominator | computes the remainder from `numerator` divided by `denominator` | same as `/`, base cost: 998. cost per byte (sum of both operands): 4 |

The fuzzing infrastructure is updated to generate initial corpus for the new operators.

The benchmark program was refactored to more easily support new kinds of operator signatures, and extended to support the new operators.

test cases, generated from python, are added as its own test file under `op-tests`.


op-tests/test-bls-ops.txt
```diff
@@ -22,6 +22,117 @@ coinid 0x1234500000000000000000000000000000000000000000000000000000000000 0x6789
...
+modpow -12345 6789 4444 => 2339 | 11094
+modpow -12345 6789 -4444 => -2105 | 11094
+modpow 12345 6789 -4444 => -2339 | 11094
+
+modpow 12345 6789 44444444444 => 13456191581 | 11292
+
+modpow 0 6789 4444 => 0 | 11044
+modpow 2 64 32 => 0 | 11026
+modpow 2 64 31 => 16 | 11036
+
```
richardkiss, line 65: A typical use case for secp would be three random-looking 256 bit numbers. Maybe add a couple tests that look like that?

richardkiss: These costs seem way too low. Have you benchmarked this? I'd think the cost should scale linearly with the log of the exponent, but also with the size of the modulus, since the implementation is probably square $log_2 X$ times (where X = exponent) and multiply the accumulated result about half the time.

richardkiss: Maybe there's a smarter implementation I'm not aware of.

richardkiss: It uses Montgomery multiplication in the case of an odd modulus. I thought if the modulus were even, it would do some reduction to make the modulus odd, but I don't see any evidence of that. So maybe this is a really smart and really fast implementation, at least for odd modulus. For secp, the modulus is constant, and odd.

op-tests/test-bls-ops.txt
```diff
@@ -22,6 +22,117 @@ coinid 0x1234500000000000000000000000000000000000000000000000000000000000 0x6789
...
+% 100 70 => 30 | 1006
+
+; mod by zero
+% 3 0 => FAIL
+; invalid arguments
+% 3 => FAIL
+% => FAIL
+% 1 2 3 => FAIL
+% (1 2 3 ) 2 => FAIL
+% 2 (1 2 3 ) => FAIL
```
richardkiss, line 134: So these match the `divmod` values?

op-tests/test-bls-ops.txt
richardkiss: https://github.com/rust-num/num-bigint/blob/master/src/biguint/monty.rs

richardkiss: I see you've added additional tests in a later change.

op-tests/test-modpow.txt
```diff
@@ -0,0 +1,82 @@
...
+% 379381184782583886160798882693322550895543756104221513547 4827293201730901888701785199482402666774902234258786888061 => 379381184782583886160798882693322550895543756104221513547 | 1424
+modpow 3195368787213672865566401839626864639167178250334332902956 73504708037419427367735470928041825274387433527777430073 4811637027825861720291821784099893200774466654637793791600 => 3475690977553257351143127975640803368777626976201134848416 | 18353
+% 3195368787213672865566401839626864639167178250334332902956 4811637027825861720291821784099893200774466654637793791600 => 3195368787213672865566401839626864639167178250334332902956 | 1438
+modpow 3393328799375776415256184972208980097810780581787744702798 3052118175933073395388339784015782524774488551946159481690 5436126395750286638399600726623817593321195552838489065236 => 4440460779126825370214984702000853599873253048512412194128 | 18353
+% 3393328799375776415256184972208980097810780581787744702798 5436126395750286638399600726623817593321195552838489065236 => 3393328799375776415256184972208980097810780581787744702798 | 1438
```
richardkiss, line 44: A lot of these tests are of the form `(% A B)` where `A < B`, which isn't very interesting. Maybe add some where `B` is 128 bits instead of 256?

richardkiss (approved):
Looks good. Amazed at how fast `modpow` seems to be. Maybe add some `(% 256-bit 128-bit)` test cases.

op-tests/test-bls-ops.txt
arvidn: yes. I took many of the same test cases from `divmod`

op-tests/test-modpow.txt
arvidn: yeah, I would expect about half of them are like that. I'll add a few more cases

op-tests/test-bls-ops.txt
arvidn: I did benchmark the cost, but I did not perform an exhaustive search, and I made the assumption/simplification that the cost grows exponentially in the size of the exponent and the modulus. Specifically, I just searched one parameter at a time, leaving the others with relatively small numbers. At small numbers the cost is probably dominated by the parsing and allocation of the number.

I'll add some rigour to these measurements and post an update

arvidn: I just noticed I also had failed to add the python program I used to generate the test cases. This is what validates the results against python. I'll add that as adjust the test cases.

arvidn:
I updated the last commit to include the program that generated the test cases (`tools/generate-modpow-tests.py`) while updating it to have test cases where the numerator is larger than the denominator.

I added a new commit where I add the ability for the benchmark program to generate gnuplots of the measurements along with the line fittings.

I added a new commit where I changed the value used for the two other arguments to `modpow` that are not being tested.

I think you're right that the cost was a bit low. I'm adjusting it upwards. I'm updating the original commit, but basically making these changes:

`MODPOW_BASE_COST` from 11000 -> 17000
`MODPOW_COST_PER_BYTE_BASE_VALUE` from 15 -> 38
`MODPOW_COST_PER_BYTE_EXPONENT` unchanged at 3. This is the cost per byte squared.
`MODPOW_COST_PER_BYTE_MOD` from 8 to 21. This is the cost per byte squared.

Here's the output from the updated benchmark program:

```
cost scale: 6.425771579565472
base cost scale: 2.3786823529411767
arg cost scale: 10.418449612403101
opcode: modpow (modulus cost) (60)
   time: per-byte: 3.32ns
   cost: per-byte: 21
   time: base: 7442.54ns
   cost: base: 17703
opcode: modpow (exponent cost) (60)
   time: per-byte: 0.36ns
   cost: per-byte: 2
   time: base: 7427.67ns
   cost: base: 17668
opcode: modpow (value cost) (60)
   time: per-byte: 5.85ns
   cost: per-byte: 38
   time: base: 7139.68ns
   cost: base: 16983
```

The model fits well for size of base-value and modulus. the size of the exponent doesn't quite grow exponentially, but I think it's a close enough match. We want to avoid spending too much resource on computing the cost too. Note that the y-axis is the square root of the measured value on the plots where the x label ends with "log". This is because the cost model of these is that they grow exponentially, but they are brought down to linear for the linear regression analysis.

modpow (exponent cost)-per-byte
modpow (modulus cost)-per-byte
modpow (value cost)-per-byte

PR #315, opened 2023-06-15
`get_args`, and drop `Node`

arvidn:
This change moves away from `Node` as the abstraction to iterate over lists, replacing it with two lower level facilities:

* `Allocator::next()` - a function that can be used in `while` conditions to loop over a list.
* `get_args()` and `get_varargs()` - functions that match and extracts arguments from argument lists. This replaces the need for checking the number of arguments as well as picking out the nodes by iterating or calling `first()?.node`.

Some utility functions (in `op_utils.rs`) were changed to take an `&Allocator` + `NodePtr` rather than `&Node` as well, to allow dropping the `Node` class.

The `get_args()` abstraction is a quite nice improvement to the ergonomics of the operator implementations, and in my mind is the main motiviation for this patch.

I added a few missing unit tests for the updated utility functions as well as `Allocator::atom_len()`.

The original intention of removing the `Node` abstraction was to allow for `Allocator` references to be mutable, and have `Allocator::atom()` (and related functions) take a mutable reference to the allocator. However, some experimentation beyond this point suggests that's actually impractical. But I think this step is still an improvement.


src/op_utils.rs
```diff
@@ -13,87 +12,269 @@ use num_integer::Integer;
...
+    }
+
+    if counter != N {
         err(
-            args.node,
+            args,
             &format!(
-                "{name} takes exactly {expected} argument{}",
-                if expected == 1 { "" } else { "s" }
+                "{name} takes exactly {N} argument{}",
```
richardkiss, line 43: No need to change, but the grammar trick I often use is "argument count should be {N}" to avoid plural switching.

richardkiss: (In fact, I guess you can't easily change if tests expect this exact message.)

arvidn: yeah, I tried to keep error messages compatible. Although, I don't think there are that many tests to update

src/allocator.rs
```diff
@@ -945,3 +957,59 @@ fn test_roundtrip(#[case] test_value: &str, #[case] make: MakeFun, #[case] check
...
+    let buffer = &hex::decode(buffer_hex).unwrap();
+    let g1 =
+        G1Projective::from(G1Affine::from_compressed(&buffer[..].try_into().unwrap()).unwrap());
+    let atom = a.new_g1(g1).unwrap();
+    assert_eq!(a.atom_len(atom), expected);
+}
+
+#[cfg(test)]
+#[rstest]
+#[case("93e02b6052719f607dacd3a088274f65596bd0d09920b61ab5da61bbdc7f5049334cf11213945d57e5ac7d055d042b7e024aa2b2f08f0a91260805272dc51051c6e47ad4fa403b02b4510b647ae3d1770bac0326a805bbefd48056c8c121bdb8", 96)]
```
richardkiss, line 1005: This might work (and `cargo fmt` will clean it up better):

```suggestion
#[case(
"93e02b6052719f607dacd3a088274f65596bd0d09920b61ab5da61bbdc7f5049"
"334cf11213945d57e5ac7d055d042b7e024aa2b2f08f0a91260805272dc51051"
"c6e47ad4fa403b02b4510b647ae3d1770bac0326a805bbefd48056c8c121bdb8",
96)]
```

Works in python. Don't know offhand in rust though.

src/run_program.rs
```diff
@@ -763,7 +743,7 @@ const TEST_CASES: &[RunProgramTest] = &[
         flags: 0,
         result: None,
         cost: 0,
-        err: "in ((X)...) syntax X must be lone atom",
+        err: "((X)...) syntax takes exactly 1 argument",
```
richardkiss, line 746: This message isn't quite right. The idea it's trying to get across is that `X` must be an atom and not a cons. It can take multiple arguments (which are implicitly quoted).

Example:

`((+) 1 2 3) = 6`

We should make sure there is at least one test of this.

richardkiss: The intended use for this is for a clvm implementation of a clvm emulator. It would descend trees and evaluated subtrees. It evaluated arguments first, and when it sees an operator it wants to pass through, it would package it up into `(op)` then cons it to the operator list yielding something like `((+) 1 2 3)`. Then you "apply" with `a` and you get `6`.

richardkiss: Without this trick, you'd have to iteratively quote the arguments to build up `(+ (q . 1) (q . 2) (q . 3))` which scales per operators, and would slow things down quite a lot.

richardkiss (approved):
Looks good. The `get_args` and `get_varargs` seem really handy, and the implementation uses templating over constants, which is a relatively new thing. I suppose we could have done this with `get_args_1`, `get_args_2`, etc.

It's a little scary changing code for every single operator, but as long as they're all reasonably well tested, I think it's probably okay.

I know the `((+) 1 2 3)` thing is a bit mysterious (and probably unused in the blockchain), so let me know if you need more discussion of that.

arvidn:
I'll address these comments in a separate PR. I hope that's OK.

> This message isn't quite right. The idea it's trying to get across is that X must be an atom and not a cons. It can take multiple arguments (which are implicitly quoted).

Yeah, I agree that I was a bit lazy in making that use of `get_args()` produce a good error message. I'd prefer to keep `get_args()` relatively simple, and not make the error message it produces configurable (because it's good in almost all cases). I could make it say: `in the ((X)...) syntax, the inner list takes exactly 1 argument`

> This might work (and cargo fmt will clean it up better):

string literals don't automatically concatenate in rust, but you can insert line breaks using `\`.

> ((+) 1 2 3) = 6
> We should make sure there is at least one test of this.

There is at least one positive test for this construct, I'll add this example as well.

arvidn:
https://github.com/Chia-Network/clvm_rs/pull/316

PR #316, opened 2023-06-17
address comments from get-args PR

arvidn:


src/allocator.rs
```diff
@@ -702,8 +702,18 @@ fn test_point_atom_pair(#[case] fun: TestFun, #[case] expected: &str) {
...
+97f1d3a73197d7942695638c4fa9ac0f\
+c3688c4f9774b905a14e3a3f171bac58\
+6c55e83ff97a1aeffb3af00adb22c6bb"
+)]
+#[case(
+    "\
+a572cbea904d67468808c8eb50a9450c\
+9721db309128012543902d0ac358a62a\
+e28f75bb8f1c7c42c39a8c5529bf0f4e"
+)]
```
arvidn, line 716: using the `hex_literal` `hex!` macro would be ideal here, but that's a bigger change

src/bls_ops.rs
```diff
@@ -193,7 +193,7 @@ pub fn op_bls_g2_negate(a: &mut Allocator, input: NodePtr, _max_cost: Cost) -> R
         .is_none()
         .into()
     {
-        return err(args.first()?.node, "atom is not a valid G2 point");
+        return err(point, "atom is not a valid G2 point");
```
arvidn, line 196: this was an incompatibility between the two most recent PRs I merged.

richardkiss (approved):

PR #319, opened 2023-06-19
Change from using podman and using the rustup script to a docker imag…

pmaslana:
…e and source


wjblanke (approved):
aok

prozacchiwawa (commented):
Looks ok, based on inspection and running a modified version locally.

.github/workflows/build-test.yml
```diff
@@ -194,13 +192,15 @@ jobs:
...
         run: |
+          rustup update
           cd fuzz
           cargo clippy
       - name: cargo-fuzz
         run: cargo +nightly install cargo-fuzz
       # TODO: it would be nice to save and restore the corpus between runs
       - name: build corpus
         run: |
+          cargo update -p proc-macro2  # Temporary fix to get the nightly build to run correctly
```
arvidn, line 203: we fixed this by updating the `Cargo.lock` file in the repository. I think I would prefer it that way

pmaslana: The cargo.lock file has been updated by running cargo update for proc-macro2

PR #325, opened 2023-07-06
Move wheel and fuzz into main workspace and fix CI errors

Rigidity:
This PR makes the following changes:
- Adds `wheel` and `fuzz` to the workspace.
- Updates the `.gitignore` file to include `target` so the generated folder isn't considered by source control in editors.
- Fixes CI permission error relating to Twine.
- Fixes CI linkage error by excluding fuzz from the workspace build (still enabled in tests).


arvidn:
what do we gain by doing this?
I think there's a downside in the workflow files needing to explicitly run `clippy` and `fmt` in more workspaces (I would think we would want to move in the direction of putting everything in a single workspace).

Rigidity:
As you said, `wheel` can't be part of the workspace and both `wheel` and `fuzz` need to have a Cargo.lock file pinning their dependencies. This achieves all of that without needing a Cargo.lock for the library.

Rigidity:
Looks like Cargo will ignore the lock file for libraries anyway. So if you feel strongly that they should be combined into a single workspace I can do so

Rigidity:
The main reason for this PR is the gitignore. It's unusably annoying to have my editor create target folders (with hundreds of generated files) every time I use Rust Analyzer or compile the code.

arvidn:
> As you said, wheel can't be part of the workspace and both wheel and fuzz need to have a Cargo.lock file pinning their dependencies. This achieves all of that without needing a Cargo.lock for the library.

If we can find a way for the fuzzer and pyo3 to be able to coexist in a workspace, and use the *same* lockfile, I think that would be ideal. The versions we build the releases from really *should* be the same version we fuzz. Right now we can't guarantee that because we build the production releases from the `chia_rs` repo.

I don't think addressing your IDE issues should take precedence over that. Worst case you can reconfigure your IDE.

> It's unusably annoying to have my editor create target folders (with hundreds of generated files) every time I use Rust Analyzer or compile the code

Does it do this because the lockfile is not in `.gitignore`?

Rigidity:
The IDE issues were from `target` not being gitignored. This absolutely should be in the gitignore file and isn't really controversial.

wheel/Cargo.lock
arvidn, line 1: I would have expected this to no longer be used, and removed from the repository. Is there a reason to leave it in?

wheel/Cargo.toml
```diff
@@ -9,6 +9,9 @@ homepage = "https://github.com/Chia-Network/clvm_rs/"
 repository = "https://github.com/Chia-Network/clvm_rs/"
 readme = "README.md"
 
+[workspace]
+members = ["."]
```
arvidn: my understanding is that this will prevent `wheel` to be included in the main workspace.
The name of this commit moves the wheel project into its *own* workspace, which contradicts the PR title.

fuzz/.gitignore
```diff
@@ -1,3 +1,3 @@
 target
 corpus
-artifacts
+artifacts
```
arvidn, line 2: did you remove the newline at the end of this file? it seems a bit odd

fuzz/Cargo.toml
```diff
@@ -5,6 +5,9 @@ authors = ["Arvid Norberg <arvid@chia.net>"]
 publish = false
 edition = "2021"
 
+[workspace]
+members = ["."]
+
```
arvidn: this change seems redundant, and it's also not making the change suggested by the PR title

.gitignore
```diff
@@ -1,2 +1,2 @@
-.idea/
-wasm/target
+/target
+/Cargo.lock
```
arvidn: I think we should follow the convention of having trailing newlines in text files

Cargo.lock
arvidn, line 1: this is also moving away from the title of this PR. this is not moving towards all tests and fuzzing using the same dependencies.
I don't believe you justify this change in the PR description.

wheel/.gitignore
```diff
@@ -1 +0,0 @@
-target
```
arvidn: you just added this file in a previous commit. I think it would be better to not add it at all if you're just removing it again

.gitignore
```diff
@@ -1,2 +1,2 @@
-/target
+target
```
arvidn: you've changed this file multiple times now. I think you should just keep the most recent version of the file

fuzz/Cargo.toml
```diff
@@ -5,9 +5,6 @@ authors = ["Arvid Norberg <arvid@chia.net>"]
 publish = false
 edition = "2021"
 
-[workspace]
-members = ["."]
-
```
arvidn: this is undoing an earlier commit. I think it would be best to just squash them

.github/workflows/benchmark.yml
arvidn, line 1: I don't think this is a clear cut improvement. You add noise to the git history (`git blame` specifically) with virtually no benefit. If the formatting is important, we should enforce it with `pre-commit` (or something like that), and on CI.

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+  "useTabs": false,
+  "tabWidth": 2,
+  "singleQuote": true
+}
```
arvidn, line 5: This commits to a specific tool to maintain this formatting as well. Is this the best tool?

.github/workflows/build-test.yml
```diff
@@ -237,7 +237,7 @@ jobs:
           components: rustfmt, clippy
 
       - name: build
-        run: cargo build --workspace
+        run: cargo build --workspace --exclude clvm_rs-fuzz
```
arvidn, line 240: I think this warrants a comment

.github/workflows/build-arm64-wheels.yml
```diff
@@ -59,7 +59,7 @@ jobs:
 
       - name: Install Twine
         run: |
-          if [ ! -f "venv" ]; then rm -rf venv; fi
+          if [ ! -f "venv" ]; then sudo rm -rf venv; fi
```
arvidn, line 62: why is this necessary now? Is this an unrelated regression in the build environment?

arvidn (commented):
I don't think the formatting update is an obvious improvement, but I'm not opposed (I just think it's poorly justified).

I think everything else looks good but should be squashed

Rigidity:
Replaced by cleaner #328

richardkiss:
> The IDE issues were from `target` not being gitignored. This absolutely should be in the gitignore file and isn't really controversial.

@Rigidity I've kind of hated `target` directories being dropped everywhere and being backed up by Time Machine on my Mac, so I added the following to my `~/.cargo/config`:

```
[build]
target-dir = "/Users/kiss/tmp/cargo-target/"
```

Now all builds go into `tmp`, which I've excluded from backup.

Also, if you add

```
[net]
git-fetch-with-cli = true
```

your updates of the cargo database will be a lot faster.

PR #337, opened 2023-08-17
Switch to trusted publishing for pypi

Starttoaster:
With some additional housekeeping coming from review comments:
- `arch -arm64` in m1 wheel job commands no longer required, this job originally ran on intel procs which did require this command prefix.
- Use more generic macos/arm64 runner tags in preparation for future M2 runners

### DIFF
--- a/.github/workflows/build-arm64-wheels.yml
+++ b/.github/workflows/build-arm64-wheels.yml
@@ -11,6 +11,10 @@ on:
     branches:
       - '**'
 
+permissions:
+  contents: read
+  id-token: write
+
 jobs:
   build_wheels:
     name: ARM64 Python Wheels on ARM64 Ubuntu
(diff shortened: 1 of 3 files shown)


.github/workflows/build-m1-wheel.yml
```diff
@@ -99,27 +103,14 @@ jobs:
...
-        if [ -n "$SECRET" ]; then HAS_SECRET='true' ; fi
-        echo "HAS_SECRET=${HAS_SECRET}" >>$GITHUB_OUTPUT
-      env:
-        SECRET: "${{ secrets.test_pypi_password }}"
-
     - name: Install twine
       run: |
         . ./venv/bin/activate
-        arch -arm64 pip install twine
+        pip install twine
```
cmmarslender, line 109: Don't need this step with the new action

.github/workflows/build-arm64-wheels.yml
```diff
@@ -67,25 +71,12 @@ jobs:
         pip install setuptools_rust
         pip install twine
```
cmmarslender, line 72: Dont need to isntall twine. The new action will run in a container with twine in it

.github/workflows/build-test.yml
```diff
@@ -132,23 +136,12 @@ jobs:
     - name: Install Twine
       run: pip install twine
```
cmmarslender, line 137: Dont need to install twine w/ new action

cmmarslender (changes requested):

cmmarslender (approved):

PR #338, opened 2023-08-22
simplify raw-loops into while-loops

arvidn:
in some cases this saves one or two levels of indentation of the loop body.
It always saves a few lines of code. It makes it more idiomatic rust I believe.


Rigidity (approved):
lgtm

PR #339, opened 2023-08-24
Move is_bit_set into match body

ardikars:
Move is_bit_set into match body.

### DIFF
--- a/src/run_program.rs
+++ b/src/run_program.rs
@@ -475,8 +475,7 @@ impl<'a, D: Dialect> RunProgramContext<'a, D> {
             if cost > effective_max_cost {
                 return err(max_cost_ptr, "cost exceeded");
             }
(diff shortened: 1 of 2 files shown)


src/run_program.rs
```diff
@@ -475,8 +475,7 @@ impl<'a, D: Dialect> RunProgramContext<'a, D> {
             if cost > effective_max_cost {
                 return err(max_cost_ptr, "cost exceeded");
             }
-            let top = self.op_stack.pop();
-            let op = match top {
+            let op = match self.op_stack.pop() {
```
arvidn, line 478: I suspect this whole `loop` should probably be turned into something like:
```
while let Some(op) = self.op_stack.pop() {
   ...
}
```

as long as we can ensure the cost check is still performed in the edge case of it exceeding on the last operator

PR #341, opened 2023-09-26
Bump num-traits and num-bigint

Rigidity:


arvidn:
are there new features we need from this newer version?

Rigidity:
It would just be nice if we could unpin the versions, I get version conflicts every time I install them alongside clvmr

PR #345, opened 2023-10-31
port BLS operators to use chia-bls (and BLST)

arvidn:


src/bls_ops.rs
```diff
@@ -61,7 +61,7 @@ pub fn op_bls_g1_subtract(a: &mut Allocator, mut input: NodePtr, max_cost: Cost)
         if is_first {
             total = point;
         } else {
-            total -= point;
+            total -= &point;
```
richardkiss, line 64: This seems odd. Why would it need to be `&` here but not in line 62?

src/bls_ops.rs
```diff
@@ -150,7 +147,7 @@ pub fn op_bls_g2_subtract(a: &mut Allocator, mut input: NodePtr, max_cost: Cost)
...
```
richardkiss, line 150: (ditto above comment)

richardkiss (approved):

PR #359, opened 2024-01-07
`NodePtr` internal representation

arvidn:
This patch changes the internal representation of `NodePtr` to use the top 6 bits as "type" and the bottom 26 bits for "index". Currently we use positive numbers for atoms and negative numbers for pairs. The new representation supports more types.

For example, we may want to have atoms in `BigInt`, `G1` or `G2` formats internally. These could be new types.

But perhaps the most important optimization this opens up is to have atoms point to external buffers. e.g. when running a program stored in a buffer, we could have atoms point directly into that buffer rather than copying everything into the `Allocator` object first.

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -23,6 +23,11 @@ impl NodePtr {
         Self::new(ObjectType::Bytes, 0)
     }
 
+    // TODO: remove this
+    pub fn hack(val: usize) -> Self {
+        Self::new(ObjectType::Bytes, val)
+    }
+
     fn new(t: ObjectType, idx: usize) -> Self {
         assert!(idx <= NODE_PTR_IDX_MASK as usize);
         NodePtr(((t as u32) << NODE_PTR_IDX_BITS) | (idx as u32))
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -4,8 +4,56 @@ use crate::reduction::EvalErr;
 use chia_bls::{G1Element, G2Element};
 use clvm_traits::{ClvmDecoder, ClvmEncoder, FromClvmError, ToClvmError};
 
+const MAX_NUM_ATOMS: usize = 62500000;
+const MAX_NUM_PAIRS: usize = 62500000;
+const NODE_PTR_IDX_BITS: u32 = 26;
+const NODE_PTR_IDX_MASK: u32 = (1 << NODE_PTR_IDX_BITS) - 1;
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct NodePtr(pub i32);
+pub struct NodePtr(u32);
+
+enum ObjectType {
+    Pair,
+    Bytes,
+}
+
+// The top 6 bits of the NodePtr indicate what type of object it is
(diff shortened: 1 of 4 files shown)


src/allocator.rs
```diff
@@ -4,8 +4,52 @@ use crate::reduction::EvalErr;
...
 use clvm_traits::{ClvmDecoder, ClvmEncoder, FromClvmError, ToClvmError};
 
+const MAX_NUM_ATOMS: usize = 62500000;
+const MAX_NUM_PAIRS: usize = 62500000;
+const NODE_PTR_IDX_BITS: u32 = 26;
+const NODE_PTR_IDX_MASK: u32 = (1 << NODE_PTR_IDX_BITS) - 1;
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct NodePtr(pub i32);
+pub struct NodePtr(pub u32);
```
Rigidity, line 13: It would be nice if we can make this private, so that the inner type representation can be changed without breaking semver compatibility. Is there anything currently that relies on this value?

src/allocator.rs
```diff
@@ -4,8 +4,52 @@ use crate::reduction::EvalErr;
...
+                1 => ObjectType::Bytes,
+                _ => {
+                    panic!("unknown NodePtr type");
+                }
+            },
+            (self.0 & NODE_PTR_IDX_MASK) as usize,
+        )
+    }
+
+    pub fn to_index(&self) -> usize {
```
Rigidity, line 40: Does this need to be fully public? Perhaps `pub(crate)` would suffice here?

src/allocator.rs
arvidn: good point. my intention of implementing `Default` was for nothing to depend on this. but it seems there still are a few places. I'll investigate

arvidn: sadly, there's one ugly hack left. for now I just called it out as a hack to be fixed in the future

src/allocator.rs
arvidn: I also renamed this to `as_index()` to please clippy. However, I believe this function won't live for very long

Rigidity (approved):

src/allocator.rs
```diff
@@ -4,8 +4,56 @@ use crate::reduction::EvalErr;
...
+pub struct NodePtr(u32);
+
+enum ObjectType {
+    Pair,
+    Bytes,
+}
+
+// The top 6 bits of the NodePtr indicate what type of object it is
+impl NodePtr {
+    pub fn null() -> Self {
```
richardkiss, line 22: I asked ChatGPT and it suggests that in lisp `nil` is used more commonly than `null`.

richardkiss (approved):
Looks reasonable. Maybe consider `s/null/nil/g`.

src/allocator.rs
arvidn: yeah, I think that has been established. we still use `null()` though, and I think changing that (everywhere) ought to be a separate PR

arvidn: https://github.com/Chia-Network/clvm_rs/pull/360

PR #366, opened 2024-01-25
Stop using deprecated rust action

emlowe:


richardkiss (approved):
Seems fine I suppose.

.github/workflows/build-test.yml
```diff
@@ -194,10 +190,8 @@ jobs:
         with:
           fetch-depth: 1
       - name: Install rust
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@nightly
```
richardkiss, line 193: Do we want `stable` downloaded too? Not sure if it does that.

emlowe: I was assuming the previous action
```
uses: actions-rs/toolchain@v1
        with:
            toolchain: nightly
            override: true
            components: rustfmt, clippy
```
already only did the nightly version but perhaps it didn't work that way

richardkiss: You can definitely have both toolchains installed simultaneously. Might be worth understanding why nightly is requested for this step. That said, it probably is fine if there are no errors (and nightly builds are allowed to break stuff with no notice anyway).

Rigidity (approved):
lgtm

.github/workflows/build-test.yml
arvidn: I believe some feature used to require the nightly build, but I would not be surprised those features have made it into stable by now

arvidn (approved):

PR #371, opened 2024-02-02
minor updates to traverse_path tests

arvidn:


src/traverse_path.rs
```diff
@@ -109,11 +109,17 @@ fn test_traverse_path() {
     let n1 = a.new_atom(&[0, 1, 2]).unwrap();
     let n2 = a.new_atom(&[4, 5, 6]).unwrap();
 
-    assert_eq!(traverse_path(&a, &[0], n1).unwrap(), Reduction(48, nul));
```
arvidn, line 112: this is a redundant 0, that's why the cost is 48, instead of 44. This was moved down to the "cost for leading zeros" section

Rigidity (approved):

PR #372, opened 2024-02-02
`SmallAtom` optimization

arvidn:
These commits are best reviewed one at a time.

# purpose

This patch introduces a more compact representation of small atoms. A `NodePtr` is 32 bits, 6 bits identifies the kind of node it points to (Atom or Pair), 26 bits is an index into the vector of atoms or pairs.

This patch adds a 3rd kind of node, `SmallAtom`, where the value of the atom is stored in the low 26 bits directly. The benefits are:

1. It saves memory, because we don't need to allocate the atom on the heap, nor the atom itself pointing into the heap
2. indirections, since we have the value immediately, saving at least two memory lookups that are likely to be cache misses (at least on small computers with a small data cache)

A very large number of atoms are small. In a typical CLVM program for instance, all op-codes and environment paths are small.

## format

`SmallAtom` only support atoms that are:
1. *integers* in the *canonical* representation
2. positive integers
3. fit in 26 bits

The 26 bits are interpreted as an *unsigned* integer.

## leading zeros

Since our canonical integer representation requires leading zero bytes for certain positive integers (where the most significant bit in the following byte is set), the leading zero is implied by `SmallAtom` nodes. The function `len_for_value()` in `allocator.rs` implements this behavior. It returns the length an atom *would have* had, if it has been allocated on the heap.

## Allocator::nil(), Allocator::one()

The Allocator always allocate two nodes on construction, `nil` and `one`. With this patch, we no longer need to allocate these in the atom list nor on the heap. We can just construct `NodePtr` representing those values directly, as `SmallAtom`. In order to still enforce the same limit on the number of atoms that are allowed to be allocated by a program, we subtract 2 from the constant, `MAX_NUM_ATOMS`.

## MAX_NUM_ATOMS

We have a (consensus critical) rule on how many atoms are allowed to be allocated by a program. Since `SmallAtom`s aren't allocated in the `atom_vec` list, we need to keep a count of them separately, to ensure the limit is identical as prior to this optimization. We keep a counter `small_atoms` counting the number of small atoms we've allocated. Whenever we check the `MAX_NUM_ATOMS` limit, we include this counter.

## atom() -> &[u8]

A major challenge with this optimization is in the Allocator's interface. It lets you ask for a pointer to the buffer of any atom. Small atoms don't have a buffer since they're not allocated on the heap. In order to stick to the current interface of the allocator, this function is preserved, using a pool of small buffers to return a pointer to.

We have a ring buffer, `temp_vec`, with a cursor to the next free slow, `temp_idx`. When the `atom()` is requested for a node that is a `SmallAtom`, we grab the next available slot in the ring buffer, increment the cursor, print the integer into the buffer and returns a reference to it.

This relies on the return values from `atom()` being short lived, and that we never, at one point, keep more than 64 of these references around.

## visit_node()

In order to optimize functions using integers, there's a new function on `Allocator` called `visit_node()`, which calls a callback with the value of the node pass into it. This means we can avoid conversions from small integers to `BigInt` for example.

# benchmarks

The results of the benchmarks in this repository follow. This is the output from `critcmp`:

## RPi5

```
group                                               after                                  before
-----                                               -----                                  ------
deserialize/node_from_bytes                         1.01    539.1±0.62µs        ? ?/sec    1.00    531.6±0.31µs        ? ?/sec
deserialize/node_from_bytes_backrefs                1.03    573.1±0.56µs        ? ?/sec    1.00    555.4±2.25µs        ? ?/sec
deserialize/serialized_length_from_bytes            1.00    352.8±0.17µs        ? ?/sec    1.00    353.6±0.14µs        ? ?/sec
deserialize/serialized_length_from_bytes_trusted    1.02    122.0±0.81µs        ? ?/sec    1.00    119.8±1.43µs        ? ?/sec
run_program/block-2000                              1.00    358.6±0.89ns        ? ?/sec    1.02    364.6±0.21ns        ? ?/sec
run_program/compressed-2000                         1.00   1489.4±1.28ms        ? ?/sec    1.15   1705.7±1.69ms        ? ?/sec
run_program/concat                                  1.01   1027.2±1.84ms        ? ?/sec    1.00  1016.8±11.31ms        ? ?/sec
run_program/count-even                              1.00     22.7±0.03ms        ? ?/sec    1.46     33.2±0.08ms        ? ?/sec
run_program/factorial                               1.00    470.3±2.79ms        ? ?/sec    1.02    478.3±3.81ms        ? ?/sec
run_program/hash-string                             1.00   1127.9±1.38ns        ? ?/sec    1.02   1153.7±0.54ns        ? ?/sec
run_program/hash-tree                               1.00     83.0±0.08ms        ? ?/sec    1.11     92.3±0.11ms        ? ?/sec
run_program/large-block                             1.00    444.1±0.79ms        ? ?/sec    1.10    488.6±0.31ms        ? ?/sec
run_program/loop_add                                1.00       5.6±0.01s        ? ?/sec    1.34       7.5±0.02s        ? ?/sec
run_program/loop_ior                                1.00       5.1±0.00s        ? ?/sec    1.50       7.6±0.03s        ? ?/sec
run_program/loop_not                                1.00       5.3±0.01s        ? ?/sec    1.28       6.8±0.04s        ? ?/sec
run_program/loop_sub                                1.00       5.6±0.01s        ? ?/sec    1.39       7.8±0.01s        ? ?/sec
run_program/loop_xor                                1.00       5.7±0.00s        ? ?/sec    1.30       7.4±0.01s        ? ?/sec
run_program/matrix-multiply                         1.00    382.4±0.24ms        ? ?/sec    1.13    431.3±0.37ms        ? ?/sec
run_program/point-pow                               1.00   1296.1±0.24ms        ? ?/sec    1.00   1297.3±0.33ms        ? ?/sec
run_program/pubkey-tree                             1.00    682.7±0.07ms        ? ?/sec    1.00    683.2±0.10ms        ? ?/sec
run_program/shift-left                              1.00       4.4±0.01s        ? ?/sec    1.01       4.5±0.01s        ? ?/sec
run_program/substr                                  1.00     11.8±0.02ms        ? ?/sec    1.40     16.5±0.02ms        ? ?/sec
run_program/substr-tree                             1.00    254.1±0.19ms        ? ?/sec    1.40    354.8±0.29ms        ? ?/sec
run_program/sum-tree                                1.00    604.8±1.22ms        ? ?/sec    1.11    671.6±0.72ms        ? ?/sec
```

Perhaps the more interesting benchmarks are the `run_generator` stress tests in the `chia_rs` repository.
I ran it on Mac OS (M1) and a RaspberryPi 5 machine.

My main concern was to make the *worst case* more palatable. Among our stress tests, that's `deep-recursion-plus` and `duplicate-coin-announce`.

## RPi5

```
group                                                                after                                  before
-----                                                                -----                                  ------
generator/run_block_generator block-1ee588dc                         1.00    281.4±0.29ms        ? ?/sec    1.11    311.2±0.26ms        ? ?/sec
generator/run_block_generator block-1ee588dc-compressed              1.00    278.8±0.31ms        ? ?/sec    1.10    307.1±0.52ms        ? ?/sec
generator/run_block_generator block-225758                           1.00      7.8±0.02ms        ? ?/sec    1.84     14.4±0.03ms        ? ?/sec
generator/run_block_generator block-225758-compressed                1.00      7.8±0.02ms        ? ?/sec    1.84     14.4±0.02ms        ? ?/sec
generator/run_block_generator block-4671894                          1.00    442.0±0.91ms        ? ?/sec    1.85    818.1±1.37ms        ? ?/sec
generator/run_block_generator block-4671894-compressed               1.00    443.5±1.34ms        ? ?/sec    1.84    816.4±0.63ms        ? ?/sec
generator/run_block_generator block-6fe59b24                         1.00    310.9±0.25ms        ? ?/sec    1.10    342.7±0.27ms        ? ?/sec
generator/run_block_generator block-6fe59b24-compressed              1.00    304.9±0.58ms        ? ?/sec    1.11    338.3±0.29ms        ? ?/sec
generator/run_block_generator block-834752                           1.00     22.5±0.03ms        ? ?/sec    1.11     25.0±0.04ms        ? ?/sec
generator/run_block_generator block-834752-compressed                1.00     22.2±0.02ms        ? ?/sec    1.11     24.8±0.05ms        ? ?/sec
generator/run_block_generator block-b45268ac                         1.00    274.2±0.22ms        ? ?/sec    1.11    305.7±0.11ms        ? ?/sec
generator/run_block_generator block-b45268ac-compressed              1.00    272.9±0.23ms        ? ?/sec    1.11    301.6±0.20ms        ? ?/sec
generator/run_block_generator block-c2a8df0d                         1.00    326.4±0.28ms        ? ?/sec    1.11    361.3±0.17ms        ? ?/sec
generator/run_block_generator block-c2a8df0d-compressed              1.00    323.5±0.21ms        ? ?/sec    1.10    356.3±0.11ms        ? ?/sec
generator/run_block_generator block-e5002df2                         1.00    292.8±0.25ms        ? ?/sec    1.11    324.0±0.16ms        ? ?/sec
generator/run_block_generator block-e5002df2-compressed              1.00    289.9±0.20ms        ? ?/sec    1.10    320.0±0.20ms        ? ?/sec
generator/run_block_generator deep-recursion-plus                    1.00       4.4±0.00s        ? ?/sec    1.65       7.3±0.01s        ? ?/sec
generator/run_block_generator deep-recursion-plus-compressed         1.00       4.4±0.01s        ? ?/sec    1.64       7.3±0.01s        ? ?/sec
generator/run_block_generator duplicate-coin-announce                1.00       4.7±0.00s        ? ?/sec    1.38       6.5±0.01s        ? ?/sec
generator/run_block_generator duplicate-coin-announce-compressed     1.00       4.7±0.00s        ? ?/sec    1.38       6.5±0.01s        ? ?/sec
generator/run_block_generator recursion-pairs                        1.00       3.6±0.01s        ? ?/sec    1.18       4.3±0.00s        ? ?/sec
generator/run_block_generator recursion-pairs-compressed             1.00       3.6±0.00s        ? ?/sec    1.18       4.2±0.00s        ? ?/sec
generator/run_block_generator2 block-1ee588dc                        1.00    130.0±0.05ms        ? ?/sec    1.03    133.4±0.02ms        ? ?/sec
generator/run_block_generator2 block-1ee588dc-compressed             1.00    126.9±0.06ms        ? ?/sec    1.02    130.0±0.02ms        ? ?/sec
generator/run_block_generator2 block-225758                          1.00      7.8±0.01ms        ? ?/sec    1.44     11.2±0.01ms        ? ?/sec
generator/run_block_generator2 block-225758-compressed               1.00      7.8±0.01ms        ? ?/sec    1.44     11.2±0.02ms        ? ?/sec
generator/run_block_generator2 block-4671894                         1.00    441.8±0.75ms        ? ?/sec    1.47    650.7±1.24ms        ? ?/sec
generator/run_block_generator2 block-4671894-compressed              1.00    441.2±0.56ms        ? ?/sec    1.47    650.2±0.76ms        ? ?/sec
generator/run_block_generator2 block-6fe59b24                        1.00    145.0±0.07ms        ? ?/sec    1.03    149.0±0.05ms        ? ?/sec
generator/run_block_generator2 block-6fe59b24-compressed             1.00    141.6±0.08ms        ? ?/sec    1.03    145.3±0.05ms        ? ?/sec
generator/run_block_generator2 block-834752                          1.00     12.3±0.00ms        ? ?/sec    1.05     12.9±0.01ms        ? ?/sec
generator/run_block_generator2 block-834752-compressed               1.00     12.2±0.00ms        ? ?/sec    1.05     12.7±0.00ms        ? ?/sec
generator/run_block_generator2 block-b45268ac                        1.00    128.7±0.04ms        ? ?/sec    1.03    132.0±0.08ms        ? ?/sec
generator/run_block_generator2 block-b45268ac-compressed             1.00    125.6±0.11ms        ? ?/sec    1.02    128.6±0.10ms        ? ?/sec
generator/run_block_generator2 block-c2a8df0d                        1.00    149.1±0.05ms        ? ?/sec    1.02    152.7±0.08ms        ? ?/sec
generator/run_block_generator2 block-c2a8df0d-compressed             1.00    145.2±0.06ms        ? ?/sec    1.02    148.4±0.10ms        ? ?/sec
generator/run_block_generator2 block-e5002df2                        1.00    135.9±0.06ms        ? ?/sec    1.03    139.9±0.10ms        ? ?/sec
generator/run_block_generator2 block-e5002df2-compressed             1.00    132.8±0.04ms        ? ?/sec    1.03    136.2±0.16ms        ? ?/sec
generator/run_block_generator2 deep-recursion-plus                   1.00       4.5±0.01s        ? ?/sec    1.65       7.3±0.00s        ? ?/sec
generator/run_block_generator2 deep-recursion-plus-compressed        1.00       4.4±0.01s        ? ?/sec    1.65       7.3±0.01s        ? ?/sec
generator/run_block_generator2 duplicate-coin-announce               1.00       4.7±0.01s        ? ?/sec    1.39       6.5±0.01s        ? ?/sec
generator/run_block_generator2 duplicate-coin-announce-compressed    1.00       4.7±0.00s        ? ?/sec    1.39       6.6±0.01s        ? ?/sec
generator/run_block_generator2 recursion-pairs                       1.00       3.6±0.01s        ? ?/sec    1.17       4.2±0.00s        ? ?/sec
generator/run_block_generator2 recursion-pairs-compressed            1.00       3.6±0.01s        ? ?/sec    1.17       4.3±0.00s        ? ?/sec
tree-hash/tree-hash block-225758                                     1.00    431.4±0.11µs        ? ?/sec    1.00    430.3±0.05µs        ? ?/sec
tree-hash/tree-hash block-225758-compressed                          1.00    431.3±0.13µs        ? ?/sec    1.00    430.3±0.20µs        ? ?/sec
tree-hash/tree-hash block-4671894                                    1.00     18.5±0.00ms        ? ?/sec    1.00     18.4±0.04ms        ? ?/sec
tree-hash/tree-hash block-4671894-compressed                         1.00     18.5±0.01ms        ? ?/sec    1.00     18.4±0.00ms        ? ?/sec
tree-hash/tree-hash block-834752                                     1.00      7.0±0.00ms        ? ?/sec    1.00      7.0±0.00ms        ? ?/sec
tree-hash/tree-hash block-834752-compressed                          1.00      7.0±0.00ms        ? ?/sec    1.00      7.0±0.00ms        ? ?/sec
```

## Mac OS (M1)

```
group                                                                after                                  before
-----                                                                -----                                  ------
generator/run_block_generator block-1ee588dc                         1.00     85.9±0.55ms        ? ?/sec    1.06     90.8±0.18ms        ? ?/sec
generator/run_block_generator block-1ee588dc-compressed              1.00     84.3±0.85ms        ? ?/sec    1.08     91.0±0.23ms        ? ?/sec
generator/run_block_generator block-225758                           1.00      2.1±0.02ms        ? ?/sec    1.88      4.0±0.06ms        ? ?/sec
generator/run_block_generator block-225758-compressed                1.00      2.1±0.01ms        ? ?/sec    1.93      4.1±0.02ms        ? ?/sec
generator/run_block_generator block-4671894                          1.00    115.9±0.54ms        ? ?/sec    1.99    230.4±2.35ms        ? ?/sec
generator/run_block_generator block-4671894-compressed               1.00    116.1±0.90ms        ? ?/sec    1.94    225.8±0.35ms        ? ?/sec
generator/run_block_generator block-6fe59b24                         1.00     95.8±0.25ms        ? ?/sec    1.06    102.0±0.19ms        ? ?/sec
generator/run_block_generator block-6fe59b24-compressed              1.00     94.1±0.23ms        ? ?/sec    1.05     98.8±0.15ms        ? ?/sec
generator/run_block_generator block-834752                           1.00      7.5±0.10ms        ? ?/sec    1.02      7.7±0.01ms        ? ?/sec
generator/run_block_generator block-834752-compressed                1.00      7.6±0.15ms        ? ?/sec    1.01      7.6±0.06ms        ? ?/sec
generator/run_block_generator block-b45268ac                         1.00     84.0±0.59ms        ? ?/sec    1.08     90.7±0.23ms        ? ?/sec
generator/run_block_generator block-b45268ac-compressed              1.00     83.0±0.43ms        ? ?/sec    1.08     89.3±0.09ms        ? ?/sec
generator/run_block_generator block-c2a8df0d                         1.00    101.0±0.21ms        ? ?/sec    1.06    107.1±0.11ms        ? ?/sec
generator/run_block_generator block-c2a8df0d-compressed              1.00     99.9±0.10ms        ? ?/sec    1.05    105.2±0.31ms        ? ?/sec
generator/run_block_generator block-e5002df2                         1.00     89.3±0.97ms        ? ?/sec    1.07     95.1±0.21ms        ? ?/sec
generator/run_block_generator block-e5002df2-compressed              1.00     88.8±0.81ms        ? ?/sec    1.07     94.7±0.45ms        ? ?/sec
generator/run_block_generator deep-recursion-plus                    1.00   1128.9±2.08ms        ? ?/sec    1.70   1920.4±2.59ms        ? ?/sec
generator/run_block_generator deep-recursion-plus-compressed         1.00   1128.7±2.65ms        ? ?/sec    1.70   1918.4±2.36ms        ? ?/sec
generator/run_block_generator duplicate-coin-announce                1.00   1166.4±2.52ms        ? ?/sec    1.35   1577.1±2.59ms        ? ?/sec
generator/run_block_generator duplicate-coin-announce-compressed     1.00   1162.6±2.68ms        ? ?/sec    1.36   1576.2±1.33ms        ? ?/sec
generator/run_block_generator recursion-pairs                        1.00    897.7±0.82ms        ? ?/sec    1.18   1063.2±1.04ms        ? ?/sec
generator/run_block_generator recursion-pairs-compressed             1.00    899.0±1.83ms        ? ?/sec    1.18   1059.3±0.98ms        ? ?/sec
generator/run_block_generator2 block-1ee588dc                        1.00     49.2±0.19ms        ? ?/sec    1.03     50.5±0.18ms        ? ?/sec
generator/run_block_generator2 block-1ee588dc-compressed             1.00     48.9±0.28ms        ? ?/sec    1.00     48.9±0.10ms        ? ?/sec
generator/run_block_generator2 block-225758                          1.00      2.2±0.03ms        ? ?/sec    1.45      3.1±0.01ms        ? ?/sec
generator/run_block_generator2 block-225758-compressed               1.00      2.2±0.04ms        ? ?/sec    1.55      3.3±0.01ms        ? ?/sec
generator/run_block_generator2 block-4671894                         1.00    115.9±0.95ms        ? ?/sec    1.61    186.4±2.29ms        ? ?/sec
generator/run_block_generator2 block-4671894-compressed              1.00    115.3±0.95ms        ? ?/sec    1.58    181.7±0.36ms        ? ?/sec
generator/run_block_generator2 block-6fe59b24                        1.00     54.9±1.67ms        ? ?/sec    1.03     56.6±0.12ms        ? ?/sec
generator/run_block_generator2 block-6fe59b24-compressed             1.00     54.2±0.21ms        ? ?/sec    1.01     54.7±0.75ms        ? ?/sec
generator/run_block_generator2 block-834752                          1.02      4.8±0.10ms        ? ?/sec    1.00      4.7±0.06ms        ? ?/sec
generator/run_block_generator2 block-834752-compressed               1.00      4.6±0.08ms        ? ?/sec    1.00      4.6±0.01ms        ? ?/sec
generator/run_block_generator2 block-b45268ac                        1.00     48.5±0.66ms        ? ?/sec    1.01     48.9±0.38ms        ? ?/sec
generator/run_block_generator2 block-b45268ac-compressed             1.00     47.6±0.19ms        ? ?/sec    1.02     48.4±0.32ms        ? ?/sec
generator/run_block_generator2 block-c2a8df0d                        1.00     57.7±0.22ms        ? ?/sec    1.02     58.7±0.15ms        ? ?/sec
generator/run_block_generator2 block-c2a8df0d-compressed             1.00     55.1±0.08ms        ? ?/sec    1.02     56.2±0.32ms        ? ?/sec
generator/run_block_generator2 block-e5002df2                        1.00     51.5±0.12ms        ? ?/sec    1.04     53.4±0.14ms        ? ?/sec
generator/run_block_generator2 block-e5002df2-compressed             1.01     51.1±0.20ms        ? ?/sec    1.00     50.7±0.32ms        ? ?/sec
generator/run_block_generator2 deep-recursion-plus                   1.00   1128.5±3.57ms        ? ?/sec    1.70   1921.8±4.20ms        ? ?/sec
generator/run_block_generator2 deep-recursion-plus-compressed        1.00   1129.0±3.68ms        ? ?/sec    1.70   1921.4±1.90ms        ? ?/sec
generator/run_block_generator2 duplicate-coin-announce               1.00   1165.6±2.81ms        ? ?/sec    1.35   1577.9±0.84ms        ? ?/sec
generator/run_block_generator2 duplicate-coin-announce-compressed    1.00   1163.9±0.85ms        ? ?/sec    1.36   1577.4±1.00ms        ? ?/sec
generator/run_block_generator2 recursion-pairs                       1.00    896.1±1.20ms        ? ?/sec    1.19   1063.7±0.81ms        ? ?/sec
generator/run_block_generator2 recursion-pairs-compressed            1.00    896.5±1.36ms        ? ?/sec    1.19   1069.7±5.61ms        ? ?/sec
tree-hash/tree-hash block-225758                                     1.00    175.2±0.11µs        ? ?/sec    1.00    175.5±0.82µs        ? ?/sec
tree-hash/tree-hash block-225758-compressed                          1.00    175.4±0.31µs        ? ?/sec    1.00    175.5±1.77µs        ? ?/sec
tree-hash/tree-hash block-4671894                                    1.00      7.5±0.02ms        ? ?/sec    1.04      7.8±0.10ms        ? ?/sec
tree-hash/tree-hash block-4671894-compressed                         1.00      7.5±0.00ms        ? ?/sec    1.04      7.8±0.10ms        ? ?/sec
tree-hash/tree-hash block-834752                                     1.00      2.9±0.01ms        ? ?/sec    1.04      3.0±0.04ms        ? ?/sec
tree-hash/tree-hash block-834752-compressed                          1.00      2.9±0.04ms        ? ?/sec    1.01      2.9±0.00ms        ? ?/sec
```

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -2,6 +2,7 @@ use crate::err_utils::err;
 use crate::number::{node_from_number, number_from_u8, Number};
 use crate::reduction::EvalErr;
 use chia_bls::{G1Element, G2Element};
+use std::cell::RefCell;
 
 const MAX_NUM_ATOMS: usize = 62500000;
 const MAX_NUM_PAIRS: usize = 62500000;
@@ -12,13 +13,17 @@ const NODE_PTR_IDX_MASK: u32 = (1 << NODE_PTR_IDX_BITS) - 1;
 pub struct NodePtr(u32);
 
 enum ObjectType {
+    // The low bits form an index into the pair_vec
     Pair,
+    // The low bits form an index into the atom_vec
     Bytes,
+    // The low bits are the atom itself (unsigned integer, 26 bits)
+    SmallAtom,
 }
 
 // The top 6 bits of the NodePtr indicate what type of object it is
 impl NodePtr {
-    pub const NIL: Self = Self::new(ObjectType::Bytes, 0);
+    pub const NIL: Self = Self::new(ObjectType::SmallAtom, 0);
 
     const fn new(t: ObjectType, idx: usize) -> Self {
         debug_assert!(idx <= NODE_PTR_IDX_MASK as usize);
@@ -30,6 +35,7 @@ impl NodePtr {
             match self.0 >> NODE_PTR_IDX_BITS {
                 0 => ObjectType::Pair,
                 1 => ObjectType::Bytes,
+                2 => ObjectType::SmallAtom,
                 _ => {
                     panic!("unknown NodePtr type");
                 }
@@ -40,8 +46,9 @@ impl NodePtr {
 
     pub(crate) fn as_index(&self) -> usize {
         match self.node_type() {
-            (ObjectType::Pair, idx) => (idx as usize) * 2,
-            (ObjectType::Bytes, idx) => (idx as usize) * 2 + 1,
+            (ObjectType::Pair, idx) => (idx as usize) * 3,
+            (ObjectType::Bytes, idx) => (idx as usize) * 3 + 1,
+            (ObjectType::SmallAtom, idx) => (idx as usize) * 3 + 2,
         }
     }
 }
@@ -83,6 +90,7 @@ pub struct Checkpoint {
     u8s: usize,
     pairs: usize,
     atoms: usize,
+    small_atoms: usize,
 }
 
 #[derive(Debug)]
@@ -100,13 +108,57 @@ pub struct Allocator {
     // on.
     atom_vec: Vec<AtomBuf>,
 
+    // index into temp_buf array
+    temp_idx: RefCell<usize>,
+
+    // temporary buffers for storing SmallAtoms in to return from atom()
+    temp_vec: Vec<RefCell<[u8; 4]>>,
+
     // the atom_vec may not grow past this
     heap_limit: usize,
+
+    // the number of small atoms we've allocated. We keep track of these to ensure the limit on the
+    // number of atoms is identical to what it was before the small-atom optimization
+    small_atoms: usize,
 }
 
 impl Default for Allocator {
     fn default() -> Self {
         Self::new()
+    }
+}
+
+pub fn canonical_positive_integer(v: &[u8]) -> bool {
+    if v.is_empty() {
+        // empty buffer is 0/nil
+        true
+    } else if (v.len() == 1 && v[0] == 0)
+        // a 1-byte buffer of 0 is not the canonical representation of 0
+        || (v[0] & 0x80) != 0
+        // if the top bit is set, it's a negative number (i.e. not positive)
+        || (v[0] == 0 && (v[1] & 0x80) == 0)
+    {
+        // if the top byte is a 0 but the top bit of the next byte is not set, that's a redundant
+        // leading zero. i.e. not canonical representation
+        false
+    } else {
+        true
+    }
+}
+
+fn len_for_value(val: u32) -> usize {
+    if val == 0 {
+        0
+    } else if val < 0x80 {
+        1
+    } else if val < 0x8000 {
+        2
+    } else if val < 0x800000 {
+        3
+    } else if val < 0x80000000 {
+        4
+    } else {
+        5
     }
 }
 
@@ -119,20 +171,26 @@ impl Allocator {
         // we have a maximum of 4 GiB heap, because pointers are 32 bit unsigned
         assert!(heap_limit <= u32::MAX as usize);
 
+        let mut temp_vec = Vec::<RefCell<[u8; 4]>>::with_capacity(64);
+        for _ in 0..16 {
+            temp_vec.push(RefCell::default());
+        }
+
         let mut r = Self {
             u8_vec: Vec::new(),
             pair_vec: Vec::new(),
             atom_vec: Vec::new(),
-            heap_limit,
+            temp_idx: RefCell::new(0),
+            temp_vec,
+            // subtract 1 to compensate for the one() we used to allocate unconfitionally
+            heap_limit: heap_limit - 1,
+            // initialize this to 2 to behave as if we had allocated atoms for
+            // nil() and one(), like we used to
+            small_atoms: 2,
         };
         r.u8_vec.reserve(1024 * 1024);
         r.atom_vec.reserve(256);
         r.pair_vec.reserve(256);
-        r.u8_vec.push(1_u8);
-        // Preallocated empty list
-        r.atom_vec.push(AtomBuf { start: 0, end: 0 });
-        // Preallocated 1
-        r.atom_vec.push(AtomBuf { start: 0, end: 1 });
         r
     }
 
@@ -144,6 +202,7 @@ impl Allocator {
             u8s: self.u8_vec.len(),
             pairs: self.pair_vec.len(),
             atoms: self.atom_vec.len(),
+            small_atoms: self.small_atoms,
         }
     }
 
@@ -158,6 +217,7 @@ impl Allocator {
         self.u8_vec.truncate(cp.u8s);
         self.pair_vec.truncate(cp.pairs);
         self.atom_vec.truncate(cp.atoms);
+        self.small_atoms = cp.small_atoms;
     }
 
     pub fn new_atom(&mut self, v: &[u8]) -> Result<NodePtr, EvalErr> {
@@ -166,16 +226,41 @@ impl Allocator {
             return err(self.nil(), "out of memory");
         }
         let idx = self.atom_vec.len();
-        if idx == MAX_NUM_ATOMS {
+        if idx + self.small_atoms == MAX_NUM_ATOMS {
             return err(self.nil(), "too many atoms");
         }
-        self.u8_vec.extend_from_slice(v);
-        let end = self.u8_vec.len() as u32;
-        self.atom_vec.push(AtomBuf { start, end });
-        Ok(NodePtr::new(ObjectType::Bytes, idx))
+        if v.len() <= 3 && canonical_positive_integer(v) {
+            let mut ret: u32 = 0;
+            for b in v {
+                ret <<= 8;
+                ret |= *b as u32;
+            }
+            self.small_atoms += 1;
+            Ok(NodePtr::new(ObjectType::SmallAtom, ret as usize))
+        } else {
+            self.u8_vec.extend_from_slice(v);
+            let end = self.u8_vec.len() as u32;
+            self.atom_vec.push(AtomBuf { start, end });
+            Ok(NodePtr::new(ObjectType::Bytes, idx))
+        }
+    }
+
+    pub fn new_small_number(&mut self, v: u32) -> Result<NodePtr, EvalErr> {
+        debug_assert!(v <= NODE_PTR_IDX_MASK);
+        if self.atom_vec.len() + self.small_atoms == MAX_NUM_ATOMS {
+            return err(self.nil(), "too many atoms");
+        }
+        self.small_atoms += 1;
+        Ok(NodePtr::new(ObjectType::SmallAtom, v as usize))
     }
 
     pub fn new_number(&mut self, v: Number) -> Result<NodePtr, EvalErr> {
+        use num_traits::ToPrimitive;
+        if let Some(val) = v.to_u32() {
+            if val <= NODE_PTR_IDX_MASK {
+                return self.new_small_number(val);
+            }
+        }
         node_from_number(self, &v)
     }
 
@@ -197,33 +282,67 @@ impl Allocator {
     }
 
     pub fn new_substr(&mut self, node: NodePtr, start: u32, end: u32) -> Result<NodePtr, EvalErr> {
-        if self.atom_vec.len() == MAX_NUM_ATOMS {
+        if self.atom_vec.len() + self.small_atoms == MAX_NUM_ATOMS {
             return err(self.nil(), "too many atoms");
         }
-        let (ObjectType::Bytes, idx) = node.node_type() else {
-            return err(node, "(internal error) substr expected atom, got pair");
-        };
-        let atom = self.atom_vec[idx as usize];
-        let atom_len = atom.end - atom.start;
-        if start > atom_len {
-            return err(node, "substr start out of bounds");
+
+        fn bounds_check(node: NodePtr, start: u32, end: u32, len: u32) -> Result<(), EvalErr> {
+            if start > len {
+                return err(node, "substr start out of bounds");
+            }
+            if end > len {
+                return err(node, "substr end out of bounds");
+            }
+            if end < start {
+                return err(node, "substr invalid bounds");
+            }
+            Ok(())
         }
-        if end > atom_len {
-            return err(node, "substr end out of bounds");
+
+        match node.node_type() {
+            (ObjectType::Pair, _) => err(node, "(internal error) substr expected atom, got pair"),
+            (ObjectType::Bytes, idx) => {
+                let atom = self.atom_vec[idx as usize];
+                let atom_len = atom.end - atom.start;
+                bounds_check(node, start, end, atom_len)?;
+                let idx = self.atom_vec.len();
+                self.atom_vec.push(AtomBuf {
+                    start: atom.start + start,
+                    end: atom.start + end,
+                });
+                Ok(NodePtr::new(ObjectType::Bytes, idx))
+            }
+            (ObjectType::SmallAtom, val) => {
+                let len = len_for_value(val) as u32;
+                bounds_check(node, start, end, len)?;
+                let buf: [u8; 4] = val.to_be_bytes();
+                let buf = &buf[4 - len as usize..];
+                let substr = &buf[start as usize..end as usize];
+                if !canonical_positive_integer(substr) {
+                    let start = self.u8_vec.len();
+                    let end = start + substr.len();
+                    self.u8_vec.extend_from_slice(substr);
+                    let idx = self.atom_vec.len();
+                    self.atom_vec.push(AtomBuf {
+                        start: start as u32,
+                        end: end as u32,
+                    });
+                    Ok(NodePtr::new(ObjectType::Bytes, idx))
+                } else {
+                    let mut new_val: u32 = 0;
+                    for i in substr {
+                        new_val <<= 8;
+                        new_val |= *i as u32;
+                    }
+                    self.small_atoms += 1;
+                    Ok(NodePtr::new(ObjectType::SmallAtom, new_val as usize))
+                }
+            }
         }
-        if end < start {
-            return err(node, "substr invalid bounds");
-        }
-        let idx = self.atom_vec.len();
-        self.atom_vec.push(AtomBuf {
-            start: atom.start + start,
-            end: atom.start + end,
-        });
-        Ok(NodePtr::new(ObjectType::Bytes, idx))
     }
 
     pub fn new_concat(&mut self, new_size: usize, nodes: &[NodePtr]) -> Result<NodePtr, EvalErr> {
-        if self.atom_vec.len() == MAX_NUM_ATOMS {
+        if self.atom_vec.len() + self.small_atoms == MAX_NUM_ATOMS {
             return err(self.nil(), "too many atoms");
         }
         let start = self.u8_vec.len();
@@ -234,19 +353,29 @@ impl Allocator {
 
         let mut counter: usize = 0;
         for node in nodes {
-            let (ObjectType::Bytes, idx) = node.node_type() else {
-                self.u8_vec.truncate(start);
-                return err(*node, "(internal error) concat expected atom, got pair");
-            };
-
-            let term = self.atom_vec[idx as usize];
-            if counter + term.len() > new_size {
-                self.u8_vec.truncate(start);
-                return err(*node, "(internal error) concat passed invalid new_size");
+            match node.node_type() {
+                (ObjectType::Pair, _) => {
+                    self.u8_vec.truncate(start);
+                    return err(*node, "(internal error) concat expected atom, got pair");
+                }
+                (ObjectType::Bytes, idx) => {
+                    let term = self.atom_vec[idx as usize];
+                    if counter + term.len() > new_size {
+                        self.u8_vec.truncate(start);
+                        return err(*node, "(internal error) concat passed invalid new_size");
+                    }
+                    self.u8_vec
+                        .extend_from_within(term.start as usize..term.end as usize);
+                    counter += term.len();
+                }
+                (ObjectType::SmallAtom, val) => {
+                    let len = len_for_value(val) as u32;
+                    let buf: [u8; 4] = val.to_be_bytes();
+                    let buf = &buf[4 - len as usize..];
+                    self.u8_vec.extend_from_slice(buf);
+                    counter += len as usize;
+                }
             }
-            self.u8_vec
-                .extend_from_within(term.start as usize..term.end as usize);
-            counter += term.len();
         }
         if counter != new_size {
             self.u8_vec.truncate(start);
@@ -265,7 +394,44 @@ impl Allocator {
     }
 
     pub fn atom_eq(&self, lhs: NodePtr, rhs: NodePtr) -> bool {
-        self.atom(lhs) == self.atom(rhs)
+        match (lhs.node_type(), rhs.node_type()) {
+            ((ObjectType::Pair, _), _) | (_, (ObjectType::Pair, _)) => {
+                panic!("atom_eq() called on pair");
+            }
+            ((ObjectType::Bytes, lhs), (ObjectType::Bytes, rhs)) => {
+                let lhs = self.atom_vec[lhs as usize];
+                let rhs = self.atom_vec[rhs as usize];
+                self.u8_vec[lhs.start as usize..lhs.end as usize]
+                    == self.u8_vec[rhs.start as usize..rhs.end as usize]
+            }
+            ((ObjectType::SmallAtom, lhs), (ObjectType::SmallAtom, rhs)) => lhs == rhs,
+            ((ObjectType::SmallAtom, val), (ObjectType::Bytes, idx))
+            | ((ObjectType::Bytes, idx), (ObjectType::SmallAtom, val)) => {
+                let atom = self.atom_vec[idx as usize];
+                let len = len_for_value(val) as u32;
+                if (atom.end - atom.start) != len {
+                    return false;
+                }
+                if val == 0 {
+                    return true;
+                }
+
+                if self.u8_vec[atom.start as usize] & 0x80 != 0 {
+                    // SmallAtom only represents positive values
+                    // if the byte buffer is negative, they can't match
+                    return false;
+                }
+
+                // since we know the value of atom is small, we can turn it into a u32 and compare
+                // against val
+                let mut atom_val: u32 = 0;
+                for i in atom.start..atom.end {
+                    atom_val <<= 8;
+                    atom_val |= self.u8_vec[i as usize] as u32;
+                }
+                val == atom_val
+            }
+        }
     }
 
     pub fn atom(&self, node: NodePtr) -> &[u8] {
@@ -273,6 +439,21 @@ impl Allocator {
             (ObjectType::Bytes, idx) => {
                 let atom = self.atom_vec[idx as usize];
                 &self.u8_vec[atom.start as usize..atom.end as usize]
+            }
+            (ObjectType::SmallAtom, val) => {
+                let len = len_for_value(val);
+                let mut idx = self.temp_idx.borrow_mut();
+                *self.temp_vec[*idx].borrow_mut() = val.to_be_bytes();
+                let ret = unsafe {
+                    self.temp_vec[*idx]
+                        .try_borrow_unguarded()
+                        .expect("(internal error) temporary buffer problem in Allocator::atom()")
+                };
+                *idx += 1;
+                if *idx == self.temp_vec.len() {
+                    *idx = 0;
+                }
+                &ret[4 - len..]
             }
             _ => {
                 panic!("expected atom, got pair");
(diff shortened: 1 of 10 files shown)


Rigidity:
This is interesting, do the bench times increase in cases where most of the atoms are large due to the extra branching to determine the type of the NodePtr?

I'm also concerned about returning a reference to a temporary value with `atom()`, this feels like a technically memory safe, but logically unsafe thing to do.

But otherwise I like this a lot

arvidn:
> This is interesting, do the bench times increase in cases where most of the atoms are large due to the extra branching to determine the type of the NodePtr?

I think there's probably more noise in those timings than one might imagine. I just basically ran it once. But, yeah, there are a few cases where there's more work done. For example, when asking whether a node is a `small_int()`, I actually construct a small int, if it doesn't use the internal representation but would fit. This is probably not necessary.

There's also some extra work when allocating a new number. There's a check to see if it could be stored as a small int, and if so, it is.

> I'm also concerned about returning a reference to a temporary value with atom(), this feels like a technically memory safe, but logically unsafe thing to do.

Yeah, I'm not super happy about this. I think a ring buffer of 16 is more than enough, but maybe it's better to overshoot by more and make it 64 or so.

In the long run I would like to phase out this part of the API though, in favor of `visit_node()` (and the other getters)

This is still work in progress, as you can see from the TODO comments and missing tests

src/more_ops.rs
```diff
@@ -411,15 +440,32 @@ pub fn op_multiply(a: &mut Allocator, mut input: NodePtr, max_cost: Cost) -> Res
...
+                }
+                NodeVisitor::U32(val) => {
+                    total *= *val;
+                    let l1 = len_for_value(*val);
+
+                    cost += MUL_COST_PER_OP;
+                    cost += (l0 + l1) as Cost * MUL_LINEAR_COST_PER_BYTE;
+                    cost += (l0 * l1) as Cost / MUL_SQUARE_COST_PER_BYTE_DIVIDER;
+                    l0 = limbs_for_int(&total);
+                    Ok(())
```
Rigidity, line 464: This is repetitive. It could be refactored by having `l1` be the output of the `match`, and in the `Pair` case, early returning. Then do the rest of this below the `match`.

src/allocator.rs
```diff
@@ -39,14 +39,14 @@ impl NodePtr {
...
+            (self.0 & NODE_PTR_IDX_MASK),
         )
     }
 
     pub(crate) fn as_index(&self) -> usize {
         match self.node_type() {
-            (ObjectType::Pair, idx) => idx * 2,
-            (ObjectType::Bytes, idx) => idx * 2 + 1,
+            (ObjectType::Pair, idx) => (idx as usize) * 2,
+            (ObjectType::Bytes, idx) => (idx as usize) * 2 + 1,
```
richardkiss: Do we want `u64` here? Is it possible that `usize` can be a `u32` or is it guaranteed to be `u64` or bigger?

richardkiss: I guess `[]` indexing uses `usize` so if `usize` is `u32` we are already in trouble.

src/allocator.rs
```diff
@@ -2,8 +2,10 @@ use crate::err_utils::err;
 use crate::number::{node_from_number, number_from_u8, Number};
 use crate::reduction::EvalErr;
 use chia_bls::{G1Element, G2Element};
+use std::cell::RefCell;
 
-const MAX_NUM_ATOMS: usize = 62500000;
+// subtract 2 to compensate for nil() and one() atoms that we used to always add, unconditionally
+const MAX_NUM_ATOMS: usize = 62500000 - 2;
```
richardkiss, line 8: What tests do you have for this? Are they just "I expect this test to exceed the limit but minimally" and "I expect this test to hit the limit but just, and finish running" versus "I expect this test to use XXX atoms"? I have concern about this tightening (or worse, loosening) consensus.

src/allocator.rs
```diff
@@ -202,33 +286,69 @@ impl Allocator {
...
+                if end > len {
+                    return err(node, "substr end out of bounds");
+                }
+                if end < start {
+                    return err(node, "substr invalid bounds");
+                }
+                let buf: [u8; 4] = val.to_be_bytes();
+                let buf = &buf[4 - len as usize..];
+                let substr = &buf[start as usize..end as usize];
+                if !canonical_positive_integer(substr) {
```
richardkiss, line 333: I wonder if this whole block could be factored elsewhere. It seems hard to believe this would be the only place with this logic.

src/allocator.rs
```diff
@@ -202,33 +286,69 @@ impl Allocator {
...
-        self.atom_vec.push(AtomBuf {
-            start: atom.start + start,
-            end: atom.start + end,
-        });
-        Ok(NodePtr::new(ObjectType::Bytes, idx))
     }
 
     pub fn new_concat(&mut self, new_size: usize, nodes: &[NodePtr]) -> Result<NodePtr, EvalErr> {
-        if self.atom_vec.len() == MAX_NUM_ATOMS {
+        if self.atom_vec.len() + self.small_atoms == MAX_NUM_ATOMS {
```
richardkiss, line 357: Consider factoring this check with a self-documenting name.

src/allocator.rs
```diff
@@ -270,7 +400,44 @@ impl Allocator {
...
+            }
+            ((ObjectType::Bytes, lhs), (ObjectType::Bytes, rhs)) => {
+                let lhs = self.atom_vec[lhs as usize];
+                let rhs = self.atom_vec[rhs as usize];
+                self.u8_vec[lhs.start as usize..lhs.end as usize]
+                    == self.u8_vec[rhs.start as usize..rhs.end as usize]
+            }
+            ((ObjectType::SmallAtom, lhs), (ObjectType::SmallAtom, rhs)) => lhs == rhs,
+            ((ObjectType::SmallAtom, val), (ObjectType::Bytes, idx))
+            | ((ObjectType::Bytes, idx), (ObjectType::SmallAtom, val)) => {
```
richardkiss, line 421: Ha ha, sneaky.

src/allocator.rs
```diff
@@ -279,6 +446,21 @@ impl Allocator {
                 let atom = self.atom_vec[idx as usize];
                 &self.u8_vec[atom.start as usize..atom.end as usize]
             }
+            (ObjectType::SmallAtom, val) => {
+                let len = len_for_value(val);
+                let mut idx = self.temp_idx.borrow_mut();
+                *self.temp_vec[*idx].borrow_mut() = val.to_be_bytes();
+                let ret = unsafe {
```
richardkiss, line 459: Okay, so here's the ugly/scary part.

I wonder if you've thought about using 24 bits instead of 26 and then just using the storage for the index. Then you don't need this potentially scary temporary buffer hack.

richardkiss: In fact, then you can repurpose the other two bits (26 vs 24) to be the length of the small atom, 0-3.

src/allocator.rs
```diff
@@ -614,38 +872,54 @@ fn test_allocate_heap_limit() {
 fn test_allocate_atom_limit() {
     let mut a = Allocator::new();
 
-    for _ in 0..MAX_NUM_ATOMS - 2 {
-        // exhaust the number of atoms allowed to be allocated
+    for _ in 0..MAX_NUM_ATOMS {
+        // exhaust the numebr of atoms allowed to be allocated
```
richardkiss, line 901: `s/numebr/number/g`

src/allocator.rs
```diff
@@ -545,6 +551,25 @@ impl Allocator {
             .map_err(|_| EvalErr(node, "atom is not a G2 point".to_string()))
     }
 
+    pub fn visit_node<'a, T>(
+        &'a self,
+        node: NodePtr,
+        visitor: impl FnOnce(&NodeVisitor<'a>) -> T,
+    ) -> T {
+        match node.node_type() {
+            (ObjectType::Bytes, idx) => {
```
richardkiss, line 560: Is there any concern that a value that could be represented as a small atom sneaks into the wrong node type and the `visitor` then does the wrong thing?

richardkiss: Have you considered having this function (or maybe a different one) request the passed node to be represented in a particular way if possible? Then it could catch & cast the standard atom being incorrectly not coded as a small atom.

richardkiss: `visit_node_as_number`, `visit_node_as_u32`, etc.

src/run_program.rs
```diff
@@ -279,7 +279,15 @@ impl<'a, D: Dialect> RunProgramContext<'a, D> {
         // put a bunch of ops on op_stack
         let SExp::Pair(op_node, op_list) = self.allocator.sexp(program) else {
             // the program is just a bitfield path through the env tree
-            let r: Reduction = traverse_path(self.allocator, self.allocator.atom(program), env)?;
+            let r: Reduction = self.allocator.visit_node(program, |node| -> Response {
+                match node {
+                    NodeVisitor::Buffer(buf) => traverse_path(self.allocator, buf, env),
+                    NodeVisitor::U32(val) => traverse_path_fast(self.allocator, *val, env),
+                    NodeVisitor::Pair(_, _) => {
+                        panic!("expected atom, got pair");
```
richardkiss, line 287: Is this new behaviour? Shouldn't it be an `EvalError` rather than a panic?

src/allocator.rs
richardkiss: Another thing to note is that `Number` could be made an enum switching between u64 and bignum, which might simplify some things.

richardkiss:
If I understand correctly, this is actually a `SmallInt`. Maybe we should call it that or `SmallAtomInt`?

richardkiss:
This is some good work. The numbers in particular look fantastic. I do worry about subtle consensus changes and wonder if adding more `visit_node_as_XXX` functions would simplify and alleviate concerns that `SmallAtom` is always correctly coded whenever it's possible.

src/more_ops.rs
arvidn: it's not so easy to "early return" in the pair case. This code is inside a callback/lambda. So early return won't return from the function. i.e. it *is* already returning.

I'll explore factoring this out though.

src/allocator.rs
arvidn: maybe a more intuitive way of addressing this is to initialize `small_atoms` to 2. This is to compensate for no longer allocating `nil` and `one` in the constructor. I'll make this change.

The tests are:
`test_allocate_atom_limit()`
`test_allocate_small_number_limit()`
`test_allocate_substr_limit()`
`test_allocate_concat_limit()`

src/allocator.rs
arvidn: I think it might be a good idea to have something like `visit_node_as_bytes()`. It might deduplicate some instances if `u32` -> `[u8]` conversions. I'd like to punt on it until there's a clearer need for it. I'm envisioning making `G1`, `G2` and `Number` their own node types as well, at which point this might become a lot more convenient.

Regarding using `u64` or `i64` for arithmetic on small numbers, that was one of my first attempts at speeding things up, two years ago or so. It has pretty far reaching consequences, but I expect the end-game to incorporate something like that.

src/run_program.rs
arvidn: it's not new. Previously we called `self.allocator.atom(program)` which panics if the node is not an atom.

Rigidity (dismissed):
Looks good enough for now on my end, I hope to improve the API a bit to ideally get rid of `unsafe` and needing to use visitors, but I don't want to hold it back on that

src/allocator.rs
richardkiss: I'm pretty sure you can make a lifetime `c'` that depends on *two* lifetimes `a'` and `b'`. So the lifetime of the resulting atom could depend on both the `Node` and the `Allocator`. If you can get this to work, it could solve 1.

src/run_program.rs
richardkiss: I see, the "it's not a pair" check has already been done at this point.

arvidn: well, in the new version, we call `visit_node()` instead of `atom()`. so this is really the counterpart to the one in `atom()`

src/allocator.rs
arvidn: I suppose the 3rd problem would be byte order.

arvidn:
@Rigidity pointed out that `visit_node()` could probably return the `NodeVisitor`, rather than taking a callback. Doing that sounds like an improvement to me. I'll investigate that.

arvidn:
I applied the suggestion from @Rigidity and made `visit_node()` return the `NodeVisitor` rather than taking a callback function to pass it to. I also renamed the function just `node()`.

The change since last version: https://github.com/Chia-Network/clvm_rs/compare/34bbdf08f7d0c59fa3312b66aa1d3173839b6866..e90cd1279ae60da5f21953486c3a454f583ca217

Rigidity (dismissed):
lgtm for now

src/allocator.rs
richardkiss: Byte order would actually *not* be a problem, if you consider the 24 bits to be an atom byte string (of up to length 3) rather than a small integer.

This might change quite a few semantics about this patch, but I believe it might actually end up simpler since the two atom types would be much more similar since they would both be `&[u8]`. If you add a `visit_as_int`, you could better optimize the int conversion to a couple cpu instructions.

arvidn: it would change the trajectory of this change and veer away from making `BigInt`, `G1Element` and `G2Element` their own node types. Even though conversion to and from `u32` and `i32` is cheap, those types are expensive.

arvidn: I prefer the direction of moving away from the `atom()` API returning pointers

Rigidity (approved):

PR #374, opened 2024-02-08
remove NodePtr hack and fix-up benchmark-clvm-cost

arvidn:
This fixes `benchmark-clvm-cost` to not need a magic `NodePtr` value for a variable placeholder. Instead, it uses `Option<NodePtr>` as it should have.

This let us remove the `NodePtr::hack()` constructor.

### DIFF
--- a/tools/src/bin/benchmark-clvm-cost.rs
+++ b/tools/src/bin/benchmark-clvm-cost.rs
@@ -8,15 +8,18 @@ use std::io::{sink, Write};
 use std::time::Instant;
 
 #[derive(Clone, Copy)]
+enum Placeholder {
+    SingleArg(Option<NodePtr>),
+    TwoArgs(Option<NodePtr>, Option<NodePtr>),
+    ThreeArgs(Option<NodePtr>, Option<NodePtr>, Option<NodePtr>),
+}
+
+#[derive(Clone, Copy)]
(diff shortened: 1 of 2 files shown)


tools/src/bin/benchmark-clvm-cost.rs
```diff
@@ -98,23 +101,17 @@ fn quote(a: &mut Allocator, v: NodePtr) -> NodePtr {
     a.new_pair(a.one(), v).unwrap()
 }
 
-fn subst_node(arg: NodePtr, substitution: NodePtr) -> NodePtr {
-    if arg == NodePtr::hack(VARIABLE_VAL) {
-        substitution
-    } else {
-        arg
-    }
+fn subst(arg: Option<NodePtr>, substitution: NodePtr) -> NodePtr {
```
Rigidity, line 104: Do we need this function?

Rigidity (approved):

tools/src/bin/benchmark-clvm-cost.rs
arvidn: well, it saves a few characters when it's called

arvidn: I would approve a PR that removes it :)

PR #379, opened 2024-02-12
fix small_number() patch

arvidn:
The interpreter relies on being able to retrieve small atoms via small_number() even if they happen to be allocated on the heap, as normal atoms (but small).

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -117,6 +117,12 @@ pub enum Atom<'a> {
     U32([u8; 4], usize),
 }
 
+impl PartialEq for Atom<'_> {
+    fn eq(&self, other: &Atom) -> bool {
+        self.as_ref().eq(other.as_ref())
+    }
+}
+
 impl<'a> AsRef<[u8]> for Atom<'a> {
     fn as_ref(&self) -> &[u8] {
         match self {
@@ -368,6 +374,9 @@ impl Allocator {
         if self.heap_limit - start < new_size {
             return err(self.nil(), "out of memory");
         }
+        // TODO: maybe it would make sense to have a special case where
+        // nodes.len() == 1. We can just return the same node
+
         self.u8_vec.reserve(new_size);
 
         let mut counter: usize = 0;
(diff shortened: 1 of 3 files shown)


Rigidity (approved):

src/allocator.rs
```diff
@@ -497,6 +506,20 @@ impl Allocator {
     pub fn small_number(&self, node: NodePtr) -> Option<u32> {
         match node.object_type() {
             ObjectType::SmallAtom => Some(node.index()),
+            ObjectType::Bytes => {
+                let atom = self.atom_vec[node.index() as usize];
+                let buf = &self.u8_vec[atom.start as usize..atom.end as usize];
+                if buf.len() > 3 || !canonical_positive_integer(buf) {
```
richardkiss, line 512: Why 3? Wouldn't it be okay to allow integers of size 4 that can still be represented by a u32?

richardkiss: Oh wait, I get it now.

arvidn: you're right that it would be possible to represent most 4-byte buffers as u32. This rule, however, is consistent with whether it can be represented as a `SmallAtom` node. Which is 26 bits (i.e. 3 bytes + 2 bits, we're rounding that to 3 bytes here).

This is the check in other places too, suche as `fn new_atom(&[u8]) -> NodePtr`. Perhaps it would make sense to factor out this check to make sure it's consistent everywhere, and maybe it would even make sense to make it exactly 26 bits (rather than 24).

For instance, it just occurred to me that the value `0x800000` would not be considered fitting in a `SmallAtom`, even though it would, just because the leading zero byte would bump it to 4.

Anyway, the two approaches I see are:

1. factor out this logic, make it more precise (to cover more large numbers)
2. make this logic special, since we don't technically *need* it to fit in a `SmallAtom`, we just need it to fit in a `u32`

I'm leaning towards (1).

arvidn: I will propose a patch to improve this situation, aiming for (1). i.e. making this function behave consistently with whether an atom can be represented as a `SmallAtom` rather than whether it can be represented as a `u32`

arvidn: https://github.com/Chia-Network/clvm_rs/pull/384

PR #383, opened 2024-02-13
add fuzzer for Allocator

arvidn:
specifically handling of `atom()`, `small_number()` and `number()`.

This checks certain properties that we expect to hold for any atom (conditioned on whether it's an canonical form integer or not).

This functions as a property-based test.

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -1082,7 +1082,9 @@ fn test_concat_launder_small_number() {
 
     // this "launders" the small number into actually being allocated on the
     // heap
-    let atom2 = a.new_concat(1, &[atom1]).expect("new_substr");
+    let atom2 = a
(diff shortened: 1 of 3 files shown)


src/allocator.rs
```diff
@@ -1082,7 +1082,9 @@ fn test_concat_launder_small_number() {
 
     // this "launders" the small number into actually being allocated on the
     // heap
-    let atom2 = a.new_concat(1, &[atom1]).expect("new_substr");
+    let atom2 = a
+        .new_concat(1, &[a.nil(), atom1, a.nil()])
```
arvidn, line 1086: this is somewhat unrelated. it makes it more likely that this `concat` actually causes a heap allocation for the return value

Rigidity (approved):
lgtm

PR #384, opened 2024-02-13
improve fidelity of check if atom fits in `SmallAtom`

arvidn:
This factors out some common code and generally simplifies and unifies the behavior. With additional tests and fuzzer checks.

The second commit addresses the use of `as_unique_address()`, which is used in `ObjectCache` as a cheap hash map (index into a `Vec`). However, with the `SmallAtom` feature, `NodePtr` indices is no longer dense, and it no longer works to use a `Vec` instead of a `HashMap`. It allocates too much memory if you create an atom with value `0x3ffffff`.

This commit is included in this PR since the increased fidelity of matching atoms to `SmallAtom` pushed the fuzzer over the limit and started failing with out-of-memory.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -190,8 +190,9 @@ jobs:
         with:
           fetch-depth: 1
       - name: Install rust
-        uses: dtolnay/rust-toolchain@nightly
+        uses: dtolnay/rust-toolchain@master
         with:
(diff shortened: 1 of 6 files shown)


Rigidity (dismissed):

Rigidity (dismissed):

Rigidity (approved):
lgtm

PR #393, opened 2024-03-15
typos

richardkiss:


src/more_ops.rs
```diff
@@ -82,7 +82,7 @@ const PUBKEY_BASE_COST: Cost = 1325730;
 const PUBKEY_COST_PER_BYTE: Cost = 38;
 
 // the new coinid operator
-// we subtract 153 cost as a discount, to incentivice using this operator rather
+// we subtract 153 cost as a discount, to incentive using this operator rather
```
arvidn, line 85:
```suggestion
// we subtract 153 cost as a discount, to incentivize using this operator rather
```

arvidn: I suppose you could say "as incentive" instead

arvidn (approved):

PR #394, opened 2024-03-25
Update Patched Fix `openssl` `X509StoreRef::objects` is unsound

bangtabil:
This function returned a reference into an OpenSSL datastructure, but there was no way to ensure OpenSSL would not mutate the datastructure behind one's back.

Use of this function should be replaced with `X509StoreRef::all_certificates.`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -110,23 +110,6 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "chia-bls"
-version = "0.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ba003303bda3019f2fe94e5478555a96addff1d6a1ea84bfe207b3f075cd8420"
-dependencies = [
- "anyhow",
- "arbitrary",
- "blst",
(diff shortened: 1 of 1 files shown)


Rigidity:
Hey, I don't quite follow this - there aren't any code changes other than a version bump in the lockfile, so I'm not sure if this would actually fix the mentioned issue?

I believe we only work with X.509 certificates in the `chia-ssl` crate, and certainly don't use the `X509StoreRef::objects` method. So it would be in a downstream dependency, which we should bump in `Cargo.toml` if it's an issue.

Rigidity:
I see, the relevant issue is https://github.com/sfackler/rust-openssl/issues/2096?

Looks like CI is failing, so will have to look into that. And ideally bump whichever crate indirectly depends on OpenSSL as well.

arvidn:
this should also be addressed in the `Cargo.toml` file, right?
Also, once addressed, we should remove this exception: https://github.com/Chia-Network/clvm_rs/blob/main/.github/workflows/dependency-review.yml#L22C24-L22C43

jack60612:
thanks for your contribution, this was merged into one big pr and included

jack60612:
Thanks! Included in #435.

PR #395, opened 2024-04-03
Support backref for wasm

ChiaMineJP:
Added support compressed form of clvm byte representation.

## Changed
- Renamed `wasm/src/api.rs` to `wasm/src/run_program.rs`
- Added `allow_backrefs()` method to `Flag` struct
- `run_clvm` requires `flag` argument. If `ALLOW_BACKREFS` bit is set, it deserializes bytes input of clvm objects as compressed forms.
- When `ALLOW_BACKREFS` bit is set when calling `run_chia_program`, it deserializes bytes input of clvm objects as compressed forms.

## Added
- Added `sexp_from_bytes` function.
- Added `to_bytes()` method to `LazyNode`


wasm/src/serialize.rs
```diff
@@ -0,0 +1,27 @@
...
+use clvmr::{Allocator, ALLOW_BACKREFS};
+use std::rc::Rc;
+use wasm_bindgen::prelude::wasm_bindgen;
+
+#[wasm_bindgen]
+pub fn serialized_length(program: &[u8]) -> Result<u64, String> {
+    match serialized_length_from_bytes(program) {
+        Ok(length) => Ok(length),
+        Err(err) => Err(err.to_string()),
+    }
```
arvidn, line 12: an alternative way of saying this is:
```
serialized_length_from_bytes(program).map_err(|x| x.to_string())
```

wasm/src/serialize.rs
```diff
@@ -0,0 +1,27 @@
...
+    let mut allocator = Allocator::new();
+    let deserializer = if (flag & ALLOW_BACKREFS) != 0 {
+        node_from_bytes_backrefs
+    } else {
+        node_from_bytes
+    };
+    match deserializer(&mut allocator, b) {
+        Ok(node) => Ok(LazyNode::new(Rc::new(allocator), node)),
+        Err(err) => Err(err.to_string()),
+    }
```
arvidn, line 26: same thing here. `map_err()` is probably what you want.

wasm/src/lazy_node.rs
```diff
@@ -35,6 +37,19 @@ impl LazyNode {
             _ => None,
         }
     }
+
+    #[wasm_bindgen]
+    pub fn to_bytes(&self, flag: u32) -> Option<Vec<u8>> {
+        let serializer = if (flag & ALLOW_BACKREFS) != 0 {
+            node_to_bytes_backrefs
+        } else {
+            node_to_bytes
```
arvidn, line 46: this case is risky. If you ever end up here on an untrusted `LazyNode`, you may blow up memory usage. The main reason we have `LazyNode` is to avoid serializing it to bytes

wasm/src/lazy_node.rs
```diff
@@ -35,6 +37,19 @@ impl LazyNode {
...
+        };
+        match serializer(&self.allocator, self.node) {
+            Ok(b) => Some(b),
+            Err(_) => None,
+        }
```
arvidn, line 51:
```suggestion
serializer(&self.allocator, self.node).ok()
```

src/flags.rs
```diff
@@ -0,0 +1,3 @@
+// when this flag is set, the block generator serialization is allowed to
+// contain back-references
+pub const ALLOW_BACKREFS: u32 = 0x2000000;
```
arvidn, line 3: it's pretty scary to duplicate this constant with the same name and same value as in `chia_rs`. It could be mistaken for being the same flag.

I think it would be better to use a different name and different value, to make it clear that it's independent from the flag in `chia_rs`. It should also probably live in the wasm binding, since that's the only place it's used.

At some point we'll have type safe flags, and then this will be less of a concern.

richardkiss: I agree that a more natural API would just be to expose both the "with backreferences" and "without backreferences" calls separately, since they're quite different, or maybe include a boolean flag. If you want to the API to resemble that `chia_rs` api, that's an argument for using a mask of flags.

wasm/src/lazy_node.rs
richardkiss: That's something I hadn't thought of: compression mitigates this DOS vector. However, the compression code is quite complex and I don't know if there aren't DOS vectors there too. They're less obvious though. :)

src/flags.rs
richardkiss: At the very least, move the flag and the branch to the `wasm` api.

richardkiss (changes requested):
Consider moving `ALLOW_BACKREFS` to `wasm` api, or simply exporting `node_to_bytes_backrefs` separately.

wasm/src/lazy_node.rs
ChiaMineJP: @arvidn How about adding `node_to_bytes_backrefs_limit` and update `node_to_bytes_backrefs` as:
```rust
pub fn node_to_bytes_backrefs(a: &Allocator, node: NodePtr) -> io::Result<Vec<u8>> {
    node_to_bytes_backrefs_limit(a, node, 2000000)
}
```

wasm/src/serialize.rs
ChiaMineJP: I'll adopt this suggestion.

wasm/src/serialize.rs
ChiaMineJP: Unlike the above, this code needs to map `Ok(NodePtr)` to `Ok(LazyNode)`
So I'll change the code as:
```rust
    let node = deserializer(&mut allocator, b).map_err(|e| e.to_string())?;
    Ok(LazyNode::new(Rc::new(allocator), node))
```

src/flags.rs
ChiaMineJP: I'd like to keep the value `0x2000000` the same as `chia_rs` because if changing the value, wasm/JavaScript developers needs special code to make it compatible with other chia applications.
I think `clvm_rs` is more likely responsible for serialization (thus `ALLOW_BACKREFS` flag) rather than `chia_rs` so in my heart `ALLOW_BACKREFS` should be defined in `clvm_rs` and not in `chia_rs`...

But it's OK to hide the value to `wasm/` code and use only in the wasm.
I don't want to change the grand design and relationship between `chia_rs` and `clvm_rs` for now.

ChiaMineJP:
> Consider moving `ALLOW_BACKREFS` to `wasm` api, or simply exporting `node_to_bytes_backrefs` separately.

I'll move `ALLOW_BACKREFS` to `wasm` api.
In Rust part of `clvm_rs`, whether to allow backrefs will be decided in the upper scope (like `chia_rs`).
However, in wasm/js part of `clvm_rs`, developers must tell to allow backref or not in the `clvm_wasm`(`clvm_rs`)'s scope because from JS world, SExp is always passed as uint8 byte array.
So whether the byte stream is compressed or not must be immediately known in `run_chia_program` or `run_clvm` and cannot defer.

wasm/src/lazy_node.rs
arvidn: As long as the case where the output is truncated is handled correctly, and not assumed to be an OK result. The function you suggest will silently ignore the failure though, which seems a bit risky.
Also, this comment was made on `node_to_bytes`, i.e. the version that doesn't support back-refs.

src/serde/ser_br.rs
```diff
@@ -77,10 +78,45 @@ pub fn node_to_stream_backrefs<W: io::Write>(
     Ok(())
 }
 
+pub fn node_to_bytes_backrefs_limit(
```
arvidn, line 81: I think this was a misunderstanding. it's the serialization *without* back-references that's at risk of blowing up the output buffer size, not the one where we allow back-refs.

arvidn (changes requested):

src/serde/ser_br.rs
richardkiss: Arvid is correct.

src/serde/mod.rs
```diff
@@ -18,8 +18,8 @@ mod test;
 pub use de::node_from_bytes;
 pub use de_br::{node_from_bytes_backrefs, node_from_bytes_backrefs_record};
 pub use de_tree::{parse_triples, ParsedTriple};
-pub use ser::node_to_bytes;
-pub use ser_br::node_to_bytes_backrefs;
+pub use ser::{node_to_bytes, node_to_bytes_limit};
+pub use ser_br::{node_to_bytes_backrefs, node_to_bytes_backrefs_limit};
```
richardkiss, line 22: I think `node_to_bytes_limit` that returns a result is a good idea. As Arvid mentions, we probably do not need `node_to_bytes_backrefs_limit`. It may be possible to DOS the serialization with certain clvm trees, but if it is, they are unknown at this point, and I don't think out-of-memory will be the result (and definitely not an exponentially large serialization like in the uncompressed case).

richardkiss: Hmm, on the other hand, creating this API probably doesn't hurt. There's not really a way to limit output memory any other way.

src/serde/ser_br.rs
ChiaMineJP: Thanks for comment. Also I discussed this with Arvid on keybase. I recognized my misunderstanding.
`LazyNode` with Allocator/NodePtr is some kind of "compressed" form because the same atom is stored only once in Allocator. This is almost the same concept with backref. When serializing this kind of `LazyNode` without backref, it does something similar to "uncompressing" the data.

src/serde/mod.rs
ChiaMineJP: OK, I'll strip `node_to_bytes_backrefs_limit` off from `mod.rs`.

ChiaMineJP: Done!

ChiaMineJP: Seems like we passed the each other.
Do you think `node_to_bytes_backrefs_limit` should be exposed in this PR?

ChiaMineJP: @richardkiss I've re-exposed `node_to_bytes_backrefs_limit`.

src/serde/ser_br.rs
```diff
@@ -77,10 +78,45 @@ pub fn node_to_stream_backrefs<W: io::Write>(
...
+    let mut writer = LimitedWriter::new(buffer, limit);
+    node_to_stream_backrefs(a, node, &mut writer)?;
+    let vec = writer.into_inner().into_inner();
+    Ok(vec)
+}
+
 pub fn node_to_bytes_backrefs(a: &Allocator, node: NodePtr) -> io::Result<Vec<u8>> {
-    let mut buffer = Cursor::new(Vec::new());
+    node_to_bytes_backrefs_limit(a, node, 2000000)
+}
```
arvidn, line 95: I don't think you should change this function. it should not have a default limit.

wasm/src/serialize.rs
```diff
@@ -0,0 +1,24 @@
...
+use clvmr::serde::{node_from_bytes, node_from_bytes_backrefs, serialized_length_from_bytes};
+use clvmr::Allocator;
+
+#[wasm_bindgen]
+pub fn serialized_length(program: &[u8]) -> Result<u64, String> {
+    serialized_length_from_bytes(program).map_err(|x| x.to_string())
+}
+
+#[wasm_bindgen]
+pub fn sexp_from_bytes(b: &[u8], flag: u32) -> Result<LazyNode, String> {
```
arvidn, line 15: is this new? maybe `node_from_bytes` would be a more uniform name.

arvidn (changes requested):
my main comment is that I don't think `node_to_bytes_backrefs()` should have a default size limit

wasm/src/serialize.rs
ChiaMineJP: Right, but `node_from_bytes` already exists.
I moved this code to `LazyNode::from_bytes()`

src/serde/ser_br.rs
ChiaMineJP: Right. I reverted my code.

wasm/src/serialize.rs
richardkiss: The `SExp` terminology originated from me, based on words use in lisp. I admit that I did a bad job describing differentiating between `CLVMObject`, `SExp`, `Program` and `Node`, and this has made things unduly confusing.

I think using `Node` in `clvm_rs` makes sense, and maybe we should avoid creating new API that uses the term `SExp`.

ChiaMineJP: > what do you mean? is there a wasm binding for `node_from_bytes` already?

There is a `node_from_bytes` in the original source.

image

I didn't like to confuse people by attaching an existing function name to a new function even if it is used in the different context (wasm).

ChiaMineJP: image

I think moving `node_from_bytes()` to `LazyNode::from_bytes()` looks concise and understandable.

arvidn: it seems resonable to use the same names/terminology in wasm as we do in the python binding and rust

ChiaMineJP: Ok, then using the same name is more consistent...
By the way, which way do you feel preferable?
1. `sexp_from_bytes()` => `LazyNode::from_bytes()`
2. `sexp_from_bytes()` => `node_from_bytes()`

arvidn: I think (2) makes the most sense. Primarily because of consistency with the rust and python API. I'm not opposed to making the API more like (1), but then I think we also need to change the rust and python APIs (which would be quite involved, since it would spill over onto other code bases)

arvidn (approved):

richardkiss (approved):

wasm/src/serialize.rs
arvidn: are we keeping this name?

ChiaMineJP: Sorry my mistake.
I've just renamed `sexp_from_bytes` to `node_from_bytes`

wasm/src/serialize.rs
```diff
@@ -12,12 +14,12 @@ pub fn serialized_length(program: &[u8]) -> Result<u64, String> {
...
 
 #[wasm_bindgen]
-pub fn sexp_from_bytes(b: &[u8], flag: u32) -> Result<LazyNode, String> {
+pub fn node_from_bytes(b: &[u8], flag: u32) -> Result<LazyNode, String> {
     let mut allocator = Allocator::new();
     let deserializer = if (flag & ALLOW_BACKREFS) != 0 {
         node_from_bytes_backrefs
     } else {
-        node_from_bytes
+        _node_from_bytes
```
arvidn, line 22: it's also possible to use a fully qualified name here, instead of an alias:
```
clvmr::serde::node_from_bytes
```

arvidn (approved):
looks good!

PR #398, opened 2024-04-25
[draft] [opinions welcome] Change the individual callback functions in run_program to a trait which simplifies lifetime management

prozacchiwawa:
This makes PreEval a trait which is passed down the stack as a mutable reference.  It contains both pre_eval and post_eval methods and calls post_eval when requested based on the result of pre_eval as before.

Because the callback's lifetime is now easily determined to be the lifetime of the run_program call and the other object that could be of use, the allocator is also passed by a mutable reference, the caller has maximal freedom in how their implementation of PreEval works without having to use the Rc<RefCell<...>> pattern and cunning use of move closures to create the callbacks and avoids construction of a Box on each callback for the second closure.

Since this changes the form of the run_program interface, commentary on that is welcome.  This is a response to the arch meeting discussion rather than something I'm actively requesting, but I do think this is better overall.

### DIFF
--- a/src/run_program.rs
+++ b/src/run_program.rs
@@ -19,15 +19,33 @@ const OP_COST: Cost = 1;
 // exceeded
 const STACK_SIZE_LIMIT: usize = 20000000;
 
+/// Tell whether to call the post eval function or not, giving a reference id
+/// for the computation to pick up.
 pub enum PreEvalResult {
     CallPostEval(usize),
-    Done
+    Done,
 }
 
 #[cfg(feature = "pre-eval")]
+/// Implementing this trait allows an object to be notified of clvm operations
+/// being performed as they happen.
 pub trait PreEval {
-    fn pre_eval(&mut self, allocator: &mut Allocator, sexp: NodePtr, args: NodePtr) -> Result<PreEvalResult, EvalErr>;
-    fn post_eval(&mut self, _allocator: &mut Allocator, _pass: usize, _result: Option<NodePtr>) -> Result<(), EvalErr> {
+    /// pre_eval is called before the operator is run, giving sexp (the operation
+    /// to run) and args (the environment).
+    fn pre_eval(
+        &mut self,
+        allocator: &mut Allocator,
+        sexp: NodePtr,
+        args: NodePtr,
+    ) -> Result<PreEvalResult, EvalErr>;
+    /// post_eval is called after the operation was performed.  When the clvm
+    /// operation resulted in an error, result is None.
+    fn post_eval(
+        &mut self,
+        _allocator: &mut Allocator,
+        _pass: usize,
+        _result: Option<NodePtr>,
+    ) -> Result<(), EvalErr> {
         Ok(())
     }
 }
@@ -122,7 +140,7 @@ fn augment_cost_errors(r: Result<Cost, EvalErr>, max_cost: NodePtr) -> Result<Co
     })
 }
 
-impl<'a, 'inner, D: Dialect> RunProgramContext<'a, D> {
+impl<'a, D: Dialect> RunProgramContext<'a, D> {
     #[cfg(feature = "counters")]
     #[inline(always)]
     fn account_val_push(&mut self) {
@@ -276,7 +294,9 @@ impl<'a, 'inner, D: Dialect> RunProgramContext<'a, D> {
     fn eval_pair(&mut self, program: NodePtr, env: NodePtr) -> Result<Cost, EvalErr> {
         #[cfg(feature = "pre-eval")]
         if let Some(pre_eval) = &mut self.pre_eval {
-            if let PreEvalResult::CallPostEval(pass) = pre_eval.pre_eval(&mut self.allocator, program, env)? {
+            if let PreEvalResult::CallPostEval(pass) =
+                pre_eval.pre_eval(self.allocator, program, env)?
+            {
                 self.posteval_stack.push(pass);
                 self.op_stack.push(Operation::PostEval);
             }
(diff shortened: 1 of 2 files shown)


prozacchiwawa:
thanks @Rigidity for the typedef suggestion and catching the ungated declaration.

src/run_program.rs
```diff
@@ -20,11 +20,39 @@ const OP_COST: Cost = 1;
...
+        sexp: NodePtr,
+        args: NodePtr,
+    ) -> Result<PreEvalResult, EvalErr>;
+    /// post_eval is called after the operation was performed.  When the clvm
+    /// operation resulted in an error, result is None.
+    fn post_eval(
+        &mut self,
+        _allocator: &mut Allocator,
+        _pass: PreEvalIndex,
+        _result: Option<NodePtr>,
```
richardkiss, line 51: I wonder why I did not just include the result verbatim, ie. use `Result` instead of `Option`. Maybe it was trying to be more like the original python? I wouldn't necessarily be opposed to changing this.

src/run_program.rs
```diff
@@ -20,11 +20,39 @@ const OP_COST: Cost = 1;
...
+        allocator: &mut Allocator,
+        sexp: NodePtr,
+        args: NodePtr,
+    ) -> Result<PreEvalResult, EvalErr>;
+    /// post_eval is called after the operation was performed.  When the clvm
+    /// operation resulted in an error, result is None.
+    fn post_eval(
+        &mut self,
+        _allocator: &mut Allocator,
+        _pass: PreEvalIndex,
```
richardkiss, line 50: Is this just a nonce to match with the `PreEval` `sexp/args` pair? I suppose `_pass` means something when this is used with the classic compiler, but I wonder if a more generic name might be better.

src/run_program.rs
```diff
@@ -269,9 +297,11 @@ impl<'a, D: Dialect> RunProgramContext<'a, D> {
 
     fn eval_pair(&mut self, program: NodePtr, env: NodePtr) -> Result<Cost, EvalErr> {
         #[cfg(feature = "pre-eval")]
-        if let Some(pre_eval) = &self.pre_eval {
-            if let Some(post_eval) = pre_eval(self.allocator, program, env)? {
-                self.posteval_stack.push(post_eval);
+        if let Some(pre_eval) = &mut self.pre_eval {
+            if let PreEvalResult::CallPostEval(pass) =
+                pre_eval.pre_eval(self.allocator, program, env)?
+            {
```
richardkiss, line 303: `cargo fmt` prob needed

src/run_program.rs
```diff
@@ -484,18 +514,20 @@ impl<'a, D: Dialect> RunProgramContext<'a, D> {
...
             };
             cost += match op {
-                Operation::Apply => augment_cost_errors(
-                    self.apply_op(cost, effective_max_cost - cost),
-                    max_cost_ptr,
-                )?,
+                Operation::Apply => {
+                    let apply_op_res = self.apply_op(cost, effective_max_cost - cost);
+                    augment_cost_errors(apply_op_res, max_cost_ptr)?
+                }
```
richardkiss, line 520: It seems like this little `diff` is the exact same code, yes?

src/run_program.rs
```diff
@@ -20,11 +20,39 @@ const OP_COST: Cost = 1;
...
+/// for the computation to pick up.
+pub enum PreEvalResult {
+    CallPostEval(PreEvalIndex),
+    Done,
+}
+
+#[cfg(feature = "pre-eval")]
+/// Implementing this trait allows an object to be notified of clvm operations
+/// being performed as they happen.
+pub trait PreEval {
```
richardkiss, line 36: I wonder if there's now a better name for this, since it could potentially implement both pre and post eval. Maybe `Tracer` or `EvalLogger` or `InvocationTracker` or `CallMonitor` or something. (Thanks to ChatGPT for some of these ideas.)

`PreEval` is maybe more useful for people who know the code already, but that's maybe eight out of eight billion. :)

richardkiss (commented):
Besides some naming nits (and possibly turning an `Option` a `Result` as it seems like it may have been a mistake in the original API), it seems like a pretty reasonable way for this code to evolve. I think you (Arty) are the only one using this functionality these days, and it seems like a clear win.

Arvid already has gated `PreEval` behind a feature flag so we don't have to worry much about it affecting performance in production.

I didn't look closely at the tests, but I assume you just made whatever changes you needed to make the tests work.

src/run_program.rs
prozacchiwawa: Hmm ya, the name could improve.  I was thinking 'passthrough' but will give it some thought.

src/run_program.rs
prozacchiwawa: ya changed that back.

src/run_program.rs
prozacchiwawa: PreEvalResult::CallPostEval is longer so it re-wraps a bit.  This is post fmt.

Rigidity: Yeah fmt can be a bit funky at times, but that's the price of consistency I suppose :P

Rigidity:
I think performance should still be a consideration, since features are additive - if you depend on `clvm_tools_rs` and `clvmr` in your project for instance, and `clvm_tools_rs` enables the feature, its performance effects will be seen on your usage of `clvmr` even if you don't specify the feature yourself. So we ideally wouldn't want it to be considerably slower, but I'm guessing the compiler can optimize this to around the same as before.

src/run_program.rs
richardkiss: Interesting, I don't recall having seen that lone `{` before.

PR #403, opened 2024-05-10
CHIA-255: implement CHIP-34

arvidn:
this PR is best reviewed one commit at a time.

It implements [CHIP-34](https://github.com/Chia-Network/chips/pull/116).

## new operators

There are 3 new operators enabled under softfork `1`:

| operator | opcode |
| --- | --- |
| `base64url_encode` | 62 |
| `base64url_decode` | 63 |
| `keccak256` | 64 |

### cost

| operator | base cost | per argument | per byte |
| -- | -- | -- | -- |
| `base64url_encode` | 40 | 130 | 2 |
| `base64url_decode` | 400 | - | 2 |
| `keccak256` | 50 | 160 | 2 |

As always, there's also a cost of 10 per byte that's allocated for the return value.

## activation

Since this is a softfork, the extension will only be available when the `ENABLE_BASE64` flag is set, which it will be after the softfork activates.

The operators will only be available for programs (`<prg>`) executing under `(softfork <cost> 1 <prg> <arg>)`

## hard fork

There's an option to hard-fork these operators out of the softfork-guard by specifying `ENABLE_BASE64_OPS_OUTSIDE_GUARD`.

## keccak-256

The test vectors are generated using the python wheel `from eth_hash.auto import keccak`.


src/base64_ops.rs
```diff
@@ -0,0 +1,96 @@
...
+use crate::reduction::{EvalErr, Reduction, Response};
+use crate::SExp;
+use base64ct::{Base64UrlUnpadded, Encoder, Encoding};
+
+const BASE64_ENCODE_BASE_COST: Cost = 40;
+const BASE64_COST_PER_ARG: Cost = 130;
+const BASE64_DECODE_BASE_COST: Cost = 400;
+const BASE64_COST_PER_BYTE: Cost = 3;
+
+const fn encoded_len(n: usize) -> Option<usize> {
```
Rigidity, line 15: What is this?

src/base64_ops.rs
```diff
@@ -0,0 +1,96 @@
...
+    while let Some((arg, rest)) = a.next(arg_list) {
+        arg_list = rest;
+        match a.atom(arg) {
+            Atom::Borrowed(b) => {
+                enc.encode(b)
+                    .map_err(|_| EvalErr(arg, "base64 (internal error)".to_string()))?;
+            }
+            Atom::U32(b, len) => {
+                enc.encode(&b[4 - len..])
+                    .map_err(|_| EvalErr(arg, "base64 (internal error)".to_string()))?;
```
Rigidity, line 64: Could this be factored out and the match only calculates the input?

src/base64_ops.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    let output_size =
+        encoded_len(input_size).ok_or(EvalErr(input, "base64 invalid input length".to_string()))?;
+
+    cost += output_size as Cost * MALLOC_COST_PER_BYTE;
+    check_cost(a, cost, max_cost)?;
+
+    let mut output: Vec<u8> = vec![0; output_size];
+    let mut enc = Encoder::<Base64UrlUnpadded>::new(&mut output[..])
+        .map_err(|_| EvalErr(input, "base64 (internal error)".to_string()))?;
```
Rigidity, line 52: I'm wondering if it makes sense to `impl From<base64ct::Error>` so you could use the `?` operator directly on all of these errors (ane change the error message later in one place)

fuzz/fuzz_targets/base64.rs
```diff
@@ -0,0 +1,25 @@
...
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    let mut a = Allocator::new();
+    let blob = a.new_atom(data).expect("failed to create atom");
+    let args = a
+        .new_pair(blob, NodePtr::NIL)
+        .expect("failed to create pair");
+    let Ok(Reduction(cost, node)) = op_base64url_encode(&mut a, args, 11000000000) else {
+        panic!("base64url_encode failed");
```
Rigidity, line 13: Might not be too relevant here but isn't this basically `.expect`?

fuzz/fuzz_targets/operators.rs
```diff
@@ -68,6 +70,9 @@ const FUNS: [Opf; 45] = [
     // Secp operators
     op_secp256k1_verify as Opf,
     op_secp256r1_verify as Opf,
+    // base64 operators
+    op_base64url_encode as Opf,
+    op_keccak256 as Opf,
```
Rigidity, line 75: Should base64 decode be here?

src/base64_ops.rs
arvidn: it computes the number of characters needed to base64 encode a buffer (without padding)

src/base64_ops.rs
arvidn: in the long run, I think we need to clean up the error type situation in `clvm_rs` to have an error type that can wrap other error categories. like in `chia_rs`. It seems a bit wrong to add an implicit conversion between these errors that assume it's an internal error. In the decode case, it's not. In an ideal world we would preserve the base64 error code in that case.

arvidn: implementing an implicit conversion won't work because we also need the `NodePtr` argument for `EvalErr`

Rigidity: Ah I missed that, good point

src/base64_ops.rs
```diff
@@ -0,0 +1,94 @@
...
+use base64ct::{Base64UrlUnpadded, Encoder, Encoding};
+
+const BASE64_ENCODE_BASE_COST: Cost = 40;
+const BASE64_COST_PER_ARG: Cost = 130;
+const BASE64_DECODE_BASE_COST: Cost = 400;
+const BASE64_COST_PER_BYTE: Cost = 3;
+
+// this computes the size of the resulting base64 encoded string, given a number
+// of input bytes. We pre-allocate the string.
+const fn encoded_len(n: usize) -> Option<usize> {
```
Rigidity, line 17: Does this need to be `const`?

src/keccak256_ops.rs
```diff
@@ -0,0 +1,32 @@
...
+
+const KECCAK256_BASE_COST: Cost = 50;
+const KECCAK256_COST_PER_ARG: Cost = 160;
+const KECCAK256_COST_PER_BYTE: Cost = 2;
+
+pub fn op_keccak256(a: &mut Allocator, mut input: NodePtr, max_cost: Cost) -> Response {
+    let mut cost = KECCAK256_BASE_COST;
+
+    let mut byte_count: usize = 0;
+    let mut hasher = Keccak256::new();
```
Rigidity, line 17: For base64 ops you do this in a separate pass, so that you can exit early if the operation is too expensive - is this worth doing here too?

Rigidity (approved):
Looks good to me (feel free to resolve these minor comments)

src/base64_ops.rs
arvidn: I don't think there are any use cases for it. I copied it from `base64ct` and it's `const` there. It's needed to pre-allocate the right size output buffer.

src/keccak256_ops.rs
arvidn: the main purpose of the first pass in the base64 functions is to establish the buffer size of the result, since the `base64ct` library only provides constant-time operations, we need to pre-allocate the buffer.

arvidn:
these features were cancelled.

PR #405, opened 2024-05-15
extend softfork test to cover the hard fork flag

arvidn:
This is to test the cases where we make the new operators available outside the softfork-guard.

We make sure they fail with "unknown operator" in mempool mode, and are treated as unknown in consensus (i.e. return `NIL`).

When the hard fork flag is set, we ensure they work as expected in both mempool and consensus mode.

### DIFF
--- a/src/run_program.rs
+++ b/src/run_program.rs
@@ -1330,53 +1330,75 @@ fn test_run_program() {
 #[cfg(test)]
 use rstest::rstest;
 
+// the test cases for this test consists of:
+// prg: the program to run inside the softfork guard
+// cost: the expected cost of the program (the test adds the apply-operator)
+// enabled: the softfork extension number that enables operator in prg
+// hard_fork_flag: the flag that enables the program to be run outside the guard
+// err: the expected error message, empty string means OK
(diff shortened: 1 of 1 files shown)


src/run_program.rs
```diff
@@ -1330,53 +1330,75 @@ fn test_run_program() {
...
+// the test cases for this test consists of:
+// prg: the program to run inside the softfork guard
+// cost: the expected cost of the program (the test adds the apply-operator)
+// enabled: the softfork extension number that enables operator in prg
+// hard_fork_flag: the flag that enables the program to be run outside the guard
+// err: the expected error message, empty string means OK
+// The test programs are carefully crafted such that they fail with "clvm raise"
+// when run in consensus mode and the operators are unknown. e.g. (coinid ...)
+// returns NIL in that case, which compares not equal to the coin ID, which
+// raises the exception.
```
arvidn, line 1342: this comment explain the tweaks to some test cases, namely: `case::coinid` and `case::modpow`.

Rigidity (approved):
A bit confusing to read at first, but makes sense now - lgtm

PR #414, opened 2024-05-23
CHIA-595: Keccak256 soft-fork

arvidn:
This is a stripped-down version of https://github.com/Chia-Network/clvm_rs/pull/403 just implementing the keccak operator. (i.e. not the base64 operators).

The main difference is that in this PR, the `keccak256` operator has opcode 62 (*not* 64)


arvidn:
cancelled

PR #430, opened 2024-07-10
CHIA-622 Fix openssl feature activation

jack60612:


Rigidity (approved):
looks good to me, @arvidn what do you think

src/sha2.rs
```diff
@@ -1,27 +1,34 @@
...
-#[cfg(openssl)]
+#[cfg(feature = "openssl")]
 use openssl;
 
-#[cfg(openssl)]
-#[derive(Clone)]
+#[cfg(not(feature = "openssl"))]
+use sha2::Digest;
+
+#[derive(Default, Clone)]
```
arvidn, line 7: is `Clone` important? It makes it easier to accidentally copy the context, where it's meant to be moved or passed by reference

arvidn (commented):
looks good! just one comment

src/sha2.rs
jack60612: I think Brandon told me to put that I'll ask him

Rigidity: `Clone` was already there before this PR - I just added `Default` and changed the order.

`sha2` and `openssl` already implement `Clone` (which is why it can be derived).

jack60612: i was wrong it was already there, i can remove it if you wish

arvidn (approved):
lgtm

PR #431, opened 2024-07-10
CHIA-622 Make target based on OS type

jack60612:
enable in config, also increase semver.


jack60612:
not sure if needed but wrote just in case

jack60612:
depends on #430

jack60612:
determined not wanted

PR #434, opened 2024-07-11
Clippy fixes

Rigidity:
Fixes clippy warnings. `wee_alloc` isn't a dependency and is 3 years old, and should not be used.


jack60612 (approved):
lgtm

PR #438, opened 2024-07-12
Improve modpow test generation

Rigidity:
Moves the `modpow` and `%` test generation into a Rust script rather than Python. I intend to move the other tests to Rust later as well.

This also improves the test coverage by including negative bases and moduli. Exponents remain positive (since it fails otherwise) and moduli must be non-zero to prevent division by zero.

Also varies the length of the bytes, as well as including more tests than previously.


tools/src/bin/generate-modpow-tests.rs
```diff
@@ -0,0 +1,85 @@
...
+        let mut exponent = BigInt::from_signed_bytes_be(&bytes(0, 32));
+        if exponent < BigInt::ZERO {
+            exponent = -exponent;
+        }
+
+        // Generate a random modulus, but ensure it's non-zero.
+        let mut modulus = BigInt::from_signed_bytes_be(&bytes(0, 32));
+        if modulus == BigInt::ZERO {
+            modulus += 1;
+        }
```
arvidn, line 37: you could `+ 1` unconditionally, and preserve the distribution

tools/src/bin/generate-modpow-tests.rs
```diff
@@ -0,0 +1,85 @@
...
+        "; This file was generated by tools/src/bin/generate-modpow-tests.rs\n\n".to_string();
+
+    for _ in 0..100 {
+        let base = BigInt::from_signed_bytes_be(&bytes(0, 32));
+
+        // Generate a random exponent, but ensure it's positive.
+        let mut exponent = BigInt::from_signed_bytes_be(&bytes(0, 32));
+        if exponent < BigInt::ZERO {
+            exponent = -exponent;
+        }
```
arvidn, line 31: isn't there a way to treat the bytes as unsigned to begin with?

tools/src/bin/generate-modpow-tests.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    fs::write("./op-tests/test-modpow.txt", tests).unwrap();
+}
+
+// Convert a `BigInt` to a CLVM atom.
+fn to_atom(num: BigInt) -> Vec<u8> {
+    let mut allocator = Allocator::new();
+    let ptr = allocator.new_number(num).unwrap();
+    let atom = allocator.atom(ptr);
```
arvidn, line 83: if all you're interested in is the length of the atom, there's a shortcut, `atom_len()`

arvidn (commented):
looks good, just a few minor comments

tools/src/bin/generate-modpow-tests.rs
Rigidity: Well technically this won't always work since it could be `-1`

tools/src/bin/generate-modpow-tests.rs
Rigidity: Thanks

tools/src/bin/generate-modpow-tests.rs
Rigidity: Changed it to `from_bytes_be(Sign::Plus, ...)`

arvidn (approved):
did you verify that the output is deterministic?

Rigidity:
It is

op-tests/test-modpow.txt
```diff
@@ -1,100 +1,100 @@
 ; This file was generated by tools/src/bin/generate-modpow-tests.rs
 
-modpow 9341829497586649593865894363733490755438507694201597 556606291647424564814177367572458554425074604463581598568935778451121529 4373238041963667894698452839741832172073 => 645485963992042551624174571461937584879 | 26775
```
arvidn: how come this changed, if the generation is deterministic?

arvidn (approved):

op-tests/test-modpow.txt
arvidn: right, you changed how the bytes are interpreted

Rigidity: I also switched out the RNG

PR #473, opened 2024-09-24
Back references

arvidn:
This PR makes two changes:

1. It simplifies the back-references CLVM deserializer using `let else` and `panic!()` to make it clear to the reader that it's always supposed to find pairs in those positions. The variables are also renamed to make it clearer what's happening.
2. It simplifies the tests by parameterizing them (with `rstest`) rather than having a series of function calls, one for each test case.

### DIFF
--- a/src/serde/de_br.rs
+++ b/src/serde/de_br.rs
@@ -47,12 +47,16 @@ pub fn node_from_stream_backrefs(
             }
             ParseOp::Cons => {
                 // cons
-                if let SExp::Pair(v1, v2) = allocator.sexp(values) {
-                    if let SExp::Pair(v3, v4) = allocator.sexp(v2) {
-                        let new_root = allocator.new_pair(v3, v1)?;
-                        values = allocator.new_pair(new_root, v4)?;
-                    }
-                }
+                // pop left and right values off of the "values" stack, then
+                // push the new pair onto it
+                let SExp::Pair(right, rest) = allocator.sexp(values) else {
+                    panic!("internal error");
+                };
+                let SExp::Pair(left, rest) = allocator.sexp(rest) else {
+                    panic!("internal error");
+                };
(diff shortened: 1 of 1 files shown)


src/serde/de_br.rs
```diff
@@ -47,12 +47,16 @@ pub fn node_from_stream_backrefs(
...
+                // pop left and right values off of the "values" stack, then
+                // push the new pair onto it
+                let SExp::Pair(right, rest) = allocator.sexp(values) else {
+                    panic!("internal error");
+                };
+                let SExp::Pair(left, rest) = allocator.sexp(rest) else {
+                    panic!("internal error");
+                };
+                let new_root = allocator.new_pair(left, right)?;
+                values = allocator.new_pair(new_root, rest)?;
```
richardkiss, line 59: IMO, it's probably not necessary to make this change, but it's also probably harmless. I suppose it gives new readers more insight about which code paths are taken, so it makes it more readable.

richardkiss (approved):
Seems good.

src/serde/de_br.rs
arvidn: yeah, this helped me understand what's going on. One thing that bothers me a little bit is that the stack itself, that we use during parsing, is allocated by the `Allocator` and so will stay allocated well after the parsing.
It's hard to do anything about it too, since the serialization format allows referencing the whole stack, not just the elements we've put on it.

richardkiss: A `copy_node(src_allocator, src_node, dst_allocator)` function could be used to garbage collect unused temporary nodes and could be generally useful. I've been looking into writing this for other reasons (a feature that allows users to define operators in python, which would require two allocators with a mapping between them) so it's an optional optimization. When selected, it would deserialize into a temporary allocator and then copy the top level object.

PR #489, opened 2024-10-28
Implement keccak256 operator and softfork

Rigidity:
Implements [CHIP-36](https://github.com/Chia-Network/chips/pull/131), using the implementation originally provided by @arvidn in https://github.com/Chia-Network/clvm_rs/pull/403 for [CHIP-34](https://github.com/Chia-Network/chips/pull/116).

## New operators

There is 1 new operator enabled under softfork `1`:

| operator | opcode |
| --- | --- |
| `keccak256` | 62 |

### Cost

| operator | base cost | per argument | per byte |
| -- | -- | -- | -- |
| `keccak256` | 50 | 160 | 2 |

As always, there's also a cost of 10 per byte that's allocated for the return value.

## Activation

Since this is a softfork, the extension will only be available when the `ENABLE_KECCAK` flag is set, which it will be after the softfork activates.

The operators will only be available for programs (`<prg>`) executing under `(softfork <cost> 1 <prg> <arg>)`

## Hard fork

There's an option to hard-fork these operators out of the softfork-guard by specifying `ENABLE_KECCAK_OPS_OUTSIDE_GUARD`.

## Tests
The test vectors are generated using the python wheel `from eth_hash.auto import keccak`.


src/run_program.rs
```diff
@@ -1165,16 +1165,67 @@ mod tests {
             err: "softfork specified cost mismatch",
         },
 
-        // without the flag to enable the BLS extensions, it's an unknown extension
+        // without the flag to enable the keccak extensions, it's an unknown extension
         RunProgramTest {
-            prg: "(softfork (q . 161) (q . 1) (q . (q . 42)) (q . ()))",
+            prg: "(softfork (q . 161) (q . 2) (q . (q . 42)) (q . ()))",
```
arvidn, line 1170: shouldn't `42` be `62` here?

Rigidity: I don't think this tests the operator, just an arbitrary body to the softfork op (it was here before keccak) @arvidn

arvidn (dismissed):

arvidn:
did you generate the fuzzing corpus and run the fuzzer for a while?

arvidn (approved):

arvidn (approved):

PR #491, opened 2024-11-04
Potential fix for softfork guard issue

Rigidity:
Moves keccak256 into softfork extension 2, to prevent conflicts with the old extension 1. It should be treated as a no-op until the fork activates, and the old extension should continue to work in mempool mode. This is a bug fix that allows that.

### DIFF
--- a/src/chia_dialect.rs
+++ b/src/chia_dialect.rs
@@ -72,6 +72,7 @@ impl Dialect for ChiaDialect {
     ) -> Response {
         let flags = self.flags
             | match extension {
+                OperatorSet::Unknown => 0,
                 OperatorSet::Default => 0,
                 OperatorSet::BLS => 0,
                 OperatorSet::Keccak => ENABLE_KECCAK_OPS_OUTSIDE_GUARD,
@@ -186,10 +187,11 @@ impl Dialect for ChiaDialect {
         match ext {
             // The BLS extensions (and coinid) operators were brought into the
             // main operator set as part of the hard fork
-            0 => OperatorSet::BLS,
+            0 => OperatorSet::Default,
             1 if (self.flags & ENABLE_KECCAK) != 0 => OperatorSet::Keccak,
+            1 => OperatorSet::BLS,
             // new extensions go here
-            _ => OperatorSet::Default,
+            _ => OperatorSet::Unknown,
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/chia_dialect.rs
```diff
@@ -186,10 +186,10 @@ impl Dialect for ChiaDialect {
         match ext {
             // The BLS extensions (and coinid) operators were brought into the
             // main operator set as part of the hard fork
-            0 => OperatorSet::BLS,
-            1 if (self.flags & ENABLE_KECCAK) != 0 => OperatorSet::Keccak,
+            0 | 1 => OperatorSet::Default,
```
arvidn, line 189: When I reviewed your first PR, I (for some reason) was thinking that the BLS softfork had id 0, so 1 was the next one. I realize I was wrong. Can you think of ways to prevent this mistake in the future?
There should probably be a comment here explaining what operator set 1 *is* (or was, prior to hard fork).
Maybe it even makes sense to preserve the `OperatorSet::BLS` enum value.

src/dialect.rs
```diff
@@ -6,8 +6,8 @@ use crate::reduction::Response;
 #[derive(Clone, Copy, Eq, PartialEq)]
 pub enum OperatorSet {
     Default,
-    BLS,
     Keccak, // keccak256 operator
+    Unknown,
```
arvidn, line 10: I think it would make sense to, briefly, describe the issue we just introduced and how `Default` and `Unknown` differ and solve the problem.

src/run_program.rs
```diff
@@ -336,7 +336,7 @@ impl<'a, D: Dialect> RunProgramContext<'a, D> {
         let extension =
             self.dialect
                 .softfork_extension(uint_atom::<4>(self.allocator, extension, "softfork")? as u32);
-        if extension == OperatorSet::Default {
+        if extension == OperatorSet::Unknown {
```
arvidn, line 339: this failure must depend on whether we're in mempool or not, right? Do we ignore the "unknown softfork extension" error we issue below, in the caller, if we're in consensus mode?

arvidn (commented):
My understanding is that a softfork guard with operator set 3 is either a failure (in mempool mode) or a no-op (in consensus mode). I don't see this logic. Is it just outside of the patch?
I would have expected this case to be handled in connection with `OperatorSet::Unknown`

PR #497, opened 2024-11-15
CHIA-1527: Update chia-bls and switch to chia-sha2

Rigidity:


arvidn (approved):
my understanding is that you preserve the `openssl` feature, but rather than including a dependency on `openssl` it in turn enables the `openssl` feature in the `chia-sha2` crate.

Rigidity:
> my understanding is that you preserve the `openssl` feature, but rather than including a dependency on `openssl` it in turn enables the `openssl` feature in the `chia-sha2` crate.

yes

PR #501, opened 2024-11-27
Enable collection of secp256k1_verify and secp256r1_verify operators

Rigidity:
With BLS signatures, you can simply run a puzzle with its solution to figure out all of the public key and message pairs that need to be signed, since they're included in the list of output conditions.

However, the secp256k1 and secp256r1 opcodes will fail at runtime if the signature passed into the solution doesn't match, instead of returning the expected values in conditions.

This PR enables a similar workflow by adding a new `run_program_with_signatures` function that skips and collects these operators and returns them in a list. Instead of failing the program, they will be treated as if they were successful in validating the signature, to allow the rest of the program to proceed. You can iterate the list of collected ops to find the public keys, messages, and signatures needed for signing.

The only problem with this is you need a way to update the original solution afterward to use the new signatures (and then you could run it for real to see if it actually worked). This is the responsibility of the code consuming this API, and can potentially be done in two ways:
1. Walk the solution as a tree of NodePtrs and compare the NodePtr directly via a HashMap. This assumes nothing has been done to change the NodePtr itself (quoting does not, for example).
2. Instead of nil and relying on the NodePtr matching, you could use a completely random fake signature and check for equality of the signature value in order to replace it with the real signature later.

I believe this PR will enable considerable simplification of wallet code, if my assumptions are correct.

### DIFF
--- a/src/collect_dialect.rs
+++ b/src/collect_dialect.rs
@@ -76,6 +76,9 @@ where
         let opcode = u32::from_be_bytes(atom.as_ref().try_into().unwrap());
 
         let cost = match opcode {
+            // We special case these opcodes and allow the response to pass through otherwise.
+            // If new operators are added to the main dialect, they likely shouldn't be included here.
+            // We're using the same cost to ensure that softfork conditions behave the same.
             0x13d61f00 => SECP256K1_VERIFY_COST,
             0x1c3a8f00 => SECP256R1_VERIFY_COST,
             _ => return response,
--- a/src/collect_dialect.rs
+++ b/src/collect_dialect.rs
@@ -1,0 +1,90 @@
+use std::cell::RefCell;
+
+use crate::{
+    cost::Cost,
+    dialect::{Dialect, OperatorSet},
+    reduction::{Reduction, Response},
+    secp_ops::{SECP256K1_VERIFY_COST, SECP256R1_VERIFY_COST},
+    Allocator, NodePtr,
+};
+
+#[derive(Debug, Clone, Copy)]
+pub struct CollectedOp {
+    pub op: NodePtr,
+    pub args: NodePtr,
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct CollectDialect<T> {
+    dialect: T,
+    collected_ops: RefCell<Vec<CollectedOp>>,
+}
+
+impl<T> CollectDialect<T> {
+    pub fn new(dialect: T) -> Self {
+        Self {
+            dialect,
+            collected_ops: RefCell::new(Vec::new()),
+        }
+    }
+
+    pub fn collect(self) -> Vec<CollectedOp> {
+        self.collected_ops.into_inner()
+    }
+}
+
+impl<T> Dialect for CollectDialect<T>
+where
+    T: Dialect,
+{
+    fn apply_kw(&self) -> u32 {
+        self.dialect.apply_kw()
+    }
+
+    fn quote_kw(&self) -> u32 {
+        self.dialect.quote_kw()
+    }
+
+    fn softfork_kw(&self) -> u32 {
+        self.dialect.softfork_kw()
+    }
+
+    fn allow_unknown_ops(&self) -> bool {
+        self.dialect.allow_unknown_ops()
+    }
+
+    fn softfork_extension(&self, ext: u32) -> OperatorSet {
+        self.dialect.softfork_extension(ext)
+    }
+
+    fn op(
+        &self,
+        allocator: &mut Allocator,
+        op: NodePtr,
+        args: NodePtr,
+        max_cost: Cost,
+        extensions: OperatorSet,
+    ) -> Response {
+        let response = self.dialect.op(allocator, op, args, max_cost, extensions);
+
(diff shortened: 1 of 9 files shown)


src/run_program.rs
```diff
@@ -99,11 +105,11 @@ struct RunProgramContext<'a, D> {
     softfork_stack: Vec<SoftforkGuard>,
     #[cfg(feature = "counters")]
     pub counters: Counters,
-
     #[cfg(feature = "pre-eval")]
     pre_eval: Option<PreEval>,
     #[cfg(feature = "pre-eval")]
     posteval_stack: Vec<Box<PostEval>>,
+    collected_ops: Option<Vec<CollectedOp>>,
```
richardkiss, line 112: I do have reservations about making more complex this code which is already a bit too complex and nasty. You might be able to implement this functionality with `pre_eval`, and you can surely do it by creating a new separate "wallet/debugging" dialect where you can override the opcodes you need.

richardkiss:
This is a great observation and there are some great ideas here.

src/run_program.rs
Rigidity: Good point, a new dialect might be cleaner

richardkiss: I've done a similar thing with the python api without mutable dialects [here](https://github.com/richardkiss/clvm_rs/commit/5945a5649e1591b8118646f9df9e52ce4f484f4d), but I guess `pyo3` is essentially `RefCell`-like objects everywhere. Might be worth trying it both ways: mutable dialects and with a `RefCell` to see what is easiest. `RefCell` is usually kind of a bad smell because if you make a mistake, you get a runtime error instead of a compile error. But I think in this case, it should be pretty easy to avoid a runtime error so `RefCell` might be the least intrusive without being *too* gross.

arvidn (commented):
I would prefer a solution (which I believe Richard suggested) where you make a new `Dialect` class that collects all the public keys and messages, and forwards everything to the default `ChiaDialect`.

I think that would make this feature the most contained.

Rigidity:
> I would prefer a solution (which I believe Richard suggested) where you make a new `Dialect` class that collects all the public keys and messages, and forwards everything to the default `ChiaDialect`.
> 
> 
> 
> I think that would make this feature the most contained.

Could even make it wrap a generic inner dialect

Rigidity:
This new implementation uses a custom Dialect and RefCell, since it requires no changes to the existing code. This begs the question of whether this should be part of the core CLVM crate or if it should just be part of the wallet code (or chia-wallet-sdk)? As is, it can be done already externally without any changes.

If we make `Dialect::op` take self mutably, it would probably be slightly more efficient/safe with the downside of requiring a breaking change to the API.

src/run_program.rs
```diff
@@ -1580,4 +1581,35 @@ mod tests {
 
         assert_eq!(result.unwrap().0, cost);
     }
+
+    #[test]
```
richardkiss, line 1585: Is it possible to move this to `collect_dialect`? It'd be nice to have no changes to `run_program.rs`

src/secp_ops.rs
```diff
@@ -7,8 +7,8 @@ use k256::ecdsa::{Signature as K1Signature, VerifyingKey as K1VerifyingKey};
 use p256::ecdsa::signature::hazmat::PrehashVerifier;
 use p256::ecdsa::{Signature as P1Signature, VerifyingKey as P1VerifyingKey};
 
-const SECP256R1_VERIFY_COST: Cost = 1850000;
-const SECP256K1_VERIFY_COST: Cost = 1300000;
+pub const SECP256R1_VERIFY_COST: Cost = 1850000;
```
richardkiss, line 10: Prob don't need this change anymore.

src/collect_dialect.rs
```diff
@@ -0,0 +1,90 @@
...
+        allocator: &mut Allocator,
+        op: NodePtr,
+        args: NodePtr,
+        max_cost: Cost,
+        extensions: OperatorSet,
+    ) -> Response {
+        let response = self.dialect.op(allocator, op, args, max_cost, extensions);
+
+        let op_len = allocator.atom_len(op);
+        if op_len != 4 {
```
richardkiss, line 71: This is a nice trick to restrict to these new opcode, but it deserves a comment at least since it may cause surprising things to happen if we add additional length four operators in future.

Rigidity: Yeah and frankly I should probably capture BLS opcodes too.

src/secp_ops.rs
Rigidity: I wonder if all of the costs should be defined in one place, like `src/costs.rs` and exported publicly, for consistency?

src/run_program.rs
Rigidity: Yeah I'll move it

src/secp_ops.rs
Rigidity: To avoid changing the public API on accident, I changed it to `pub(crate)` for now.

Rigidity:
This can be implemented externally to clvm_rs so barring a more general solution I'm going to leave this for now.

PR #511, opened 2024-12-13
ObjectCache doesn't need the Allocator as member

arvidn:
Passing it in on-demand is more flexible from a borrow-checker point of view.

I'm interested in using the `ObjectCache` in a scenario where I also keep building new structures in the `Allocator`. Currently, the `ObjectCache` holding a reference to it, won't allow anyone else to make any changes to the allocations.

### DIFF
--- a/src/serde/de_br.rs
+++ b/src/serde/de_br.rs
@@ -126,8 +126,8 @@ mod tests {
         let buf = Vec::from_hex(serialization_as_hex).unwrap();
         let mut allocator = Allocator::new();
         let node = node_from_bytes_backrefs(&mut allocator, &buf).unwrap();
-        let mut oc = ObjectCache::new(&allocator, treehash);
-        let calculated_hash = oc.get_or_calculate(&node).unwrap();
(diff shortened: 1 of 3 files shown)


Rigidity (approved):

richardkiss (approved):
This patch introduces a new footgun where a dumb user could pass in different allocators at different times and confused the heck out of everything.

Except for this caveat, this patch seems pretty reasonable and should allow you to mutate the allocator. Since an allocator can mutate "grow-only" using the standard API, it seems like this should be okay.

arvidn:
yeah, I was thinking about that. it seems overkill to, somehow, assert that the same allocator is passed in every time. It's also a key feature of the allocator, that `NodePtr` aren't actual pointers, and that you're already (somewhat) detached from the borrow-checker because of it.

richardkiss:
Yeah, you're right, we already have lots of places where the allocator and `NodePtr` are detached. So I guess this isn't really that new a category of footgun.

PR #517, opened 2024-12-20
Optimize `ReadCacheLookup`, 6.7 speed-up

arvidn:
This is a major hot-spot when serializing with back references. This patch speeds it up by:

* pre-allocating hash maps and hash sets
* using `IdentityHash`, to avoid hashing the tree-hashes again
* exiting early if the structure we're trying to deduplicate is too small (i.e. trade-off compression in favor of speed)
* use `BitVec` instead of `Vec<u8>` to store paths (which only contain 0 and 1 anyway).

Compared to `main`, the relevant benchmarks indicate a time reduction of between 80-90%, of serializing clvm with back references.

```
$ cargo bench node_to_bytes_backrefs -- --baseline main
     Running benches/serialize.rs (target/release/deps/serialize-21b0d813d80dbf8a)
Benchmarking serialize/node_to_bytes_backrefs 0: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 41.6s, or reduce sample count to 10.
serialize/node_to_bytes_backrefs 0
                        time:   [417.23 ms 420.23 ms 423.98 ms]
                        change: [-85.682% -85.544% -85.390%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  4 (4.00%) high mild
  7 (7.00%) high severe
Benchmarking serialize/node_to_bytes_backrefs 1: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 48.2s, or reduce sample count to 10.
serialize/node_to_bytes_backrefs 1
                        time:   [478.85 ms 481.87 ms 485.32 ms]
                        change: [-91.144% -91.077% -91.007%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 19 outliers among 100 measurements (19.00%)
  8 (8.00%) high mild
  11 (11.00%) high severe
```

One concern is that, not mixing in any random state in the hasher, opens up for hash-collision attacks. I'm not sure if it's worth addressing or what the best way would be. Maybe add a random value to the `IdentityHash` object, and add it to the hash value.


src/serde/read_cache_lookup.rs
```diff
@@ -28,10 +56,10 @@ pub struct ReadCacheLookup {
     /// the tree hashes of the contents on the left and right
     read_stack: Vec<(Bytes32, Bytes32)>,
 
-    count: HashMap<Bytes32, u32>,
+    count: HashMap<Bytes32, u32, IdentityHash>,
 
     /// a mapping of tree hashes to `(parent, is_right)` tuples
-    parent_lookup: HashMap<Bytes32, Vec<(Bytes32, u8)>>,
+    parent_lookup: HashMap<Bytes32, Vec<(Bytes32, bool)>, IdentityHash>,
```
richardkiss, line 62: I'd be surprised to see that this change makes any difference.

src/serde/read_cache_lookup.rs
```diff
@@ -16,9 +18,35 @@
...
+    fn finish(&self) -> u64 {
+        self.0
+    }
+
+    fn write(&mut self, bytes: &[u8]) {
+        self.0 = u64::from_le_bytes(bytes[0..8].try_into().expect("expected 32 byte hashes"));
+    }
+
+    fn write_u64(&mut self, _i: u64) {
+        panic!("This hasher only takes bytes");
```
richardkiss, line 38: I don't know if this is necessary. If you want to use a standalone hasher I tried a couple years back, you could steal it from https://github.com/richardkiss/clvm_rs/commit/50f6c94a02b909d28c5479453ff2d9d8e2800749#diff-b696b12532e34a5d93b4bf96927a9049a2ad3d7c04bfde786611f6a560622e2f ; I don't think it'd be much slower, if at all.

src/serde/read_cache_lookup.rs
richardkiss: I wonder if you could benchmark with & without this change. If it doesn't improve anything, I think it reads nicer with 0s and 1s than trues and falses.

src/serde/read_cache_lookup.rs
```diff
@@ -147,11 +187,11 @@ impl ReadCacheLookup {
                     for (parent, direction) in items.iter() {
                         if *(self.count.get(parent).unwrap_or(&0)) > 0 && !seen_ids.contains(parent)
                         {
-                            let mut new_path = path.clone();
-                            new_path.push(*direction);
-                            if new_path.len() > max_path_length {
+                            if path.len() + 1 > max_path_length {
```
richardkiss, line 190: how about using `>=` rather then `+1`?

richardkiss: And we're sure this doesn't change anything? Oh I see, the `new_path` stuff is deferred until after this line. Okay, seems good.

src/serde/read_cache_lookup.rs
```diff
@@ -185,13 +225,13 @@ impl ReadCacheLookup {
 /// If `A` => `v` then `[A] + [0]` => `v * 2` and `[A] + [1]` => `v * 2 + 1`
 /// Then the integer is turned into the minimal-length array of `u8` representing
 /// that value as an unsigned integer.
-fn reversed_path_to_vec_u8(path: &[u8]) -> Vec<u8> {
+fn reversed_path_to_vec_u8(path: &BitSlice) -> Vec<u8> {
```
richardkiss, line 228: I'm user switching to bitvec instead of u8 uses less memory, but I don't really know if it would be faster. On the one hand, we get more data in cache; on the other hand, now we're doing a lot of masking. I guess it comes down to if we ever actually blow out the cache, which depends a lot on how big the cpu cache is, which varies a lot.

So I suppose this change is probably never negative.

richardkiss (approved):
Looks good. I'd be curious to know if the `u8` -> `bool` change does anything. But it seems at worst harmless.

src/serde/read_cache_lookup.rs
arvidn: The `IdentityHash` makes a difference. The CPU cost of this function was dominated by the hashing for the `HashSet` and `HashMap`s. (after this patch, it's dominated by growing the `HashMap`.

The `u8` -> `bool` change was because `BitVec` uses `bool`. It emulates the interface of `Vec<bool>`.

src/serde/read_cache_lookup.rs
arvidn: this "hasher" depends on the input already having hash-properties. If anyone, for some reason, puts a `u64` into it, that assumption likely no longer holds. I think it's preferable to panic in that case.

That said, the one risk I see with this is that there's no salt, so an attacker can grind hashes that causes worst-case performance of the hash set/map.

src/serde/read_cache_lookup.rs
arvidn: yeah, I don't think this makes a huge difference. I don't think it happens very often, I was just on the look-out for avoiding heap allocations.

arvidn: I think I like `+ 1` a bit better, since we're checking the state *as-if* we had added the new path

src/serde/read_cache_lookup.rs
arvidn: I don't think this made a significant difference either. This was, again, while looking to minimize and avoid heap allocations. I think there's a possibility that this causes fewer allocations, as it might not grow as many times as a `Vec<u8>`

src/serde/read_cache_lookup.rs
richardkiss: Possible solution; don't always take the first eight bytes, but randomly select an offset between 0 and 24 to start taking the bytes.

True, that only makes it 24 times as hard. But we could choose, say, five random offsets between 0 and 24 and eor them together. Then it's 24**5 times as hard.

richardkiss: One attack I see would be, create a legit spend that has so many hash tree collisions that the hash tree gets O(n^2) performance. If we only have a 1/24 chance of having that happen, it won't be long before it gets included in a block by someone who chooses a failing offset, and then the spend is gone forever.

That's really the only attack I can think of.

arvidn:
There are still some worst-case scenarios where this isn't fast enough to serialize a full block. I'm working on a more ambitions PR to speed this up more.

arvidn:
a material portion of the speed-up in this patch comes from not using a salt in the hasher. Which is a bit surprising, but I suppose we initialize new `HashSet` and `HashMaps` quite often. adding back the salt makes the benchmarks take 3x the time

src/serde/read_cache_lookup.rs
arvidn: I've addressed this now

richardkiss: I'm kind of averse to using random state because it makes things no longer be "pure" functions without side effects/global state, which means failing test cases can come and go. That said, it's not really a problem here since even if our random function is terrible, sha256 is going to more than make up for it.

arvidn: this mimics the standard library's `RandomState`, since it's also salting its hash. The state isn't global exactly, it's just per `HashMap`/`HashSet` instance. I don't believe there are any workable alternatives to salting the hash.

richardkiss (approved):
Looks good

PR #530, opened 2025-01-07
fix Serializer to correctly restore()

arvidn:
to allow resumption of serialization.

It turns out my original patch wasn't tested quite well enough. This adds a test for the important case of restoring the `Serializer` to an earlier state and still be valid to support serialize more CLVM structures.

A somewhat unrelated change is to pass in the sentinel node to the `Serializer` constructor, rather than at every call to `add()`.

### DIFF
--- a/src/serde/incremental.rs
+++ b/src/serde/incremental.rs
@@ -27,19 +27,15 @@ pub struct Serializer {
     thc: ObjectCache<Bytes32>,
     slc: ObjectCache<u64>,
 
+    sentinel: Option<NodePtr>,
     output: Cursor<Vec<u8>>,
-}
-
-impl Default for Serializer {
-    fn default() -> Self {
-        Self::new()
-    }
 }
 
 #[derive(Clone)]
 pub struct UndoState {
     read_op_stack: Vec<ReadOp>,
     write_stack: Vec<NodePtr>,
+    read_cache_lookup: ReadCacheLookup,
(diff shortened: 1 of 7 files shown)


src/serde/incremental.rs
```diff
@@ -315,17 +308,17 @@ mod tests {
         assert_eq!(ser.size(), 8);
         assert_eq!(hex::encode(ser.get_ref()), "ffffff0102ff0304");
 
-        let (done, _) = ser.add(&a, item, None).unwrap();
+        let (done, _) = ser.add(&a, item).unwrap();
         assert!(done);
 
         assert_eq!(ser.size(), 10);
-        assert_eq!(hex::encode(ser.get_ref()), "ffffff0102ff0304fe04");
```
arvidn, line 322: this was one way this bug was exposed, but I had missed it

richardkiss (approved):
Near as I can tell, you've modified it so the `sentinel` node is now a property of the instance rather than being passed in to `.add`. I don't completely understand all the other changes made or the genesis of the bug, but I expect you have a pretty good handle on this.

richardkiss (commented):
Ah, you're now copying the `ReadCacheLookup` to the undo state. Seems reasonable.

arvidn:
yeah, exactly. the `ReadCacheLookup` is recording the current state of the parse stack. so it has to be restored as well

PR #532, opened 2025-01-13
Node from stream backrefs optimisation

matt-o-how:
Use a `Vec<NodePtr>` stack instead of `NodePtr` / `SExp`s in `node_from_stream_backrefs` and add a new `traverse_path_with_vec()` function to handle backrefs


arvidn (commented):
it looks correct, as far as I can tell. I think we need tests for all interesting cases, to make sure it works. I'm also interested in seeing a benchmark. Does this make a difference? I would expect it to at least use less memory, which typically means faster on small machines (like Raspberry PI)

src/traverse_path.rs
```diff
@@ -72,6 +72,83 @@ pub fn traverse_path(allocator: &Allocator, node_index: &[u8], args: NodePtr) ->
...
+    args: &[NodePtr],
+) -> Response {
+    // the vec is a stack so a ChiaLisp list of (3 . (2 . (1 . NIL))) would be [1, 2, 3]
+    // however entries in this vec may be ChiaLisp SExps so it may look more like [1, (2 . NIL), 3]
+    let mut arg_list: Vec<NodePtr> = args.to_owned();
+
+    // find first non-zero byte
+    let first_bit_byte_index = first_non_zero(node_index);
+
+    let mut cost: Cost = TRAVERSE_BASE_COST
```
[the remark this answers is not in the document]
matt-o-how: Done

src/traverse_path.rs
```diff
@@ -72,6 +72,83 @@ pub fn traverse_path(allocator: &Allocator, node_index: &[u8], args: NodePtr) ->
...
 }
 
+pub fn traverse_path_with_vec(
+    allocator: &mut Allocator,
+    node_index: &[u8],
+    args: &[NodePtr],
+) -> Response {
+    // the vec is a stack so a ChiaLisp list of (3 . (2 . (1 . NIL))) would be [1, 2, 3]
+    // however entries in this vec may be ChiaLisp SExps so it may look more like [1, (2 . NIL), 3]
+    let mut arg_list: Vec<NodePtr> = args.to_owned();
```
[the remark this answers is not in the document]
matt-o-how: Done

src/traverse_path.rs
```diff
@@ -72,6 +72,83 @@ pub fn traverse_path(allocator: &Allocator, node_index: &[u8], args: NodePtr) ->
...
+        } else {
+            bitmask <<= 1;
+        }
+        cost += TRAVERSE_COST_PER_BIT;
+    }
+    if parsing_sexp {
+        return Ok(Reduction(cost, sexp_to_parse));
+    }
+    if arg_list.is_empty() {
+        return Ok(Reduction(cost, allocator.nil()));
```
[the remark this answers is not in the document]
matt-o-how: Replaced all instances of `allocator.nil()` and removed this case as I think it was redundant

arvidn (commented):
I think these things are still needed:

- [x] preserve the existing function, partly to control when we switch over to the new one, and also to be able to test that both behave the same
- [x] ensure the new function produce the same result as the old one, e.g. with a fuzzer.
- [x] ensure the new function behave the same with regards to limits to the number of pairs created by `Allocator`. It can be tested in a fuzzer by building with the `counters` build feature
- [x] benchmark to demonstrate that this is an improvement (this should probably be done early, as we might want to scrap this idea if it doesn't carry its weight)
- [ ] survey the mainnet and testnet blockchains to see if back references into the parse-stack eveer exists in the wild
- [x] unit tests for all edge cases

src/traverse_path.rs
```diff
@@ -72,6 +72,83 @@ pub fn traverse_path(allocator: &Allocator, node_index: &[u8], args: NodePtr) ->
...
+
+    // find first non-zero bit (the most significant bit is a sentinel)
+    let last_bitmask = msb_mask(node_index[first_bit_byte_index]);
+
+    // follow through the bits, moving left and right
+    let mut byte_idx = node_index.len() - 1;
+    let mut bitmask = 0x01;
+
+    // if we move from parsing the Vec stack to parsing the SExp stack use the following variables
+    let mut parsing_sexp = false;
```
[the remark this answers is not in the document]
matt-o-how: Given the change I had to make for parsing in the case of the stack being empty I don't believe we should make this change

src/serde/de_br.rs
```diff
@@ -83,6 +129,86 @@ pub fn node_from_bytes_backrefs_record(
     Ok((ret, backrefs))
 }
 
+pub fn traverse_path_with_vec(
```
arvidn, line 132: this is a pretty big function for not having a unit test.

.github/workflows/build-test.yml
```diff
@@ -212,7 +212,7 @@ jobs:
           cd tools
           cargo run --bin generate-fuzz-corpus
       - name: build
-        run: cargo fuzz list | xargs -I "%" sh -c "cargo fuzz run % -- -max_total_time=30 || exit 255"
+        run: cargo fuzz list | xargs -I "%" sh -c "cargo fuzz run % -- -max_total_time=30 || exit 255  --all-features"
```
arvidn, line 215: If you make the fuzzer depend on `clvmr/counters` I don't think you would need this

fuzz/Cargo.toml
```diff
@@ -8,6 +8,11 @@ edition = "2021"
...
 cargo-fuzz = true
 
+[features]
+# when enabling the "counters" features, the CLVM interpreter is instrumented to
+# collect counters about the programs it executes
+counters = ["clvmr/counters"]
+
 [dependencies]
 libfuzzer-sys = { workspace = true }
 clvmr = { workspace = true }
```
arvidn, line 18: I would expect you can add `features = ["counters"]` here, to always depend on `clvmr` with `counters` emabled

fuzz/fuzz_targets/deserialize_br.rs
```diff
@@ -17,7 +18,31 @@ fuzz_target!(|data: &[u8]| {
...
+fuzz_target!(|data: &[u8]| {
+    let mut allocator = Allocator::new();
+    let program = match node_from_bytes_backrefs(&mut allocator, data) {
+        Err(_) => {
+            return;
+        }
+        Ok(r) => r,
+    };
+
+    let b1 = node_to_bytes_backrefs(&allocator, program).unwrap();
```
arvidn, line 38: I don't think you need this step. Also, you may loose interesting cases by "filtering" the tree when you re-encode it. It won't necessarily round-trip.
Is there a good reason to do this rather than using `data` directly?

fuzz/fuzz_targets/deserialize_br.rs
```diff
@@ -17,7 +18,31 @@ fuzz_target!(|data: &[u8]| {
...
+
+    // reset allocators
+    let mut allocator = Allocator::new();
+    let mut allocator_old = Allocator::new();
+
+    let mut allocator = Allocator::new();
```
arvidn, line 44: this line looks like a mistake

fuzz/fuzz_targets/deserialize_br.rs
```diff
@@ -17,7 +18,31 @@ fuzz_target!(|data: &[u8]| {
...
+            return;
+        }
+        Ok(r) => r,
+    };
+
+    let b1 = node_to_bytes_backrefs(&allocator, program).unwrap();
+
+    // reset allocators
+    let mut allocator = Allocator::new();
+    let mut allocator_old = Allocator::new();
```
arvidn, line 42: this fuzzer will be faster (and more efficient) if you use a single allocator that you reset (via `checkpoint()` and `restore_checkpoint()`). That way the memory allocations only need to be made once and can be reused.

src/traverse_path.rs
```diff
@@ -13,7 +13,7 @@ const TRAVERSE_COST_PER_BIT: Cost = 4;
 
 // return a bitmask with a single bit set, for the most significant set bit in
 // the input byte
-fn msb_mask(byte: u8) -> u8 {
+pub fn msb_mask(byte: u8) -> u8 {
```
arvidn, line 16: I'm not sure, but I think you can say `pub(crate)` to avoid exporting them. If so, it would make sense for these functions

fuzz/fuzz_targets/deserialize_br.rs
```diff
@@ -17,7 +18,31 @@ fuzz_target!(|data: &[u8]| {
...
+
+    let mut allocator = Allocator::new();
+    let program = node_from_bytes_backrefs(&mut allocator, &b1).unwrap();
+    let program_old = node_from_bytes_backrefs_old(&mut allocator_old, &b1).unwrap();
+    assert!(allocator.pair_count() <= allocator_old.pair_count());
```
arvidn, line 47: Another important test (that I believe would fail right now) is to deserialize `bytes` again, using both old and new functions and an allocator that has a pair-limit of `allocator_old.pair_count() - 1`. i.e. it should run out of memory, and so should the new function. Otherwise consensus rules will have changed, and we have a potential for a chain split.

I believe the simplest way to implement this is to have a function on `Allocator` to create a "fake" pair. I just needs to decrement its limit. It also means `checkpoint()` and `restore_checkpoint()` would need to track and update this limit.

src/serde/de_br.rs
```diff
@@ -22,7 +22,7 @@ pub fn node_from_stream_backrefs(
     f: &mut Cursor<&[u8]>,
     mut backref_callback: impl FnMut(NodePtr),
 ) -> io::Result<NodePtr> {
-    let mut values = allocator.nil();
+    let mut values = Vec::<NodePtr>::new();
```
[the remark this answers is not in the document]
arvidn: I think this is required to avoid consensus problems. consider the worst case scenario where you have a tree that's then referenced thousands of times. With todays deserializer, they will point to the same tree, but in your version, you'll create the same pairs over and over again. In fact, I would expect the fuzzer to fail if you let it run long enough because of this.

src/serde/de_br.rs
```diff
@@ -83,6 +129,86 @@ pub fn node_from_bytes_backrefs_record(
...
+
+    // if we move from parsing the Vec stack to parsing the SExp stack use the following variables
+    let mut sexp_to_parse = NodePtr::NIL;
+
+    while byte_idx > first_bit_byte_index || bitmask < last_bitmask {
+        let is_bit_set: bool = (node_index[byte_idx] & bitmask) != 0;
+        if parsing_sexp {
+            match allocator.sexp(sexp_to_parse) {
+                SExp::Atom => {
+                    return Err(EvalErr(sexp_to_parse, "path into atom".into()).into());
```
arvidn, line 169: we could make this error message better now I think. We know it's a backref and that the serialization is invalid

src/serde/de_br.rs
```diff
@@ -83,6 +129,86 @@ pub fn node_from_bytes_backrefs_record(
...
+    node_index: &[u8],
+    args: &[NodePtr],
+) -> io::Result<NodePtr> {
+    // the vec is a stack so a ChiaLisp list of (3 . (2 . (1 . NIL))) would be [1, 2, 3]
+    // however entries in this vec may be ChiaLisp SExps so it may look more like [1, (2 . NIL), 3]
+
+    let mut parsing_sexp = false;
+    if args.is_empty() {
+        parsing_sexp = true;
+    }
```
arvidn, line 143:
```suggestion
    let mut parsing_sexp = args.is_empty();
```

src/serde/de_br.rs
```diff
@@ -83,6 +129,86 @@ pub fn node_from_bytes_backrefs_record(
...
+    let last_bitmask = msb_mask(node_index[first_bit_byte_index]);
+
+    // follow through the bits, moving left and right
+    let mut byte_idx = node_index.len() - 1;
+    let mut bitmask = 0x01;
+
+    // if we move from parsing the Vec stack to parsing the SExp stack use the following variables
+    let mut sexp_to_parse = NodePtr::NIL;
+
+    while byte_idx > first_bit_byte_index || bitmask < last_bitmask {
```
arvidn, line 164: did you explore making this two loops, one for the `Vec` stack and one for traversing the `SExp` tree?
I think it would be easier to follow. partly because you wouldn't need two loop bodies separated by the `parsing_sexp` check. It would also place the tail of this function in between those loops. If we exit after having pointed to a stack node

fuzz/fuzz_targets/deserialize_br.rs
arvidn: I think it would be great to also ensure that the deseialization yields the same tree. You can use `node_eq` (see other fuzzers). This precludes using a single `Allocator` though, as I suggested earlier.

src/serde/de_br.rs
arvidn: But I also think this function would likely be simpler without this variable

fuzz/fuzz_targets/deserialize_br.rs
matt-o-how: fixed

src/traverse_path.rs
matt-o-how: done

fuzz/Cargo.toml
matt-o-how: Done

src/serde/de_br.rs
matt-o-how: Added a unit test

src/serde/de_br.rs
matt-o-how: Changed error message to: "invalid backreference during deserialisation"

src/serde/de_br.rs
```diff
@@ -21,6 +21,47 @@ pub fn node_from_stream_backrefs(
...
+                }
+            }
+            ParseOp::Cons => {
+                // cons
+                // pop left and right values off of the "values" stack, then
+                // push the new pair onto it
+                let right = values.pop().expect("No cons without two vals.");
+                let left = values.pop().expect("No cons without two vals.");
+                let root_node = allocator.new_pair(left.0, right.0)?;
+                values.push((root_node, None));
```
arvidn, line 54: the allocator must be informed, somehow, that we're pretending to create a pair here, for the purpose of the allocator limits. One option could be to decrement the limit by one. Then, if `traverse_path_with_vec()` has to create some of the pairs to make up the stack, you would have to increment the limit again, for each pair created.

This is important in order to not change consensus rules between the old and the new function.

src/serde/de_br.rs
```diff
@@ -126,11 +265,69 @@ mod tests {
...
+            nul
+        );
+        assert_eq!(
+            traverse_path_with_vec(&mut a, &[0, 0, 0], &mut list).unwrap(),
+            nul
+        );
+        assert_eq!(
+            traverse_path_with_vec(&mut a, &[0, 0, 0, 0], &mut list).unwrap(),
+            nul
+        );
```
arvidn, line 308: a path with all zero-bits are invalid, aren't they? How come this isn't failing?

src/serde/de_br.rs
```diff
@@ -126,11 +265,69 @@ mod tests {
...
+        let mut a = Allocator::new();
+        let nul = a.nil();
+        let n1 = a.new_atom(&[0, 1, 2]).unwrap();
+        let n2 = a.new_atom(&[4, 5, 6]).unwrap();
+
+        let mut list: Vec<(NodePtr, Option<NodePtr>)> = vec![(n1, None)];
+
+        assert_eq!(traverse_path_with_vec(&mut a, &[], &mut list).unwrap(), nul);
+
+        // cost for leading zeros
```
arvidn, line 292: we don't account for cost in this function

src/serde/de_br.rs
```diff
@@ -126,11 +265,69 @@ mod tests {
...
+            n2
+        );
+        assert_eq!(
+            traverse_path_with_vec(&mut a, &[0b101], &mut list).unwrap(),
+            n1
+        );
+        assert_eq!(
+            traverse_path_with_vec(&mut a, &[0b111], &mut list).unwrap(),
+            nul
+        );
```
arvidn, line 323: I think you should have cases where the function creates new pairs as well. and probably also where it reuses the pairs already created by an earlier call.

And also that the pretend-pairs are counted against the allocator limits. You could create an allocator with a really low limit and make sure it's hit

src/serde/de_br.rs
```diff
@@ -83,8 +129,95 @@ pub fn node_from_bytes_backrefs_record(
...
+
+    let mut parsing_sexp = args.is_empty();
+
+    // instead of popping, we treat this as a pointer to the end of the virtual stack
+    let mut arg_index: usize = if parsing_sexp { 0 } else { args.len() - 1 };
+
+    // find first non-zero byte
+    let first_bit_byte_index = first_non_zero(node_index);
+    if first_bit_byte_index >= node_index.len() {
+        return Ok(NodePtr::NIL);
```
arvidn, line 148: I think this is an error. A path with no terminator bit is not valid

src/serde/de_br.rs
```diff
@@ -83,8 +129,95 @@ pub fn node_from_bytes_backrefs_record(
...
+    } else if args[arg_index].1.is_some() {
+        // cached entry exists - use that instead of recreating the pairs
+        return Ok(args[arg_index].1.unwrap());
+    }
+    // take bottom of stack and make (item . NIL)
+    let mut backref_node = allocator.new_pair(args[0].0, NodePtr::NIL)?;
+    if arg_index == 0 {
+        args[arg_index].1 = Some(backref_node);
+        return Ok(backref_node);
+    }
```
arvidn, line 209: it looks like this could be folded into the loop below. you just have to initialize `backref_node` to `NodePtr::NIL`

src/serde/de_br.rs
matt-o-how: Unsure, but this is parity with the old behaviour. 
See: the tests in `src\traverse_path.rs` line 157 on which this test was adapted from which also behaves like this

src/serde/de_br.rs
matt-o-how: Again, this is parity with the old traverse_path behaviour in `src\traverse_path.rs` , line 44

src/serde/de_br.rs
matt-o-how: Done

fuzz/fuzz_targets/deserialize_br.rs
```diff
@@ -15,9 +16,37 @@ fuzz_target!(|data: &[u8]| {
...
     let program = node_from_bytes_backrefs(&mut allocator, &b1).unwrap();
-
+    // let new_pair_count = allocator.pair_count();
+    // allocator.restore_checkpoint(&cp);
+    let program_old = node_from_bytes_backrefs_old(&mut allocator, &b1).unwrap();
+    // assert!(new_pair_count <= allocator.pair_count());
+    let b3 = node_to_bytes_backrefs(&allocator, program_old).unwrap();
+    assert_eq!(b1, b3);
     let b2 = node_to_bytes_backrefs(&allocator, program).unwrap();
     assert_eq!(b1, b2);
```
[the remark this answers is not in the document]
matt-o-how: changed this to do a NOT XOR of `program.is_err()` and `program_old.is_err()` and then `assert!(node_eq(` between them

src/serde/de_br.rs
matt-o-how: changed comment

src/allocator.rs
```diff
@@ -332,13 +336,36 @@ impl Allocator {
...
+        }
+        self.max_num_pairs -= amount;
+        Ok(())
+    }
+
+    pub fn increase_pair_max(&mut self, amount: usize) -> Result<(), EvalErr> {
+        // increase the max number of pairs by the amount, but don't exceed usize::MAX
+        let max_increase = usize::MAX - self.max_num_pairs;
+        let increase = amount.min(max_increase);
+        self.max_num_pairs += increase;
```
arvidn, line 365:
```suggestion
        self.max_num_pairs = u32::.saturating_add(self.max_num_paurs, increase);
```

src/allocator.rs
```diff
@@ -332,13 +336,36 @@ impl Allocator {
...
 
+    pub fn reduce_pair_max(&mut self, amount: usize) -> Result<(), EvalErr> {
+        if self.max_num_pairs - amount <= self.pair_vec.len() {
+            return err(
+                self.nil(),
+                format!(
+                    "reduce_pair_max: amount exceeds current pair count ({} pairs allocated, {} max pairs)",
+                    self.pair_vec.len(),
+                    self.max_num_pairs
+                ).as_str(),
```
arvidn, line 354: I think it would be nice to have the same error message as `new_pair()` here. a string literal "too many pairs".

src/allocator.rs
```diff
@@ -332,13 +336,36 @@ impl Allocator {
...
-        if idx == MAX_NUM_PAIRS {
+        if idx >= self.max_num_pairs {
             return err(self.nil(), "too many pairs");
         }
         self.pair_vec.push(IntPair { first, rest });
         Ok(NodePtr::new(ObjectType::Pair, idx))
     }
 
+    pub fn reduce_pair_max(&mut self, amount: usize) -> Result<(), EvalErr> {
+        if self.max_num_pairs - amount <= self.pair_vec.len() {
```
arvidn, line 347: I think this risks wrapping, doesn't it? How about this instead:
```suggestion
        if self.max_num_pairs - self.pair_vec.len() < amount {
```

src/allocator.rs
```diff
@@ -332,13 +336,36 @@ impl Allocator {
 
     pub fn new_pair(&mut self, first: NodePtr, rest: NodePtr) -> Result<NodePtr, EvalErr> {
         let idx = self.pair_vec.len();
-        if idx == MAX_NUM_PAIRS {
+        if idx >= self.max_num_pairs {
```
arvidn, line 339: note that we're checking this *before* we add the new pair. i.e. `idx == max_num_pairs` is a valid state. It's just the last valid, we can't add another pair after this.

src/allocator.rs
arvidn: if `max_num_pairs == pair_vec.len()`, that's fine. No failure. See comment above

src/serde/de_br.rs
arvidn: you need this call here too:
```
allocator.reduce_pair_max(1)?;
```

src/serde/de_br.rs
arvidn: I see. I'm a bit surprised

src/serde/de_br.rs
```diff
@@ -83,8 +133,103 @@ pub fn node_from_bytes_backrefs_record(
...
+
+    if parsing_sexp {
+        return Ok(sexp_to_parse);
+    } else if args[arg_index].1.is_some() {
+        // cached entry exists - use that instead of recreating the pairs
+        return Ok(args[arg_index].1.unwrap());
+    }
+    // take bottom of stack and make (item . NIL)
+    let mut backref_node = allocator.new_pair(args[0].0, NodePtr::NIL)?;
```
arvidn, line 210: you don't call `increase_pair_max(1)?;` here

arvidn: it also looks like you fail to store this pair in the cache, in the `Option<NodePtr>` slot

matt-o-how: fixed, and also fetch cached NIL if it exists

src/allocator.rs
matt-o-how: done with `self.max_num_pairs.saturating_add(amount)`

src/serde/de_br.rs
matt-o-how: Done

src/serde/de_br.rs
```diff
@@ -83,8 +135,113 @@ pub fn node_from_bytes_backrefs_record(
...
+        }
+    };
+
+    if arg_index == 0 {
+        args[arg_index].1 = Some(backref_node);
+        return Ok(backref_node);
+    }
+
+    // for the rest of items starting from last + 1 in stack
+    for x in args.iter_mut().take(arg_index + 1).skip(1) {
```
arvidn, line 229: I don't think you need to treat the bottom element of the stack as a special case. If you start with `backref_node = NodePtr::NIL` you can just loop over `0..arg_index` and create pairs and update the cache, if needed.

src/serde/de_br.rs
```diff
@@ -83,8 +135,113 @@ pub fn node_from_bytes_backrefs_record(
...
+        if let Some(pair) = x.1 {
+            backref_node = pair;
+            continue;
+        }
+        allocator.increase_pair_max(1)?;
+        backref_node = allocator.new_pair(x.0, backref_node)?;
+        x.1 = Some(backref_node);
+    }
+    args[arg_index].1 = Some(backref_node);
```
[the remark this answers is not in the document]
matt-o-how: the line isn't wrong, but it is now redundant given the loop before it - I forgot to remove it when I added caching at every step

src/serde/de_br.rs
```diff
@@ -83,8 +135,113 @@ pub fn node_from_bytes_backrefs_record(
...
+            bitmask <<= 1;
+        }
+    }
+
+    if parsing_sexp {
+        return Ok(sexp_to_parse);
+    } else if args[arg_index].1.is_some() {
+        // cached entry exists - use that instead of recreating the pairs
+        return Ok(args[arg_index].1.unwrap());
+    }
```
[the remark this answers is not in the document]
matt-o-how: Yes, the idea was to add an O(1) lookup to save the O(N) loop, but the result will be the same if I remove this. 

I'll comment it out for now.

src/serde/de_br.rs
matt-o-how: Done

src/serde/de_br.rs
```diff
@@ -83,8 +135,113 @@ pub fn node_from_bytes_backrefs_record(
...
+            allocator.increase_pair_max(1)?;
+            args[0].1 = Some(node);
+            node
+        }
+    };
+
+    if arg_index == 0 {
+        args[arg_index].1 = Some(backref_node);
+        return Ok(backref_node);
+    }
```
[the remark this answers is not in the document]
matt-o-how: Done

fuzz/fuzz_targets/deserialize_br.rs
```diff
@@ -15,9 +19,13 @@ fuzz_target!(|data: &[u8]| {
```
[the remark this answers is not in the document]
matt-o-how: Fixed

fuzz/fuzz_targets/deserialize_br.rs
```diff
@@ -15,9 +19,13 @@ fuzz_target!(|data: &[u8]| {
 
     let b1 = node_to_bytes_backrefs(&allocator, program).unwrap();
```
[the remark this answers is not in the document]
matt-o-how: Done

fuzz/fuzz_targets/deserialize_br.rs
```diff
@@ -15,9 +19,13 @@ fuzz_target!(|data: &[u8]| {
...
 
-    let mut allocator = Allocator::new();
-    let program = node_from_bytes_backrefs(&mut allocator, &b1).unwrap();
+    let program = node_from_bytes_backrefs(&mut allocator, &b1);
```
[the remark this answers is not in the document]
matt-o-how: Removed b1, and now just reference data with no repeated calls

fuzz/fuzz_targets/deserialize_br_test_pair_count.rs
```diff
@@ -0,0 +1,29 @@
...
+use clvmr::serde::node_from_bytes_backrefs;
+use clvmr::serde::node_from_bytes_backrefs_old;
+use clvmr::serde::node_to_bytes_backrefs;
+use libfuzzer_sys::fuzz_target;
+
+// #[cfg(feature = "counters")]
+fuzz_target!(|data: &[u8]| {
+    let mut allocator = Allocator::new();
+    let cp = allocator.checkpoint();
+    let program = match node_from_bytes_backrefs(&mut allocator, data) {
```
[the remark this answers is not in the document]
matt-o-how: Done

src/allocator.rs
```diff
@@ -332,13 +336,34 @@ impl Allocator {
...
-        if idx == MAX_NUM_PAIRS {
+        if idx == self.max_num_pairs {
             return err(self.nil(), "too many pairs");
         }
         self.pair_vec.push(IntPair { first, rest });
         Ok(NodePtr::new(ObjectType::Pair, idx))
     }
 
+    pub fn reduce_pair_max(&mut self, amount: usize) -> Result<(), EvalErr> {
+        if self.max_num_pairs - self.pair_vec.len() < amount {
```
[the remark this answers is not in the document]
matt-o-how: Implemented this along with a test which uses the counters flag

src/allocator.rs
matt-o-how: Done

src/allocator.rs
matt-o-how: Done

src/allocator.rs
matt-o-how: Done

src/allocator.rs
arvidn: but there's also a risk that the `+` will exceed the capacity of the variable. Perhaps you only ever call this with `amount=1`, in which case it's safe. But in that case, maybe this function should always increment by 1.

src/allocator.rs
```diff
@@ -332,13 +342,31 @@ impl Allocator {
...
+            return err(self.nil(), "too many pairs");
+        }
+        self.num_ghost_pairs += amount;
+        Ok(())
+    }
+
+    // this code is used when we actually create the pairs that were previously skipped ghost pairs
+    pub fn remove_ghost_pair(&mut self, amount: usize) -> Result<(), EvalErr> {
+        // currently let this panic with overflow if we go below 0 to debug if/where it happens
+        self.num_ghost_pairs -= amount;
```
arvidn, line 366: it might make sense to have a `debug_assert!(self.num_ghost_pairs >= amount)`.
But I suppose that's implied in the integer wrapping

src/serde/de_br.rs
```diff
@@ -83,8 +135,95 @@ pub fn node_from_bytes_backrefs_record(
...
+                    .into());
+                }
+                SExp::Pair(left, right) => {
+                    sexp_to_parse = if is_bit_set { right } else { left };
+                }
+            }
+        } else if is_bit_set {
+            // we have traversed right ("rest"), so we keep processing the Vec
+            if arg_index == 0 {
+                // if we have reached the end of the stack, we must start parsing as NIL
```
arvidn, line 185: I don't quite understand this comment. I think it might be outdated.
Shouldn't you initialize `sexp_to_parse` to `NIL` here? Right now, it will just be set to the value at the bottom of the stack, iiuc.
I see there's a test for this, but I don't see why it works

src/serde/de_br.rs
```diff
@@ -126,11 +271,93 @@ mod tests {
...
+        assert_eq!(
+            traverse_path_with_vec(&mut a, &[0b10], &mut list).unwrap(),
+            n2
+        );
+        assert_eq!(
+            traverse_path_with_vec(&mut a, &[0b101], &mut list).unwrap(),
+            n1
+        );
+        assert_eq!(
+            traverse_path_with_vec(&mut a, &[0b111], &mut list).unwrap(),
```
arvidn, line 351: I think these cases would be interesting as well:
```
traverse_path_with_vec(&mut a, &[0b1011], &mut list)
traverse_path_with_vec(&mut a, &[0b1111], &mut list)
```

The should both fail as they try to traverse NIL

src/serde/de_br.rs
matt-o-how: `sexp_to_parse` is initialised as `NIL` on line 165, before we start the loop and doesn't change unless we traverse left (line 194)

src/serde/de_br.rs
matt-o-how: `traverse_path_with_vec(&mut a, &[0b1011], &mut list)` was already tested on line 356, but I've now added `[0b1111]` next to it

src/allocator.rs
matt-o-how: Added

src/allocator.rs
```diff
@@ -177,6 +179,8 @@ pub struct Allocator {
     // the number of small atoms we've allocated. We keep track of these to ensure the limit on the
     // number of atoms is identical to what it was before the small-atom optimization
     small_atoms: usize,
+    max_num_atoms: usize,
+    num_ghost_pairs: usize,
```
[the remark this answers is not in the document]
matt-o-how: Added

src/allocator.rs
```diff
@@ -107,6 +107,8 @@ pub struct Checkpoint {
     pairs: usize,
     atoms: usize,
     small_atoms: usize,
+    max_num_atoms: usize,
```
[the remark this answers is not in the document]
matt-o-how: Reverted this to previous functionality

arvidn:
These are the benchmark results I get on raspberry Pi 5:
```
deserialize/node_from_bytes_backrefs-compressed
                        time:   [352.36 µs 358.58 µs 367.28 µs]
deserialize/node_from_bytes_backrefs_old-compressed
                        time:   [344.03 µs 350.53 µs 358.28 µs]
```

So, a slight performance degradation. I would have expected it to be a slight improvement on a system with small caches. But presumably the memory savings is worth it. Do we have a measurement on that?

src/serde/de_br.rs
```diff
@@ -338,7 +340,43 @@ mod tests {
...
+        );
+        // confirm we have no cache before we call the function
+        assert!(list[0].1.is_none());
+        assert!(list[1].1.is_none());
+        a.add_ghost_pair(2).unwrap();
+        let test_val = traverse_path_with_vec(&mut a, &[0b1], &mut list).unwrap();
+        assert!(node_eq(&a, test_val, list_node));
+        // check cache has been created
+        assert!(list[0].1.is_some());
+        assert!(list[1].1.is_some());
```
arvidn, line 367: you could even compare these, with `node_eq()` against `list_node_part` and `list_node`, right?
They won't be the same nodes, but they should compare equal.

src/serde/de_br.rs
```diff
@@ -338,7 +340,43 @@ mod tests {
...
+        assert!(node_eq(&a, test_val, list_node));
+
+        // try with cache generated outside of function
+        let mut list: Vec<(NodePtr, Option<NodePtr>)> =
+            vec![(n1, Some(list_node_part)), (n2, Some(list_node))];
+
+        assert_eq!(
+            traverse_path_with_vec(&mut a, &[0b1], &mut list).unwrap(),
+            list_node
+        );
```
arvidn, line 379: you might as well throw in:
```
        assert_eq!(
            traverse_path_with_vec(&mut a, &[0b11], &mut list).unwrap(),
            list_node_part
        );
```

src/serde/de_br.rs
matt-o-how: done

src/serde/de_br.rs
matt-o-how: done

PR #536, opened 2025-01-14
[CHIA-2071] [CHIA-2130] Improve fuzzer

arvidn:
re-implement `make_tree()` to use the arbitrary crate, and make it able to generate trees that reuse nodes.

prior to this patch, `make_tree()` was recursive and could cause stack overflow in the fuzzer. Likewise, `node_eq()` could also cause stack overflow, and was also fixed as part of this PR.


fuzz/fuzz_targets/fuzzing_utils.rs
```diff
@@ -1,83 +1,6 @@
...
-        self.bit_offset = (bits % 8) as u8;
-    }
-}
-
-const BUFFER: [u8; 63] = [
-    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-];
-
-pub fn make_tree(a: &mut Allocator, cursor: &mut BitCursor, short_atoms: bool) -> NodePtr {
```
arvidn, line 57: this function moved into the `make_tree.rs` file, and completely changed as well

fuzz/fuzz_targets/object_cache.rs
```diff
@@ -1,30 +1,61 @@
...
+    let mut unstructured = arbitrary::Unstructured::new(data);
     let mut allocator = Allocator::new();
-    let program = make_tree(&mut allocator, &mut cursor, short_atoms);
+    let program = make_tree::make_tree(&mut allocator, &mut unstructured);
 
     let mut hash_cache = ObjectCache::new(treehash);
     let mut length_cache = ObjectCache::new(serialized_length);
     visit_tree(&allocator, program, |a, node| {
         let expect_hash = tree_hash(a, node);
-        let expect_len = node_to_bytes(a, node).unwrap().len() as u64;
```
arvidn, line 19: now that `make_tree()` can generate trees that reuse nodes, the *uncompressed* serialization can be very very large. This runs out of memory, so the function above is a cheaper way to compute the would-be serialized length

Rigidity (approved):

PR #537, opened 2025-01-15
address incremental-serializer fuzzer failures

arvidn:
There's a performance problem in `insert_sentinel()` if it's passed a tree that is large because it reuses sub-trees. This case can cause the iteration to pick every tree node to run for a very long time. This change makes `insert_sentinel()` only consider nodes once, and skips them the second time they're encountered.

Without this change the fuzzer can time out.

### DIFF
--- a/fuzz/fuzz_targets/incremental_serializer.rs
+++ b/fuzz/fuzz_targets/incremental_serializer.rs
@@ -1,16 +1,18 @@
 #![no_main]
 
 mod make_tree;
+mod node_eq;
 
-use clvmr::serde::{node_from_bytes_backrefs, node_to_bytes, Serializer};
+use clvmr::serde::{node_from_bytes_backrefs, Serializer};
 use clvmr::{Allocator, NodePtr, SExp};
 use make_tree::make_tree_limits;
+use std::collections::HashMap;
 
 use libfuzzer_sys::fuzz_target;
 
 enum TreeOp {
     SExp(NodePtr),
-    Cons,
+    Cons(NodePtr),
 }
 
 // returns the new root (with a sentinel) as well as the sub-tree under the
@@ -30,6 +32,7 @@ fn insert_sentinel(
     let mut copy = Vec::new();
     let mut ops = vec![TreeOp::SExp(root)];
     let mut subtree: Option<NodePtr> = None;
+    let mut copied_nodes = HashMap::<NodePtr, NodePtr>::new();
 
     while let Some(op) = ops.pop() {
         match op {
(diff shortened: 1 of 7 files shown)


fuzz/fuzz_targets/object_cache.rs
```diff
@@ -43,19 +43,47 @@ fn compute_serialized_len(a: &Allocator, n: NodePtr) -> u64 {
...
 fuzz_target!(|data: &[u8]| {
     let mut unstructured = arbitrary::Unstructured::new(data);
     let mut allocator = Allocator::new();
-    let program = make_tree::make_tree(&mut allocator, &mut unstructured);
+    let (tree, node_count) =
+        make_tree::make_tree_limits(&mut allocator, &mut unstructured, 10_000, true);
 
     let mut hash_cache = ObjectCache::new(treehash);
     let mut length_cache = ObjectCache::new(serialized_length);
-    visit_tree(&allocator, program, |a, node| {
```
arvidn, line 53: this fuzzer was changed to testing a single node, picked based on the input data, rather than looping over all of them. testing all nodes risks taking a long time, timing out the fuzzer.

fuzz/fuzz_targets/incremental_serializer.rs
```diff
@@ -81,22 +91,22 @@ fuzz_target!(|data: &[u8]| {
...
     // this just needs to be a unique NodePtr, that won't appear in the tree
     let sentinel = allocator.new_pair(NodePtr::NIL, NodePtr::NIL).unwrap();
 
     let checkpoint = allocator.checkpoint();
     // count up intil we've used every node as the sentinel/cut-point
-    let mut node_idx = 0;
+    let node_idx = unstructured.int_in_range(0..=node_count).unwrap_or(5) as i32;
 
     // try to put the sentinel in all positions, to get full coverage
-    while let Some((first_step, second_step)) =
```
arvidn, line 95: this fuzzer was changed from testing every node to just picking a single one, picked based on the input. This is to avoid timing out when generating large trees

fuzz/fuzz_targets/node_eq.rs
```diff
@@ -1,12 +1,17 @@
...
 
 /// compare two CLVM trees. Returns true if they are identical, false otherwise
 pub fn node_eq(allocator: &Allocator, lhs: NodePtr, rhs: NodePtr) -> bool {
     let mut stack = vec![(lhs, rhs)];
+    let mut visited = HashSet::<NodePtr>::new();
 
     while let Some((l, r)) = stack.pop() {
         match (allocator.sexp(l), allocator.sexp(r)) {
             (SExp::Pair(ll, lr), SExp::Pair(rl, rr)) => {
+                if !visited.insert(l) {
```
arvidn, line 12: avoiding comparing the same sub-trees multiple times is critical to make this function perform under a fuzzer

matt-o-how (approved):

PR #553, opened 2025-02-20
back-reference path length

arvidn:
fix issue where sometimes the back-reference path would be as long as the serialized length of the node being referenced.

There are two related issues:

1. paths are searched breadth-first, so when we find a path that's about to exceed the limit, we know all other paths are at the limit (and there's no point in resuming the search). Then it's safe to `return possible_responses`.
2. When incorporating a completed search into `possible_responses` we do not take a potential length-prefix into account. This results in back-reference paths sometimes being the same size or longer than the node we're referencing. This is fixed by an extra check to reject paths that are too long.

The new fuzzer implements serialization but with an extra check to ensure every back-reference is within the `node_serialized_length` limit.

To make sure this doesn't materially impact performance, I ran the serialization benchmarks against `main`.

```
Benchmarking serialize/node_to_bytes_backrefs 0: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 42.0s, or reduce sample count to 10.
serialize/node_to_bytes_backrefs 0
                        time:   [427.89 ms 432.34 ms 438.02 ms]
                        change: [-5.0810% -2.5045% -0.1270%] (p = 0.04 < 0.05)
                        Change within noise threshold.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe
Benchmarking serialize/node_to_bytes_backrefs 1: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 49.0s, or reduce sample count to 10.
serialize/node_to_bytes_backrefs 1
                        time:   [482.66 ms 483.77 ms 485.05 ms]
                        change: [-2.2317% -1.6262% -1.0767%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) high mild
  3 (3.00%) high severe
Benchmarking serialize/node_to_bytes_backrefs 2: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 7.8s, enable flat sampling, or reduce sample count to 50.
serialize/node_to_bytes_backrefs 2
                        time:   [1.5393 ms 1.5408 ms 1.5426 ms]
                        change: [-4.8583% -4.4210% -4.0202%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe
```

### DIFF
--- a/fuzz/fuzz_targets/serializer_cmp.rs
+++ b/fuzz/fuzz_targets/serializer_cmp.rs
@@ -1,0 +1,103 @@
+#![no_main]
+
+mod make_tree;
+mod node_eq;
+
+use clvmr::allocator::{Allocator, NodePtr, SExp};
+use clvmr::serde::node_from_bytes_backrefs;
+use clvmr::serde::write_atom::write_atom;
+use clvmr::serde::ReadCacheLookup;
+use clvmr::serde::{serialized_length, treehash, ObjectCache};
+use std::io;
+use std::io::Cursor;
+use std::io::Write;
+
+use node_eq::node_eq;
+
+use libfuzzer_sys::fuzz_target;
+
+const BACK_REFERENCE: u8 = 0xfe;
+const CONS_BOX_MARKER: u8 = 0xff;
+
+#[derive(PartialEq, Eq)]
+enum ReadOp {
+    Parse,
+    Cons,
+}
+
+// make sure back-references returned by ReadCacheLookup are smaller than the
+// node they reference
+pub fn compare_back_references(allocator: &Allocator, node: NodePtr) -> io::Result<Vec<u8>> {
+    let mut f = Cursor::new(Vec::new());
+
+    let mut read_op_stack: Vec<ReadOp> = vec![ReadOp::Parse];
+    let mut write_stack: Vec<NodePtr> = vec![node];
+
+    let mut read_cache_lookup = ReadCacheLookup::new();
+
+    let mut thc = ObjectCache::new(treehash);
+    let mut slc = ObjectCache::new(serialized_length);
+
+    while let Some(node_to_write) = write_stack.pop() {
+        let op = read_op_stack.pop();
+        assert!(op == Some(ReadOp::Parse));
+
+        let node_serialized_length = *slc
+            .get_or_calculate(allocator, &node_to_write, None)
+            .expect("couldn't calculate serialized length");
+        let node_tree_hash = thc
+            .get_or_calculate(allocator, &node_to_write, None)
+            .expect("can't get treehash");
+
+        let result1 = read_cache_lookup.find_path(node_tree_hash, node_serialized_length);
+        match result1 {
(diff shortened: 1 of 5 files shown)


fuzz/fuzz_targets/serializer_cmp.rs
```diff
@@ -0,0 +1,103 @@
...
+                {
+                    // make sure the path is never encoded as more bytes than
+                    // the node we're referencing
+                    use std::io::Write;
+                    let mut temp = Cursor::new(Vec::<u8>::new());
+                    temp.write_all(&[BACK_REFERENCE])?;
+                    write_atom(&mut temp, &path)?;
+                    let temp = temp.into_inner();
+                    assert!(temp.len() < node_serialized_length as usize);
+                }
```
arvidn, line 67: this is the main test

src/serde/read_cache_lookup.rs
```diff
@@ -162,12 +170,14 @@ impl ReadCacheLookup {
                     for (parent, direction) in items.iter() {
                         if *(self.count.get(parent).unwrap_or(&0)) > 0 && !seen_ids.contains(parent)
                         {
-                            if path.len() + 1 > max_path_length {
+                            if path.len() > max_path_length {
```
richardkiss, line 173: There's an argument to be where with ties, we go with the back-ref, the idea being, it's no bigger but we end up with (maybe?) less memory used since we avoid allocating objects. Not really a deal-breaker though.

richardkiss (approved):
Makes sense. Left one comment, feel free to self-resolve it since I'm not requesting any changes unless you want to.

src/serde/read_cache_lookup.rs
arvidn: yeah, I had the same thought. It probably makes sense to do that. I wanted to start with fixing the intention of the original code though.

src/serde/read_cache_lookup.rs
```diff
@@ -153,7 +154,14 @@ impl ReadCacheLookup {
             let mut new_partial_paths = vec![];
             for (node, path) in partial_paths.iter_mut() {
                 if *node == self.root_hash {
-                    possible_responses.push(reversed_path_to_vec_u8(path));
+                    // make sure we never return a path that needs more (or the
+                    // same) bytes to serialize than the node we're referencing.
+                    if let Some(path_len) = atom_length_bits(path.len() as u64) {
+                        if path_len < max_bytes_for_path_encoding {
```
arvidn, line 160: this `<` was a mistake. It should be `<=` since `max_bytes_for_path_encoding` has already subtracted the mandatory 1 byte saving.

PR #554, opened 2025-02-20
Use Vec::with_capacity() in Allocator

matt-o-how:
allocate memory for the allocator ahead of time to prevent memcopy when we grow


Rigidity:
How much memory does this use?

PR #555, opened 2025-02-23
tie-break in favor of back-references

arvidn:
allow back-reference paths to encode to the same number of bytes as the node it's referencing.

The rationale is that deduplicating nodes have more benefits than just saving space, it may also makes it cheaper to compute tree hashes.

### DIFF
--- a/src/serde/read_cache_lookup.rs
+++ b/src/serde/read_cache_lookup.rs
@@ -142,7 +142,7 @@ impl ReadCacheLookup {
             RandomState::default(),
         );
 
-        let max_bytes_for_path_encoding = serialized_length - 2; // 1 byte for 0xfe, 1 min byte for savings
+        let max_bytes_for_path_encoding = serialized_length - 1; // 1 byte for 0xfe
         let max_path_length: usize = (max_bytes_for_path_encoding.saturating_mul(8) - 1)
             .try_into()
             .unwrap_or(usize::MAX);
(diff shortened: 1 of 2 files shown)


src/serde/read_cache_lookup.rs
```diff
@@ -156,10 +156,13 @@ impl ReadCacheLookup {
...
                     // same) bytes to serialize than the node we're referencing.
-                    if let Some(path_len) = atom_length_bits(path.len() as u64) {
-                        if path_len < max_bytes_for_path_encoding {
-                            let p = reversed_path_to_vec_u8(path);
-                            possible_responses.push(p);
+                    // path.len() + 1 is because reversed_path_to_vec_u8() will
+                    // also add the "terminator" bit, at the far left (MSB)
+                    // if we have 8 steps to traverse, we need 9 bits to represent
+                    // it as a path
+                    if let Some(path_len) = atom_length_bits(path.len() as u64 + 1) {
```
arvidn, line 163: this `+ 1` fixes a mistake. `reversed_path_to_vec_u8()` adds the left-most "terminator" bit to the path. That bit has to be taken into account when predicting how many bytes the path will be encoded as

richardkiss: I wonder if the `+1` should be absorbed into the implementation of `.len`? What's the type of `path` here?

arvidn: it used to be `Vec<bool>` but I recently changed it to `BitVec`

richardkiss (approved):

PR #557, opened 2025-02-24
factor out parts of the fuzzers into separate files

arvidn:
since we don't have a "fuzzer utils" crate yet, these functions are individual modules included in the fuzzers directly. If we put them in the same file, we can get warnings for dead code if we don't use all the functions.

### DIFF
--- a/src/serde/ser_br.rs
+++ b/src/serde/ser_br.rs
@@ -63,7 +63,7 @@ pub fn node_to_stream_backrefs<W: io::Write>(
                 }
             },
         }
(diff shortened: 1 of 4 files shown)


src/serde/ser_br.rs
```diff
@@ -63,7 +63,7 @@ pub fn node_to_stream_backrefs<W: io::Write>(
                 }
             },
         }
-        while !read_op_stack.is_empty() && read_op_stack[read_op_stack.len() - 1] == ReadOp::Cons {
+        while let Some(ReadOp::Cons) = read_op_stack.last() {
```
arvidn, line 66: this was an unrelated clean-up

Rigidity (approved):

PR #567, opened 2025-03-21
[CHIA-2052] TreeCache optimization

arvidn:
Pairs don't need to be hashed, since all atoms are deduplicated identical pairs have the same child nodes, so they can just be deduplicated based on those. only atoms are hashed for deduplication.

This improves timing with up to 25%, but more commonly in the 10% range in the benchmarks. 

MacBook M1:
```
group            optimized-tree-cache    tree-cache
-----            --------------------    ----------
Serializer 0     1.00    131.8±7.03ms    1.18    156.1±6.97ms
Serializer 1     1.00    112.8±2.90ms    1.13    127.3±3.73ms
Serializer 2     1.00    109.6±0.92µs    1.33    145.3±1.23µs
Serializer 3     1.00  1870.4±64.51µs    1.09      2.0±0.01ms
Serializer 4     1.00    235.1±0.96µs    1.11    260.2±2.36µs
```

RPi-5:
```
group            optimized-tree-cache    main
-----            --------------------    ----------
Serializer 0     1.00   453.1±41.70ms    1.14   518.4±49.44ms
Serializer 1     1.00   366.0±32.49ms    1.13   414.5±32.71ms
Serializer 2     1.00   362.0±59.69µs    1.23   444.9±68.23µs
Serializer 3     1.00      5.2±0.91ms    1.10      5.7±0.95ms
Serializer 4     1.00  750.2±113.10µs    1.02  761.9±118.06µs
```

### DIFF
--- a/src/serde/tree_cache.rs
+++ b/src/serde/tree_cache.rs
@@ -13,12 +13,10 @@ const MIN_SERIALIZED_LENGTH: u64 = 4;
 
 type Bytes20 = [u8; 20];
 
-fn hash_blobs(salt: &[u8], blobs: &[&[u8]]) -> Bytes20 {
+fn hash_atom(salt: &[u8], blob: &[u8]) -> Bytes20 {
     let mut ctx = Sha1::default();
     ctx.update(salt);
-    for blob in blobs.iter() {
-        ctx.update(blob);
-    }
+    ctx.update(blob);
     ctx.finalize().into()
 }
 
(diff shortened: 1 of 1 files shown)


src/serde/tree_cache.rs
```diff
@@ -274,37 +276,24 @@ impl TreeCache {
                     let right = &self.node_entries[right_idx];
                     let serialized_length = 1 + left.serialized_length + right.serialized_length;
 
-                    let (hash, idx) = if let (Some(left_hash), Some(right_hash)) =
-                        (left.tree_hash, right.tree_hash)
```
arvidn, line 278: this check for whether the `tree_hash` is `None` needs a counterpart in the new way of doing this. As this is written now, the parent of the sentinel node will have a non-zero serialized length, but it's supposed to be "viral".

src/serde/tree_cache.rs
```diff
@@ -95,7 +93,11 @@ pub struct TreeCache {
     /// node_entries vector. For any given tree hash, we're only supposed to
     /// have a single NodeEntry. There may be multiple NodePtr referring to
     /// the same NodeEntry (if they are identical sub trees).
-    hash_to_node: HashMap<Bytes20, u32, RandomState>,
+    atom_lookup: HashMap<Bytes20, u32, RandomState>,
+
+    /// maps left + right child indices to the index of the pair with those
+    /// children. This is the atom_lookup counterpart for pairs
+    pair_lookup: HashMap<u64, u32>,
```
richardkiss, line 100: Any reason not write it as the (IMO more clear) `(u32, u32)` ?

Who knows how rust would encode it. Are you concerned it would be slower?

PR #580, opened 2025-05-21
[CHIA-2843] add function to check if a serialization uses canonical encoding of atoms

arvidn:


src/serde/tools.rs
```diff
@@ -282,124 +340,116 @@ mod tests {
         );
     }
 
-    #[cfg(test)]
-    mod test {
-        use super::*;
-        use crate::serde::node_from_bytes_backrefs;
-        use crate::Allocator;
-        use rstest::rstest;
```
arvidn, line 290: this inner `test` module seemed like a copy-paste error. I removed it, so this is best reviewed ignoring whitespace

Rigidity (approved):

PR #586, opened 2025-06-11
Introduce reference counters for atoms and pairs in Allocator

arvidn:
These work as substitutes for node_from_bytes_backrefs_record(), which records which nodes are referenced more than once in deserialization. These are used for memoization in tree_hash_cached(). Using reference counters in the Allocator allows memoization of structures generated programmatically as well.

TODO


Rigidity (approved):

arvidn:
I updated this to not keep reference counts for atoms, since it doesn't seem to be worth memoizing anyway

Rigidity (dismissed):

Rigidity (dismissed):

PR #601, opened 2025-07-24
CHIA-2072 Change fuzzers to use new chia-fuzzers lib + Update bls & sha2

jack60612:
chia_rs updated, ready for review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,6 +67,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.98"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e16d2d3311acee920a9eb8d33b8cbc1787ce4a264e85f964c2404b969bdcd487"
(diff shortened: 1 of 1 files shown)


jack60612:
this is expected to fail as the versions dont exist yet.

jack60612 (commented):
cargo.lock needs to be pushed once chia_rs is out.

Cargo.toml
```diff
@@ -89,6 +90,10 @@ sha1 = { workspace = true }
 bumpalo = { workspace = true }
 thiserror = "1.0.69"
 
+[patch.crates-io]
+# So the chia fuzzing crate can be used in the workspace
+clvmr = { path = "." }
```
arvidn, line 95: this looks suspicious. why can't we use the crates.io version?

jack60612: clvmr, a dependency of chia-fuzzing was a different version then the currently being developed version / the local source version so you would be unable to run the fuzzing tests

PR #606, opened 2025-07-31
update crate workflow to support workspaces

jack60612:


Cargo.toml
```diff
@@ -38,8 +38,8 @@ openssl = ["chia-sha2/openssl"]
 lto = "thin"
 
 [workspace.dependencies]
-clvmr = { path = "." }
-clvm-fuzzing = { path = "./clvm-fuzzing" }
+clvmr = { path = ".", version = "0.16.1" }
+clvm-fuzzing = { path = "./clvm-fuzzing", version = "0.16.1" }
```
arvidn, line 42: I don't think you need to specify the versions here

jack60612: discussed on keybase.

PR #632, opened 2025-09-25
Add op_sha256tree

matt-o-how:
This PR adds a costed and cached shatree which accounts for cost as if it isn't caching so that further improvements may be made in the future without breaking consensus.

Attached are the benchmarked performance graphs for cost of call, cost per pair and cost for 32byte chunk.

sha256tree-per-pair
sha256tree-per-byte
sha256tree-base

### DIFF
--- a/docs/sha256tree.md
+++ b/docs/sha256tree.md
@@ -1,6 +1,6 @@
 # Information on the new (sha256tree) operator
 
-Adding `sha256tree` as a native operator made sense as it is one of the most common functions, used in nearly every shipped ChiaLisp puzzle. 
+Adding `sha256tree` as a native operator made sense as it is one of the most common functions, used in nearly every shipped ChiaLisp puzzle.
 Furthermore it has an innate inefficiency in it's in-language implementation.
 Every internal hash is allocated as an atom in clvm_rs allocator.
 In addition to this, a native operator also opens the door to future optimisations via caching.
@@ -9,8 +9,9 @@ In addition to this, a native operator also opens the door to future optimisatio
 
 The matter of how to assign Cost to the new operator was the subject of intense thought and debate.
 It should be costed in proportion to:
+
 - the time to Cost vs other operators, and especially the `sha256` operator
-- the time to Cost ratio of the ChiaLisp implementation 
+- the time to Cost ratio of the ChiaLisp implementation
 - the size of its inputs
 
 The lattermost being a unique problem with regards to shatree as it is the first operator that parses trees, so utmost care must be taken when assigning cost.
@@ -21,7 +22,7 @@ One final consideration while costing is that we required it to tally the cost d
 
 The `BASE_COST` was set to equal the base cost of `sha256`.
 
-The `COST_PER_BYTES32` was designed as the sha256 operation operates on 32byte chunks. We set the Cost to be on parity with the Cost of `sha256` although sha256 costs `per byte` and `per arg`. 
+The `COST_PER_BYTES32` was designed as the sha256 operation operates on 32byte chunks. We set the Cost to be on parity with the Cost of `sha256` although sha256 costs `per byte` and `per arg`.
 We can ignore `per arg` as `sha256tree` only takes a single argument, and we benchmarked the `cost-per-bytes32` so that it matches `sha256`'s `cost-per-byte`.
 
 The calculations for this can be seen in the file `benchmark-clvm-cost.rs`.
@@ -34,8 +35,9 @@ The calculations for this can be seen in the file `sha256tree-benching.rs`.
 ## Costing Results
 
 `MacOS M1`
+
 ```
-Costs based on an increasing atom per bytes32 chunks: 
+Costs based on an increasing atom per bytes32 chunks:
 Native time per bytes32  (ns): 95.1425
 CLVM   time per bytes32  (ns): 94.9895
 Native implementation takes 100.1611% of the time.
@@ -43,7 +45,7 @@ Native cost per bytes32      : 64.0000
 CLVM   cost per bytes32      : 64.0000
 100.1611% of the CLVM cost is:  : 64.1031
 
-Costs based on growing a balanced binary tree: 
+Costs based on growing a balanced binary tree:
 Native time per node  (ns): 203.8718
 CLVM   time per node  (ns): 517.8038
 Native implementation takes 39.3724% of the time.
@@ -51,7 +53,7 @@ Native cost per node      : 564.0000
 CLVM   cost per node      : 1463.0000
 39.3724% of the CLVM cost is:  : 576.0185
 
-Costs based on growing a list: 
+Costs based on growing a list:
 Native time per node  (ns): 115.0891
 CLVM   time per node  (ns): 397.1927
 Native implementation takes 28.9756% of the time.
@@ -61,6 +63,7 @@ CLVM   cost per node      : 1399.0000
 ```
 
 `Windows`
+
 ```
 Costs based on an increasing atom per bytes32 chunks:
 Native time per bytes32  (ns): 10.4049
@@ -87,7 +90,6 @@ CLVM   cost per node      : 1399.0000
 10.0416% of the CLVM cost is:  : 140.4821
 ```
 
-
 ## Costing Graphs
 
 Below are the generated benchmarking graphs (PNG) from running `sha256tree-benching.rs` on Macbook M1.
@@ -103,5 +105,3 @@ Below are the generated benchmarking graphs (PNG) from running `sha256tree-bench
 ![Tree - Time per Node](graphs/tree_bench.png)
 
 ![Tree - Cost per Node](graphs/tree_cost.png)
-
-


src/treehash.rs
```diff
@@ -0,0 +1,219 @@
...
+    Cons,
+    ConsAddCacheCost(NodePtr, Cost),
+}
+
+macro_rules! th {
+    ($hash:expr) => {
+        TreeHash::new(hex!($hash))
+    };
+}
+pub const PRECOMPUTED_HASHES: [TreeHash; 24] = [
```
arvidn: could you reuse this constant from the `sha256` operator instead?

src/treehash.rs
```diff
@@ -0,0 +1,219 @@
...
+use crate::cost::Cost;
+use crate::ObjectType;
+use crate::SExp;
+use chia_sha2::Sha256;
+use hex_literal::hex;
+use std::fmt;
+use std::ops::Deref;
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct TreeHash([u8; 32]);
```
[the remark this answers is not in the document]
matt-o-how: Done

src/sha_tree_op.rs
```diff
@@ -0,0 +1,93 @@
...
+                // cost_left is the remaining after computing it
+                // the cost of this subtree = after - before
+                let used = cost - cost_before;
+                cache.insert(original_node, &hash, used);
+            }
+        }
+    }
+
+    assert!(hashes.len() == 1);
+    Ok(Reduction(cost, a.new_atom(&hashes[0])?))
```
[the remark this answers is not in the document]
matt-o-how: Added

src/chia_dialect.rs
```diff
@@ -29,6 +30,8 @@ pub const LIMIT_HEAP: u32 = 0x0004;
 // This is a hard-fork and should only be enabled when it activates
 pub const ENABLE_KECCAK_OPS_OUTSIDE_GUARD: u32 = 0x0100;
 
+pub const ENABLE_SHA256_TREE: u32 = 0x0200;
```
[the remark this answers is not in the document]
matt-o-how: Done

op-tests/test-sha256tree.txt
arvidn: you should include negative tests here as well. e.g. passing an atom, passing 0 arguments, passing > 1 argument

src/treehash.rs
```diff
@@ -151,47 +151,48 @@ pub fn tree_hash_cached_costed(
...
-                NodeVisitor::U32(val) => {
-                    cost += SHA256TREE_COST_PER_BYTE * a.atom_len(node) as u64;
-                    check_cost(cost, cost_left)?;
-                    if (val as usize) < PRECOMPUTED_HASHES.len() {
-                        hashes.push(PRECOMPUTED_HASHES[val as usize]);
-                    } else {
-                        hashes.push(tree_hash_atom(a.atom(node).as_ref()));
+            TreeOp::SExp(node) => {
+                // charge a call cost for processing this op
+                cost += SHA256TREE_COST_PER_CALL;
```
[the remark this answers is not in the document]
matt-o-how: This is confusing, and it was tripping me up, but this is costing both atoms and pairs.
It's now adding the cost per call when it matches any SExp, the Cons and ConsCache arms are completing calls that started inside the SExp branch. 

It was over-charging before for both the processing of the SExp, and then completion of the task in the processing of the Cons.

src/treehash.rs
```diff
@@ -1,85 +1,37 @@
...
+pub fn tree_hash_atom(bytes: &[u8]) -> [u8; 32] {
     let mut sha256 = Sha256::new();
     sha256.update([1]);
     sha256.update(bytes);
-    TreeHash::new(sha256.finalize())
+    sha256.finalize()
 }
 
-pub fn tree_hash_pair(first: TreeHash, rest: TreeHash) -> TreeHash {
+pub fn tree_hash_pair(first: [u8; 32], rest: [u8; 32]) -> [u8; 32] {
```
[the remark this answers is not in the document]
matt-o-how: done

src/sha_tree_op.rs
```diff
@@ -0,0 +1,93 @@
+use crate::allocator::NodeVisitor;
+use crate::allocator::{Allocator, NodePtr};
+use crate::cost::{check_cost, Cost};
+use crate::op_utils::get_args;
+use crate::reduction::{Reduction, Response};
+use crate::treehash::*;
+
+const SHA256TREE_BASE_COST: Cost = 0;
```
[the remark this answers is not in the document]
arvidn: I would just expect that the cost of traversing the tree would be smaller than the up-front cost of just calling an operator. But perhaps not. If it's zero, I don't think we need it.

tools/src/bin/benchmark-clvm-cost.rs
```diff
@@ -512,6 +512,13 @@ pub fn main() {
             extra: None,
             flags: NESTING_BASE_COST | PER_ARG_COST | PER_BYTE_COST | LARGE_BUFFERS,
         },
+        Operator {
+            opcode: 65,
+            name: "sha256tree (atom)",
+            arg: Placeholder::SingleArg(Some(g1)),
```
arvidn: I don't think `g1` is right here, it looks like it's just going to pass in valid G1 points. I can't recall exactly how this tool works though.

arvidn (commented):
do you feel confident that the cost benchmarks are good? specifically the cost per byte, cost per pair and cost per atom?

tools/generate-sha256tree-tests.py
```diff
@@ -0,0 +1,87 @@
...
+        if len(obj) == 0:
+            return "0"
+        return f"0x{obj.hex()}"
+    else:
+        left, right = obj
+        return f"({sexp_repr(left)} . {sexp_repr(right)})"
+
+
+test_cases = set()
+with open("../op-tests/test-sha256tree-hash.txt", "w") as f:
```
richardkiss: This causes this tool to depend upon being started from a particular directory. A clever trick is to ask the script where it lives and build the relative path from there. Something like

```python
p = pathlib.Path(__file__).parent.parent / "op-tests/test-sha256tree-hash.txt"
```

Now current directory doesn't matter; just the relative path between the `.py` and `.txt` file.

tools/generate-sha256tree-tests.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
+from hashlib import sha256
+from random import randbytes, choice, randint, seed
```
richardkiss: It's tempting to use these nice functions from `random`. Be cautioned that they hide global state (which you correctly set with `seed` to make these tests deterministic). That's fine in this case because things are so constrained but I want to be sure you're aware that `random` also includes stateful `Random` objects that you could explicitly pass through the call stack that would make these functions more "pure" and not dependent on local state.

So yeah, this is fine, but I just wanted to pass on an annoyance I have with random number generator libraries: they often hide state in secret globals.

tools/src/bin/benchmark-clvm-cost.rs
arvidn: maybe richard's suggestion is simpler. However, traversing the tree (probably) has a cost as well, not just sha256.

tools/src/bin/benchmark-clvm-cost.rs
```diff
@@ -118,7 +118,7 @@ fn substitute(args: Placeholder, s: NodePtr) -> OpArgs {
 fn time_invocation(a: &mut Allocator, op: u32, arg: OpArgs, flags: u32) -> f64 {
     let call = build_call(a, op, arg, 1, None);
     //println!("{:x?}", &Node::new(a, call));
-    let dialect = ChiaDialect::new(0);
+    let dialect = ChiaDialect::new(0x0200);
```
arvidn: is it not possible to use the named flags here?

tools/src/bin/sha256tree-benching.rs
```diff
@@ -0,0 +1,111 @@
...
+        writeln!(output, "{}\t{}", sample.0, sample.1).expect("failed to write");
+        samples.push(sample);
+
+        a.restore_checkpoint(&checkpoint);
+    }
+
+    linear_regression_of(&samples).expect("linreg failed")
+}
+
+fn time_per_cons_for_list(a: &mut Allocator, output: &mut dyn Write) -> (f64, f64) {
```
arvidn: I thought you would fold this into the `benchmark-clvm-costs.rs`. I think seeing the plots would be helpful as well, as a confirmation that it looks right, and that it's linear.

matt-o-how: Integrated now

tools/src/bin/benchmark-clvm-cost.rs
```diff
@@ -242,7 +241,7 @@ fn base_call_time(
     }
 
     let (slope, _): (f64, f64) = linear_regression_of(&samples).expect("linreg failed");
-    slope
+    slope.max(100.0)
```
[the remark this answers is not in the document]
matt-o-how: Done

tools/src/bin/benchmark-clvm-cost.rs
```diff
@@ -264,7 +264,69 @@ fn base_call_time_no_nest(a: &mut Allocator, op: &Operator, per_arg_time: f64) -
...
+
+// this adds 32 bytes at a time compared to per_byte which adds 5 at a time
+// at the moment this func is only for sha256tree and will need adapting in the future for other uses
+fn time_per_byte_for_atom(a: &mut Allocator, output: &mut dyn Write) -> (f64, f64) {
+    let mut samples = Vec::<(f64, f64)>::new();
+    let dialect = ChiaDialect::new(ENABLE_SHA256_TREE); // enable shatree
+
+    let op_code = a.new_number(63.into()).unwrap();
+    let quote = a.new_number(1.into()).unwrap();
+    let mut atom_str = "ff".repeat(10_000);
```
arvidn: I think it would be simpler to just create the bytes directly, instead of creating a hex string to then convert to bytes.

```suggestion
    let mut atom = [0xff].repeat(10_000);
```

tools/src/bin/benchmark-clvm-cost.rs
```diff
@@ -264,7 +264,69 @@ fn base_call_time_no_nest(a: &mut Allocator, op: &Operator, per_arg_time: f64) -
...
+    let checkpoint = a.checkpoint();
+
+    for i in 0..10000 {
+        // make the atom longer as a function of i
+        atom_str.push_str(&((i % 89) + 10).to_string().repeat(32)); // just to mix it up
+        let atom = a.new_atom(&hex::decode(&atom_str).unwrap()).unwrap();
+        // let args = a.new_pair(atom, a.nil()).unwrap();
```
arvidn:
```suggestion
        atom.append([(i % 89 + 10) as u8].repeat(32)) // just to mix it up
```

tools/src/bin/benchmark-clvm-cost.rs
```diff
@@ -264,7 +264,69 @@ fn base_call_time_no_nest(a: &mut Allocator, op: &Operator, per_arg_time: f64) -
...
+
+    let op_code = a.new_number(63.into()).unwrap();
+    let quote = a.new_number(1.into()).unwrap();
+    let mut list = a.nil();
+
+    for _ in 0..500 {
+        list = a.new_pair(a.nil(), list).unwrap();
+    }
+
+    for i in 0..10000 {
```
arvidn: I think you could probably add `.step(500)` here, and add 500 entries in each step. i.e. move the above loop into here

tools/src/bin/benchmark-clvm-cost.rs
```diff
@@ -273,6 +335,8 @@ const NESTING_BASE_COST: u32 = 4;
 const EXPONENTIAL_COST: u32 = 8;
 const LARGE_BUFFERS: u32 = 16;
 const ALLOW_FAILURE: u32 = 32;
+const LIST_LENGTH_COST: u32 = 64;
+const ALT_PER_BYTE_COST: u32 = 128;
```
arvidn: I think it would be good to add a comment here to explain what `ALT_` means. (and maybe even the descriptions I posted on keybase for the other flags)

arvidn:
I get this output on a MacBookPro M1:
```
opcode: sha256tree (65)
   time: per-byte: 94.27ns
   cost: per-byte: 606
   time: base: 32.37ns
   cost: base: 100
list length slope: 393.014814881, intercept: 191116.295402940, cost (slope * cost_scale): 2525.423427810
```

It would be nice if the `list length` output would be consistent with the other metrics

tools/src/bin/benchmark-clvm-cost.rs
```diff
@@ -535,6 +605,13 @@ pub fn main() {
             println!("   time: per-byte: {time_per_byte:.2}ns");
             println!("   cost: per-byte: {:.0}", time_per_byte * cost_scale);
             time_per_byte
+        } else if (op.flags & ALT_PER_BYTE_COST) != 0 {
+            let mut output = maybe_open(options.plot, op.name, "per-byte.log");
+            let (slope, _intercept) = time_per_byte_for_atom(&mut a, &mut output);
+            let cost = slope * cost_scale;
+            println!("   time: per-byte: {slope:.2}ns");
+            println!("   cost: per-byte: {:.0}", cost);
```
arvidn:
```suggestion
            println!("   time: per-32bytes: {slope:.2}ns");
            println!("   cost: per-32bytes: {:.0}", cost);
```
Maybe something like this?

tools/src/bin/benchmark-clvm-cost.rs
```diff
@@ -264,7 +264,69 @@ fn base_call_time_no_nest(a: &mut Allocator, op: &Operator, per_arg_time: f64) -
...
+    linear_regression_of(&samples).expect("linreg failed")
+}
+
+// at the moment this func is only for sha256tree and will need adapting in the future for other uses
+fn time_per_cons_for_list(a: &mut Allocator, output: &mut dyn Write) -> (f64, f64) {
+    let mut samples = Vec::<(f64, f64)>::new();
+    let dialect = ChiaDialect::new(ENABLE_SHA256_TREE); // enable shatree
+
+    let op_code = a.new_number(63.into()).unwrap();
+    let quote = a.new_number(1.into()).unwrap();
```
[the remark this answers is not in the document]
matt-o-how: Done

tools/src/bin/benchmark-clvm-cost.rs
matt-o-how: Done

src/chia_dialect.rs
```diff
@@ -193,7 +202,10 @@ impl Dialect for ChiaDialect {
             // Extension 1 is for the keccak256 operator.
             1 => OperatorSet::Keccak,
 
-            // Extensions 2 and beyond are considered invalid by the mempool.
+            // Extension 2 is for the sha256tree operator.
+            2 => OperatorSet::Sha256tree,
+
+            // Extensions 3 and beyond are considered invalid by the mempool.
```
arvidn: This is also part of the `softfork` operator. I'm not sure we want to do this. I think it would keep things simpler by not adding another extension here.

tools/src/bin/benchmark-clvm-cost.rs
```diff
@@ -588,6 +678,18 @@ pub fn main() {
...
             );
         }
+        if (op.flags & LIST_LENGTH_COST) != 0 {
+            write_gnuplot_header(&mut *gnuplot, op, "per-pair", "num pairs");
+            let mut output = maybe_open(options.plot, op.name, "per-pair.log");
+            let (slope, intercept): (f64, f64) = time_per_cons_for_list(&mut a, &mut output);
+            let cost = slope * cost_scale;
+
+            println!("   time: per-pair: {:.2}ns", slope);
+            println!("   cost: per-pair: {:.0}", cost);
```
arvidn:
```suggestion
            println!("   time: per-node: {:.2}ns", slope);
            println!("   cost: per-node: {:.0}", cost);
```
You call this `PER_NODE` in the cost constant

src/run_program.rs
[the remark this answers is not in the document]
matt-o-how: Reverted these changes now

src/chia_dialect.rs
matt-o-how: Reverted these changes now

arvidn:
these are the results I get on RPi5:

```
opcode: sha256tree (65)
   time: per-32byte: 132.39ns
   cost: per-32byte: 851
   time: base: 69.35ns
   cost: base: 165
   time: per-node: 598.82ns
   cost: per-node: 3848
   intercept: 296832.75
```

src/treehash.rs
```diff
@@ -0,0 +1,537 @@
...
+            };
+            if self.visit(left) {
+                nodes.push(left);
+            }
+            if self.visit(right) {
+                nodes.push(right);
+            }
+        }
+    }
+}
```
arvidn: I think all code related to the cached version should be removed until we have a good solution. I suspect it will need to look very different from this

src/treehash.rs
```diff
@@ -0,0 +1,537 @@
...
+
+enum TreeOp {
+    SExp(NodePtr),
+    Cons,
+    ConsAddCacheCost(NodePtr, Cost),
```
arvidn: including this one

tools/src/bin/benchmark-clvm-cost.rs
```diff
@@ -268,12 +267,81 @@ fn base_call_time_no_nest(a: &mut Allocator, op: &Operator, per_arg_time: f64) -
     (total_time - per_arg_time * num_samples as f64) / num_samples as f64
 }
 
+// this adds 32 bytes at a time compared to per_byte which adds 5 at a time
+fn time_per_byte_for_atom(a: &mut Allocator, op: &Operator, output: &mut dyn Write) -> (f64, f64) {
```
arvidn: it makes sense that just looking at the slope here, you isolate the cost per 32 bytes. The constant factor is unknown as it includes base cost and cost per node.

tools/src/bin/sha256tree-benching.rs
```diff
@@ -130,19 +136,20 @@ fn time_per_cons_for_list(
         let cost = red.0;
         let result_2 = node_to_bytes(a, red.1).expect("should work");
         assert_eq!(result_1, result_2);
-
         let duration = start.elapsed().as_nanos() as f64;
+        let duration = (duration - (3.0 * bytes32_clvm_time)) / 2.0;
+        let cost = (cost as f64 - (3.0 * bytes32_clvm_cost)) / 2.0;
```
arvidn: same here, I would expect this program to compare the *cost* of the native operator to the CLVM implementation. But here you alter the cost.

src/treehash.rs
```diff
@@ -8,8 +8,11 @@ use crate::reduction::Reduction;
 use crate::reduction::Response;
 use chia_sha2::Sha256;
 
+// the base cost is the cost of calling it to begin with
 const SHA256TREE_BASE_COST: Cost = 30;
-const SHA256TREE_COST_PER_NODE: Cost = 3000;
+// this is the cost per node, whether it is a cons box or an atom
+const SHA256TREE_COST_PER_NODE: Cost = 2000;
```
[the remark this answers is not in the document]
matt-o-how: It was applying this cost for both pairs and atoms.

tools/src/bin/benchmark-clvm-cost.rs
```diff
@@ -318,7 +323,8 @@ fn time_per_cons_for_list(a: &mut Allocator, op: &Operator, output: &mut dyn Wri
         let start = Instant::now();
         run_program(a, &dialect, call, a.nil(), 11000000000).unwrap();
         let duration = start.elapsed();
-        let sample = (i as f64, duration.as_nanos() as f64);
+        let duration_f64 = (duration.as_nanos() as f64 - (4.0 * time_per_byte32)) / 2.0;
```
arvidn: I think this `/ 2.0` warrants a comment

tools/src/bin/benchmark-clvm-cost.rs
```diff
@@ -297,6 +297,10 @@ fn time_per_byte_for_atom(a: &mut Allocator, op: &Operator, output: &mut dyn Wri
     linear_regression_of(&samples).expect("linreg failed")
 }
 
+// this function is used for calculating a theoretical cost per node
+// we pass in the time it takes for a byte32 chunk and subtract 4*chunk_time
+// we then divide by two to account for the fact that we are adding a nil atom each time the list grows too
+// this is because atoms are nodes too
```
arvidn: What's the point of this function? It doesn't look like it's used for anything. If you were to make it measure the cost of 1 pair and 1 NIL-atom, you could use the result to validate the model of only charging for the blocks of hashed bytes. But as it is right now, what do you do with the result?

Do you check to see that it's close to 0?

tools/src/bin/sha256tree-benching.rs
```diff
@@ -7,6 +7,12 @@ use std::fs::File;
 use std::io::{BufWriter, Write};
 use std::time::Instant;
 
+/*
+This file is for comparing the native sha256tree with the clvm implementation which previously existed.
+The costs for the native implementation should be lower as it is not required to make allocations.
+*/
```
arvidn: It's not clear to me how the output of this program should be interpreted. We want it to tell whether the current cost for the native shatree operator is reasonable compared to the CLVM implementation.

How can we tell?

Why does the actual timing matter? It's just the cost that matters, isn't it?

src/treehash.rs
```diff
@@ -0,0 +1,180 @@
...
+use chia_sha2::Sha256;
+
+// the base cost is the cost of calling it to begin with
+// this is set to the same as sha256
+const SHA256TREE_BASE_COST: Cost = 87;
+// this cost is applied for every node we traverse to
+const SHA256TREE_NODE_COST: Cost = 500;
+// this is the cost for every 32 bytes in a sha256 call
+// it is set to the same as sha256
+const SHA256TREE_COST_PER_32_BYTES: Cost = 64;
```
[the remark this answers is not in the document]
matt-o-how: I've added an extra set of calculations and outputs to `sha256tree-benching` representing a growing in depth balanced tree

tools/src/bin/sha256tree-benching.rs
```diff
@@ -0,0 +1,291 @@
...
+        atom_nat_t,
+        atom_clvm_t,
+        cons_native_time,
+        cons_clvm_time,
+        cons_native_cost,
+        cons_clvm_cost,
+    );
+
+    // taken from benchmark-clvm-cost.rs
+    let cost_scale = ((101094.0 / 39000.0) + (1343980.0 / 131000.0)) / 2.0;
```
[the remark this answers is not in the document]
matt-o-how: `sha256tree-benching` now takes a measurement of the time -> cost ration between the native and clvm implementations

tools/src/bin/sha256tree-benching.rs
```diff
@@ -0,0 +1,449 @@
...
+        let args = a.new_pair(quote, atom_node).unwrap();
+        let call = a.new_pair(args, a.nil()).unwrap();
+        let call = a.new_pair(op_code, call).unwrap();
+
+        let checkpoint = a.checkpoint();
+
+        // native
+        let start = Instant::now();
+        let red = run_program(a, &dialect, call, a.nil(), 11_000_000_000).unwrap();
+        let cost = red.0;
```
arvidn: this cost seems like a circular dependency. This program is meant to *establish* this cost, not present what it happens to be set to right now. It really seems like something that easy to misunderstand. I don't know what it tells us here, but I do see how it can be confusing. I think it would be best to remove it, but at least it should have a comment explaining what it's for.

The main risk I see is that it, accidentally, is used to make it appear as if  a bad cost is good, because it matches this.

tools/src/bin/sha256tree-benching.rs
```diff
@@ -0,0 +1,449 @@
...
+        atom_nat_t,
+        atom_clvm_t,
+        cons_native_time,
+        cons_clvm_time,
+        cons_native_cost,
+        cons_clvm_cost,
+    );
+
+    // taken from benchmark-clvm-cost.rs
+    let cost_scale = ((101094.0 / 39000.0) + (1343980.0 / 131000.0)) / 2.0;
```
arvidn: Since you measure time and cost for the clvm implementation, you meant to compute this scale factor, right? and not use this constant

docs/sha256tree.md
```diff
@@ -0,0 +1,100 @@
...
+## Costing Results
+
+`MacOS M1`
+```
+Costs based on an increasing atom per bytes32 chunks: 
+Native time per bytes32  (ns): 95.1425
+CLVM   time per bytes32  (ns): 94.9895
+Native implementation takes 100.1611% of the time.
+Native (time_per_bytes32  * cost_ratio): 611.3642
+CLVM   (time_per_bytes  * cost_ratio) : 610.3807
```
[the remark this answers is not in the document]
matt-o-how: These are removed now for irrelevancy

tools/src/bin/sha256tree-benching.rs
matt-o-how: This is removed now for irrelevancy

tools/src/bin/sha256tree-benching.rs
```diff
@@ -0,0 +1,449 @@
...
+        let result_1 = node_to_bytes(a, red.1).expect("should work");
+        let duration = start.elapsed().as_nanos() as f64;
+        // a new list entry is 2 nodes (a cons and a nil) and a 3 chunk hash operation and a 1 chunk hash operation
+        // this equation lets us figure out a theoretical cost just for a node
+        let duration = (duration - (500.0 + i as f64) * (4.0 * bytes32_native_time)) / 2.0;
+        let cost = (cost as f64 - (500.0 + i as f64) * (4.0 * bytes32_native_cost)) / 2.0;
+        writeln!(output_native_time, "{}\t{}", i, duration).unwrap();
+        writeln!(output_native_cost, "{}\t{}", i, cost).unwrap();
+        samples_time_native.push((i as f64, duration));
+        samples_cost_native.push((i as f64, cost as f64));
```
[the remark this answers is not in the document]
matt-o-how: Done

PR #708, opened 2026-02-23
[CHIA-3823] 2026 serde format

richardkiss:
This pull request introduces an experimental new serialization format ("2026 serialization format") for CLVM, focused on efficiency through atom and pair deduplication, variable-length integer encoding, and improved compression. The new format is implemented in a dedicated module and is enabled via the `ser-2026` feature flag. Benchmarks and tests have been added to evaluate and validate the new serialization and deserialization logic.

This format is not backward compatible, but it's dead simple. It interns the contents first, then serializes all the atoms, then builds the tree. It's much faster to serialize than backref-style, and it compresses better. It approaches the speed of "classic" (and can even be faster in some cases because of interning).

The main trick is a `varint`, a variable-length encoded integer that uses bit-prefix tricks like how utf-8 and clvm serialization do. With it, numbers close to 0 can be encoded in 1 byte, larger (absolute value) numbers can be encoded in 2, and even larger can be encoded in 3. We don't need to support any integers larger than that.

Here's a sample compression benchmark:

  | file | format | ser (µs) | deser (µs) | size | vs legacy |
|------|--------|----------|------------|------|-----------|
| 0.generator | legacy | 3,702 | 3,728 | 948,848 | — |
| | backrefs | 397,100 | 452 | 253,078 | 26.7% |
| | **2026** | **69,478** | **449** | **201,539** | **21.2%** |
| 1.generator | legacy | 4,784 | 3,966 | 678,024 | — |
| | backrefs | 468,451 | 782 | 242,450 | 35.8% |
| | **2026** | **42,378** | **192** | **154,849** | **22.8%** |
| 2.generator | legacy | 7 | 5 | 1,426 | — |
| | backrefs | 866 | 6 | 1,426 | 100.0% |
| | 2026 | 182 | 5 | 1,434 | 100.6% |
| 3.generator | legacy | 65 | 59 | 23,788 | — |
| | backrefs | 11,169 | 91 | 11,051 | 46.5% |
| | **2026** | **2,025** | **35** | **9,725** | **40.9%** |
| 4.generator | legacy | — | — | — | — |
| | backrefs | 1,666 | 10 | 1,002 | — |
| | **2026** | **196** | **5** | **844** | — |

---

> [!NOTE]
> **Medium Risk**
> Adds a brand-new serialization/deserialization implementation (including varint parsing and new framing), which is an error-prone surface area and could introduce correctness or DoS edge cases. Also changes Python `Program.from_bytes` to auto-detect formats, which may alter behavior for callers expecting legacy-only parsing.
> 
> **Overview**
> Implements a new **magic-prefixed `serde_2026` wire format** (atom/pair interning + varint-based atom table and instruction stream) with new Rust APIs (`serialize_2026`, `deserialize_2026`, and `serialized_length_serde_2026`) and re-exports via `serde`.
> 
> Extends the Python wheel with **explicit legacy/backrefs/2026 serializers/deserializers**, an `auto` format switch, and a `clvm_tree_to_lazy_node` converter to intern Python trees for efficient 2026 serialization; updates `Program` helpers (`from_bytes`, `from_bytes_backrefs`, `from_bytes_2026`, `to_bytes_2026`) and adds comprehensive Python tests.
> 
> Adds fuzz targets and Criterion benchmarks for 2026 serialize/deserialize plus minimal `Criterion.toml` configuration, and includes new format documentation in `docs/serde-2026.md`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1766fa26c388aad4d06802afb1c175a47771fcbb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


richardkiss:
Rough algorithm for deserialization: (APR 29, 2026: this is obsolete. The 0 opcode is now the `nil` atom, and `1` and `-1` are cons, in different orders.)

```
def deserialize_2026(allocator, data):
    cursor = Cursor(data)

    # 1) Read atom table header: how many distinct length groups there are.
    # decode_varint returns a signed varint (here used for counts/lengths).
    atoms = []                      # flat list of atom NodePtr in encoded order
    atom_lengths_count = decode_varint(cursor)  # number of length-groups
    atom_buffer = bytearray()       # reused buffer to avoid repeated allocations

    # Each group can be:
    # - a positive varint L: exactly one atom of length L follows
    # - a negative varint -L: then a varint count C follows, then C atoms of length L follow
    for _ in range(atom_lengths_count):
        atom_length = decode_varint(cursor)

        # Interpret group header: negative => (length, count), positive => (length, 1)
        if atom_length < 0:
            length = -atom_length           # actual byte length of each atom in this group
            count = decode_varint(cursor)   # how many atoms of that length
        else:
            length = atom_length
            count = 1

        # Resize the single reusable buffer to 'length' bytes.
        # This avoids allocating a new byte array for each atom.
        atom_buffer[:] = b'\x00' * length

        # Read 'count' atoms of fixed 'length' bytes each.
        for _ in range(count):
            # read_exact fills atom_buffer with the next 'length' bytes
            cursor.read_exact(atom_buffer) or raise SerializationError
            # create a new atom node from bytes and append to atoms list
            atom_node = allocator.new_atom(bytes(atom_buffer))
            atoms.append(atom_node)

    # 2) Read instruction stream length: how many instructions follow
    instruction_count = decode_varint(cursor)

    # Special-case: instruction_count == 0 means there are no pairs.
    # The serialized root must then be a single atom (the first atom read).
    if instruction_count == 0:
        if len(atoms) == 0:
            raise SerializationError()  # malformed: no atom available
        return atoms[0]

    # 3) Execute the instruction stream to reconstruct the tree
    # 'pairs' stores Pair NodePtrs in the order they are constructed so that
    # negative instructions can refer back to previously-constructed pairs by index.
    pairs = []                      # list of pair NodePtr, append when new pair is constructed
    stack = []                      # evaluation stack used by the instructions

    # Instruction meanings:
    # - instr == 0        : cons -> pop left,right; create pair(left,right); push pair
    # - instr > 0         : push atom (1-based index into atoms array: atom_idx = instr - 1)
    # - instr < 0         : push previously constructed pair (negative 1-based index into pairs: pair_idx = -instr - 1)
    for _ in range(instruction_count):
        instr = decode_varint(cursor)

        if instr == 0:
            # cons needs two items on the stack (left then right)
            if len(stack) < 2:
                raise SerializationError()   # malformed stream / stack underflow
            right = stack.pop()
            left = stack.pop()
            pair = allocator.new_pair(left, right)  # allocate pair node
            pairs.append(pair)                       # record construction order
            stack.append(pair)                       # push constructed pair onto stack

        elif instr > 0:
            # push atom: encoded as 1-based index to allow zero to be reserved for cons
            atom_idx = instr - 1
            if atom_idx < 0 or atom_idx >= len(atoms):
                raise SerializationError()   # invalid atom reference
            stack.append(atoms[atom_idx])

        else:  # instr < 0
            # push already-constructed pair: encoded as negative 1-based index into 'pairs'
            pair_idx = -instr - 1
            if pair_idx < 0 or pair_idx >= len(pairs):
                raise SerializationError()   # invalid pair reference
            stack.append(pairs[pair_idx])

    # After processing all instructions, exactly one item (the root) must remain.
    if len(stack) != 1:
        raise SerializationError()   # malformed: extra/missing operations
    return stack[0]```

File: src/serde_2026/varint.rs [hunk 1]
```diff
@@ -0,0 +1,139 @@
...
+    // Extract value bits from first byte (the high bits of the value)
+    let value_mask = (1u8 << bits_in_first_byte) - 1;
+    let mut unsigned_value = (first_byte & value_mask) as u64;
+
+    // Read additional bytes using fixed-size buffer (max 7 extra bytes)
+    if leading_ones > 0 {
+        let mut extra_bytes = [0u8; 7];
+        cursor
+            .read_exact(&mut extra_bytes[..leading_ones])
+            .map_err(|_| EvalErr::SerializationError)?;
```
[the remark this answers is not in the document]
richardkiss: fixed

src/serde_2026/varint.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    // Convert from two's complement to signed
+    let sign_bit = 1u64 << (total_value_bits - 1);
+    if unsigned_value >= sign_bit {
+        // Negative value: subtract 2^total_value_bits
+        Ok(unsigned_value as i64 - (1i64 << total_value_bits))
+    } else {
+        Ok(unsigned_value as i64)
+    }
```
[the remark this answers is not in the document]
richardkiss: Fixed

docs/serde-2026.md
```diff
@@ -0,0 +1,98 @@
...
+- `0xfd 0xff` drives legacy/backrefs decoders down an invalid atom-size path,
+  causing immediate failure.
+- `0x32 0x30 0x32 0x36` is ASCII `"2026"` for readable hexdumps.
+
+### Detection
+
+A deserializer determines the format by inspecting the first 6 bytes:
+
+- If the blob starts with `0xfd 0xff 0x32 0x30 0x32 0x36`, it is 2026-format.
+- Otherwise, parse with the legacy/backrefs path.
```
arvidn, line 23: A feature like makes sense in a soft-fork, where some nodes will stay on the old version. but in our hard-fork, I don't see when we would ever have a blob with unknown format, that we need to detect. It would reduce some complexity to just know based on whether the fork has activated or not.

docs/serde-2026.md
```diff
@@ -0,0 +1,98 @@
...
+root node.
+
+### Varint Encoding
+
+Signed integers are encoded with a variable-length prefix scheme:
+
+```
+0xxxxxxx                          →  7-bit value, range [-64, 63]
+10xxxxxx xxxxxxxx                 → 14-bit value, range [-8192, 8191]
+110xxxxx xxxxxxxx xxxxxxxx        → 21-bit value, range [-1048576, 1048575]
```
arvidn, line 101: this is the same as UTF-8, right? (if so, it might be worth mentioning)

docs/serde-2026.md
```diff
@@ -0,0 +1,98 @@
...
+
+1. **Atom table** — all unique atoms (except nil), grouped by length
+2. **Instruction stream** — stack-based operations to reconstruct the tree
+
+### Atom Table
+
+Nil (the empty atom) is **not** included in the atom table — it has a dedicated
+opcode (`0`) in the instruction stream.
+
+The atom table begins with a varint encoding the number of atom groups.
```
arvidn, line 40: Do I understand correctly that a 3 byte varint is the largest valid encoding? So the atom table is limited to 2097151 atoms?

docs/serde-2026.md
```diff
@@ -0,0 +1,98 @@
...
+110xxxxx xxxxxxxx xxxxxxxx        → 21-bit value, range [-1048576, 1048575]
+...
+```
+
+The number of leading `1` bits determines how many additional bytes follow. A
+`0` separator bit follows the leading `1`s. The remaining bits (across all
+bytes) form a two's-complement signed integer in big-endian order. This scales
+to wider integers without changing the encoding rules.
+
+A prefix of 8 leading `1` bits (`0xFF`) is invalid.
```
arvidn, line 119: If these are also valid, I think it would be good to list them. iirc, the PR description suggested that 3-byte encoding was the maximum, but this seems to say everything up to 7 bytes is valid (which is different than UTF-8)

```
1110xxxx xxxxxxxx xxxxxxxx xxxxxxxx
11110xxx xxxxxxxx xxxxxxxx xxxxxxxx xxxxxxxx
111110xx xxxxxxxx xxxxxxxx xxxxxxxx xxxxxxxx xxxxxxxx
1111110x xxxxxxxx xxxxxxxx xxxxxxxx xxxxxxxx xxxxxxxx xxxxxxxx
11111110 xxxxxxxx xxxxxxxx xxxxxxxx xxxxxxxx xxxxxxxx xxxxxxxx xxxxxxxx
```

docs/serde-2026.md
```diff
@@ -0,0 +1,98 @@
...
+
+Nil (the empty atom) is **not** included in the atom table — it has a dedicated
+opcode (`0`) in the instruction stream.
+
+The atom table begins with a varint encoding the number of atom groups.
+
+For each group (in stream order):
+
+- If the group contains **one** atom: a positive varint encoding the atom's byte
+  length, followed by the atom's raw bytes.
```
arvidn, line 45: You should document the length-limit of atoms here probably

docs/serde-2026.md
```diff
@@ -0,0 +1,98 @@
...
+  encoding the negated byte length, then a positive varint encoding the count,
+  then the raw bytes of each atom concatenated (each is exactly `length` bytes).
+
+Atoms are assigned indices starting from 0, in the order they appear in the
+table.
+
+The decoder accepts groups in any order. Multiple groups with the same byte
+length are valid (they contribute separate atom indices). A serializer may
+choose a specific ordering strategy (for example, sorting by frequency so
+commonly-referenced atoms get smaller varint indices).
```
arvidn:
```suggestion
commonly-referenced atoms get smaller indices).
```

I understand that the reason you might prefer a lower index is the smaller varint encoding, but I think it's misleading to refer to the index as "varint index"

docs/serde-2026.md
```diff
@@ -0,0 +1,98 @@
...
+| `N >= 2`     | Push the atom at index N-2 onto the stack                        |
+| `N <= -2`    | Push the already-constructed pair at index -N-2 onto the stack   |
+
+Pairs are indexed in construction order (the first pair cons'd is index 0, the
+second is index 1, etc.). A negative instruction references a pair that was
+previously constructed during this same decode, enabling shared sub-trees
+without re-encoding them.
+
+After all instructions execute, the stack must contain exactly one item: the
+root node.
```
arvidn, line 92: it's not obvious to me why an instruction stream is better than just declaring every pair as two varints (left and right). You could make the last pair be the resulting tree.

The benefits of that is that you don't need a stack. I suppose the drawback is that you can't (cheaply) reference the pair you just defined as one of the children, so it uses a bit more space, is that right?

is there a limit to the stack size?

docs/serde-2026.md
```diff
@@ -0,0 +1,98 @@
...
+0xxxxxxx                          →  7-bit value, range [-64, 63]
+10xxxxxx xxxxxxxx                 → 14-bit value, range [-8192, 8191]
+110xxxxx xxxxxxxx xxxxxxxx        → 21-bit value, range [-1048576, 1048575]
+...
+```
+
+The number of leading `1` bits determines how many additional bytes follow. A
+`0` separator bit follows the leading `1`s. The remaining bits (across all
+bytes) form a two's-complement signed integer in big-endian order. This scales
+to wider integers without changing the encoding rules.
```
arvidn: I think overlong encodings should be considered errors

fuzz/coverage/serde-2026/raw/default-serde-2026.profraw
arvidn, line 1: I don't think you meant to add this file

fuzz/fuzz_targets/serde_2026.rs
arvidn, line 1: I think you should have a fuzzer that creates an arbitrary CLVM tree (using `ArbitraryClvmTree`), serialize it and deserialize it and ensure they match.

A dedicated fuzzer for `decode_varint()` would probably be a good idea as well

.gitignore
```diff
@@ -19,10 +19,13 @@ target/
...
+*.so
 
 # Node.js
 /target
 /node_modules
 
 # Mac generates these with the benchmarking tools
-.DS_Store
\ No newline at end of file
+.DS_Storestate/
```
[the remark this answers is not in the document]
arvidn: I don't think you mean to make this change

wheel/src/lazy_node.rs
```diff
@@ -46,4 +46,12 @@ impl LazyNode {
...
         }
     }
+
+    pub(crate) fn allocator(&self) -> &Allocator {
+        &self.allocator
+    }
+
+    pub(crate) fn node(&self) -> NodePtr {
+        self.node
+    }
```
arvidn: what are these used for? They're not exposed to python, so I imagine it's something in the python binding

wheel/python/clvm_rs/serde.py
```diff
@@ -0,0 +1,101 @@
...
+}
+
+_SERIALIZERS = {
+    "legacy": ser_legacy,
+    "backrefs": ser_backrefs,
+    "2026": ser_2026,
+}
+
+
+def deserialize(
```
arvidn, line 58: this is a risky function, are you sure we need it?
I can envision this being used when parsing network messages, opening up attacks that the consumer didn't expect

src/serde_2026/varint.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    #[test]
+    fn test_decode_rejects_truncated_multibyte() {
+        // 2-byte varint (0x80 prefix) with missing second byte
+        assert!(decode_varint(&mut Cursor::new(&[0x80][..])).is_err());
+        // 3-byte varint (0xc0 prefix) with only 1 extra byte
+        assert!(decode_varint(&mut Cursor::new(&[0xc0, 0x00][..])).is_err());
+        // 4-byte varint (0xe0 prefix) with only 2 extra bytes
+        assert!(decode_varint(&mut Cursor::new(&[0xe0, 0x00, 0x00][..])).is_err());
+    }
```
arvidn, line 203: also test rejecting overlong encodings

src/serde_2026/strategy.rs
```diff
@@ -0,0 +1,99 @@
...
+    type NodeCtx = ();
+
+    fn root_ctx(&self, _state: &SerializerState) -> Self::NodeCtx {}
+
+    fn decide(&self, _state: &SerializerState, _pair_idx: usize, _ctx: ()) -> (Direction, (), ()) {
+        let s = self
+            .state
+            .get()
+            .wrapping_mul(6364136223846793005)
+            .wrapping_add(1442695040888963407);
```
arvidn: you don't like the standard rust rngs?

src/serde_2026/strategy.rs
```diff
@@ -0,0 +1,99 @@
...
+        let s = if seed == 0 {
+            0x9E37_79B9_7F4A_7C15
+        } else {
+            seed
+        };
+        Self {
+            state: std::cell::Cell::new(s),
+        }
+    }
+}
```
arvidn: I can see this being useful in the fuzzer and maybe some tests. but I don't think this belongs among the production code. Probably just in the fuzzer.

src/serde_2026/ser_optimized.rs
```diff
@@ -0,0 +1,144 @@
+use super::ser::SerializerState;
+use super::strategy::{Direction, VisitStrategy};
+
+/// Tree-DP strategy: pick left-first vs right-first per pair to maximise the
+/// number of pair back-references that fit in the 1-byte varint tier.
+///
+/// Complexity: O(N x min(subtree_size, 64)) to build; O(1) per pair at
+/// emit time.
+pub(super) struct DpOptimized {
```
arvidn: This strategy looks risky. It might have very expensive edge cases. I would have expected the choice between left or right to purely depend on what's at the top of the stack:

only left child => push the right child and create the pair
only right child => push the left child and create the pair
left then right => create the pair
right then left => create the pair

src/serde_2026/ser.rs
```diff
@@ -0,0 +1,349 @@
...
+    }
+    Ok(())
+}
+
+/// Walk the interned pair tree under `strategy`, emitting an instruction
+/// stream. Only the visit order at each pair varies across strategies.
+pub(super) fn emit_instructions<S: VisitStrategy>(
+    state: &SerializerState,
+    strategy: &S,
+) -> Vec<i64> {
```
arvidn: I'm starting to think this customization point does not carry its own weight. Even in the fuzzers, we would primarily be interested in testing the one production strategy, that optimizes.

src/serde_2026/ser.rs
```diff
@@ -0,0 +1,349 @@
...
+            atom_remap,
+            nil_old_idx,
+            pairs,
+            root_index,
+        })
+    }
+}
+
+/// Write the atom table (nil excluded) grouped by contiguous equal lengths.
+pub(super) fn write_atom_table<W: Write>(
```
arvidn: I think this deserves a unit test

docs/serde-2026.md
```diff
@@ -0,0 +1,98 @@
...
+
+The decoder accepts groups in any order. Multiple groups with the same byte
+length are valid (they contribute separate atom indices). A serializer may
+choose a specific ordering strategy (for example, sorting by frequency so
+commonly-referenced atoms get smaller varint indices).
+
+### Instruction Stream
+
+The instruction stream begins with a varint encoding the total number of
+instructions.
```
arvidn, line 70: If you omit this counter, the serializer could stream the instructions directly into the serialized format, avoiding allocating and copying them into a separate vector first.

I don't think this length-prefix is needed, just like we don't have a length prefix to the current CLVM serialization.

src/serde_2026/de.rs
```diff
@@ -0,0 +1,233 @@
...
+use crate::serde::node_from_bytes_backrefs;
+
+use super::SERDE_2026_MAGIC_PREFIX;
+use super::varint::decode_varint;
+
+/// Default maximum atom length (1 MB).
+pub const DEFAULT_MAX_ATOM_LEN: usize = 1 << 20;
+
+/// Default maximum input bytes (10 MB).
+pub const DEFAULT_MAX_INPUT_BYTES: usize = 10 * (1 << 20);
```
arvidn: are these derived from the new cost per byte and the 11billion block limit? (if so, it's a good idea to document it I think)

src/serde_2026/de.rs
```diff
@@ -0,0 +1,233 @@
...
+
+const MAX_COUNT: usize = 256 * 1024 * 1024;
```
arvidn: this is the max number of atom-groups, atoms in a single group and instructions. I would have expected a limit on the number of atoms and instructions to be separate. But with this single limit, you can still (try) to store 72'057'594'037'927'936 atoms, which is quite many.

arvidn (commented):
I think it might be better it separate out the sophisticated serializer into another PR, to just get the basic support in first.

docs/serde-2026.md
richardkiss: It's not intended for consensus-critical use -- it's more about giving existing tools, like command-line tools that parse clvm, a fast-failure in a consistent way. A `0xfd` as the first byte should cause an immediate parse error.

docs/serde-2026.md
richardkiss: Not quite. It's similar to the trick used for clvm atom parsing, but it's really a "base one" count of the number of bytes in the int. UTF-8 has to reserve some bit patterns to allow streams to pick up at any point in the UTF-8 and still self-sync to the next code point, so it's actually bit more clever (I believe -- not an expert on this).

docs/serde-2026.md
richardkiss: I think that should be enough for any actual generator, but I can make sure it extends to 4 bytes. Seems pretty harmless.

docs/serde-2026.md
richardkiss: It's configurable. I think we should use 1 MB for consensus to match prior limits.

docs/serde-2026.md
richardkiss: The key is, a few varints use one bytes, a bunch more are two bytes, but most are three bytes. So it's really about the size of the serialization – which "clump" each index is in.

docs/serde-2026.md
richardkiss: There is no stack limit. We benefit by pushing the serialized pair onto the stack because we know it will be used later by another pair (or we wouldn't bother making it), so we get to skip declaring the index for newly created pairs the first time they are used. We have to push all atom indices and all memoized pair indices, but each pair creation actually only uses one byte max this way versus two index indices without the stack. Memoized pairs are an even huger savings.

docs/serde-2026.md
richardkiss: A catch of course is, a brand new pair can be created in three bytes ("atom" "atom" "cons") so 4 byte references have to be a bit more careful that they're not serializable in three bytes. But I'm pretty sure no generator we've ever seen has enough nodes to need four bytes.

docs/serde-2026.md
richardkiss: I'll add a `strict` mode.

docs/serde-2026.md
richardkiss: I'm not sure how we would know when to stop if we were reading from a stream. Just because the stack size is 1 doesn't mean we're done.

docs/serde-2026.md
richardkiss: I had Clod (sic) write a straw man implementation your way and the compression was significantly worse without the stack. If you walk through with some small examples, you'll see why.

docs/serde-2026.md
richardkiss: Although the `varint` "standard" I made up supports this, we don't need anything beyond three bytes for consensus.

fuzz/fuzz_targets/serde_2026.rs
richardkiss: Those are done now.

docs/serde-2026.md
arvidn: the only place we'll use this encoding is in the block generator in the `FullBlock` though, right? I assume you intended this header ot be used there.

docs/serde-2026.md
arvidn: right

docs/serde-2026.md
arvidn: the `varint` format does not mention any limits, it should probably be documented either way

docs/serde-2026.md
arvidn: once it's in consensus it's not configurable though. are you suggesting these limits should be documented elsewhere?

docs/serde-2026.md
```diff
@@ -68,31 +85,61 @@ Each instruction is a varint:
...
+```
+
+The number of leading `1` bits determines how many additional bytes follow,
+similar to UTF-8 prefix-length coding. A `0` separator bit follows the leading
+`1`s. The remaining bits (across all bytes) form a two's-complement signed
+integer in big-endian order.
+
+A prefix of 8 leading `1` bits (`0xFF`) is invalid.
+
+The deserializer has a `strict` mode that rejects overlong varint encodings. In
```
arvidn, line 121: I think it's a footgun to not lock this in. Not just because consensus code will have this option to break things, but there really isn't any good reason to ever allow this.

src/serde_2026/de.rs
```diff
@@ -4,7 +4,7 @@ use crate::allocator::{Allocator, NodePtr};
 use crate::error::{EvalErr, Result};
 use crate::serde::node_from_bytes_backrefs;
 
-use super::MAGIC_PREFIX;
+use super::SERDE_2026_MAGIC_PREFIX;
 use super::varint::decode_varint;
 
 /// Default maximum atom length (1 MB).
```
arvidn: I see this is really 1 MiB

wheel/src/api.rs
```diff
@@ -86,11 +95,237 @@ fn deserialize_as_tree(
...
+
+/// Convert a Python CLVM tree (any object with `.atom` / `.pair` attributes)
+/// into a `LazyNode` backed by a Rust `Allocator`, with full interning.
+///
+/// Uses three hash maps mirroring `intern_tree`:
+/// 1. Python object identity (`id()`) -> NodePtr (prevents exponential blowup)
+/// 2. Atom byte content -> NodePtr (deduplicates identical atoms)
+/// 3. (left, right) pair -> NodePtr (deduplicates structurally identical pairs)
+#[pyfunction]
+fn clvm_tree_to_lazy_node(obj: Bound<'_, PyAny>) -> PyResult<LazyNode> {
```
arvidn, line 193: this functionality seems a bit unrelated, or at least fringe, to the main PR.

richardkiss: This is so we can serialize 2026 from python without serializing to some other format first to get it into an `Allocator`. (Claude's first crack was "serialize as backrefs -> deserialize to an Allocator obj -> serialize as 2026" which is ridiculously roundabout.)

docs/serde-2026.md
richardkiss: Even so, we might export full block content to files and then want to dump or parse them with python tools. And we don't preclude having full block content being stored in legacy formats -- only content matters now, not serialization.

docs/serde-2026.md
richardkiss: I think the idea is if we have 2 million atoms, we're paying over 11 billion just for storage, so we can never really get to 4 bytes. Probably a proof exists for this.

docs/serde-2026.md
richardkiss: I don't want consensus details to leak into `clvm_rs` – this library is supposed to be agnostic to details specific to consensus. I'll adjust the api here to be very generic and without defaults, and then we can put a version in `chia_rs` that *is* tuned to consensus.

docs/serde-2026.md
richardkiss: I'll handle this the same way – a consensus-tuned deserializer entrypoint in `chia_rs`.

docs/serde-2026.md
richardkiss: _Cursor here, working through Richard's intuition with him._

Using HF2 consensus constants (`MAX_BLOCK_COST_CLVM = 11_000_000_000`, `COST_PER_BYTE = 12_000`) and `interned_weight(tree) = atom_bytes + 2·A + 3·P` from `chia-consensus/src/generator_cost.rs`, every block satisfies

```
atom_bytes + 2A + 3P  ≤  11_000_000_000 / 12_000  =  916_666
```

Atoms in the table are non-empty, so `atom_bytes ≥ A`, which gives `3A + 3P ≤ 916_666` and therefore **`A + P ≤ 305_555`**.

That bounds the cost-driven counts directly:
- `atom_count = A ≤ 305_555`
- `group_count ≤ atom_count` (one group per distinct length, ≥ 1 atom each)
- per-group `count ≤ atom_count`
- `length ≤ max_atom_len = 1 MiB = 2^20`

`instruction_count` needs one extra step. The deserializer is a stack machine that ends with stack height 1, so `instructions = 2·cons_count + 1`. A minimal encoder uses one cons per unique pair, giving `instruction_count ≤ 2P + 1 ≤ 611_111`. A non-minimal encoder would have to spend ≥ 1 wire byte per redundant cons plus its operand pushes, so reaching `2^21` instructions requires a >2 MiB blob — past the network protocol ceiling.

Every count comes in under 2^21 = 2_097_151 with at least ~3.4× headroom, so the 3-byte varint cap can't reject anything consensus-valid. Probably worth a paragraph in `docs/serde-2026.md`.

_— generated by Cursor (Composer), reviewed but not yet personally re-derived by @richardkiss._

src/serde_2026/ser.rs
richardkiss: [Done](https://github.com/Chia-Network/clvm_rs/pull/708/changes#diff-2a1889cb731815ab753c700221b25a925c27cd43092a7761d743d88b1ba4aa24R339) `test_write_atom_table_groups_by_length`

wheel/python/clvm_rs/serde.py
richardkiss: I don't think consensus will be tempted to use this since it's in python. We do want `clvm_rs` to be a generic tool for dealing with clvm though, and this is a helpful convenience feature.

src/serde_2026/ser.rs
richardkiss: Keeping it as the hook for richer strategies. Today only `LeftFirst` exists and `level` saturates to 0, but the trait is where a future compressor that uses both `-1` and `+1` backref slots (instead of the trivial left-first walk) plugs in.

src/serde_2026/de.rs
```diff
@@ -136,17 +71,23 @@ pub fn deserialize_2026_from_stream<R: Read>(
...
     }
 
     let nil = allocator.nil();
-    let mut pairs: Vec<NodePtr> = Vec::with_capacity(instruction_count / 3);
+    // Don't pre-size `pairs` from `instruction_count`: a tiny blob could
+    // declare ~2^54 instructions and trick `Vec::with_capacity` into a
+    // multi-PB allocation request that aborts the process. The bounded reader
+    // (slice length, or caller-supplied `Take`) terminates the loop long
+    // before we reach a pathological size; geometric growth from `Vec::new()`
+    // handles the typical case in a handful of reallocs.
```
arvidn: you have a fuzzer for this, right? Do you understand why it didn't trigger this?

src/serde_2026/ser.rs
arvidn: there's a risk this will force most users to just run in release builds, not exercising this path. Do you think this adds more coverage on top of the fuzzer?

src/serde_2026/de.rs
```diff
@@ -82,47 +21,43 @@ fn checked_bounded_usize(value: i64, max: usize) -> Result<usize> {
...
+/// [`std::io::Read::take`] — otherwise a malformed blob can drive an unbounded
+/// loop. Total input policy belongs in the caller, not here. Use
+/// [`deserialize_2026_body`] when you have a slice; the slice's length is the
+/// natural bound.
+pub fn deserialize_2026_body_from_stream<R: Read>(
     allocator: &mut Allocator,
     reader: &mut R,
-    options: DeserializeOptions,
+    max_atom_len: usize,
+    strict: bool,
```
arvidn, line 43: `max_atom_len` and `strict` seems to have questionable value in my mind. I would expect we would *always* want to set `strict=true` and if `max_atom_len` was anything but the consensus constant, we'd risk consensus violations/forks.

are there any use cases for `strict=false`?

Combined with the additional test burden these add, it doesn't seem worth it. Right now, I don't see these parameters covered by tests, so they could (probably) break in the future.

The `debug_assert_roundtrip()` also uses `1 << 20` as `max_atom_len`, so that limits what you can pass in here, in debug builds.

wheel/python/clvm_rs/clvm_rs.pyi
```diff
@@ -16,14 +16,14 @@ def deser_backrefs(blob: bytes) -> "LazyNode": ...
 def deser_2026(
     blob: bytes,
     *,
-    max_atom_len: Optional[int] = None,
-    max_input_bytes: Optional[int] = None,
+    max_atom_len: int = ...,
+    strict: bool = False,
```
arvidn: why does `strict` default to `false`?

src/serde_2026/de.rs
richardkiss: It was just for round-tripping. There's a new fuzzer now that feeds in arbitrary bytes which likely would have caught it. I'll try it against the old code base just to make sure.

richardkiss: Yeah, the new fuzzer caught the old problem (`Vec::with_capacity` without checking the int value) immediately.

src/serde_2026/ser.rs
richardkiss: Yes, this was added in error. It seems useless because no one runs in debug mode, so I'm removing.

wheel/python/clvm_rs/clvm_rs.pyi
richardkiss: Since encodings aren't canonical anymore, non-strict encodings are really not a threat anymore. I'd actually lean towards allowing them in consensus since as they can't really permanently infect anything, I don't see how they could actually cause a DOS. But I can change it.

wheel/python/clvm_rs/program.py
```diff
@@ -35,25 +42,47 @@ def stream(self, f: BinaryIO) -> None:
...
+    def from_bytes(cls, blob: bytes) -> Program:
+        """Deserialize from any CLVM format (classic, backrefs, or serde_2026)."""
+        return cls.wrap(deser_auto(blob))
+
+    @classmethod
+    def from_bytes_backrefs(cls, blob: bytes) -> Program:
+        """Deserialize classic or backrefs format only (rejects serde_2026)."""
+        if blob.startswith(SERDE_2026_MAGIC_PREFIX):
+            raise ValueError("unexpected serde_2026 format; use from_bytes() for auto-detection")
+        obj, cursor = cls.from_bytes_with_cursor(blob, 0)
```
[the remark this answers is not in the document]
richardkiss: Fixed.

src/serde_2026/de.rs
richardkiss: These will be set in `chia_rs` now so consensus constants don't leak into here.

src/serde_2026/de.rs
richardkiss: (This comment stream is hard to understand now as the comment it's referring to is gone.)

docs/serde-2026.md
arvidn: can you think of a use case for allowing overlong encodings?
I think it's really questionable to have a feature like this without any tests or fuzzer. It applies to both `strict` and `max_atom_size`.

benches/deserialize.rs
arvidn, line 1: It's not obvious why you removed the benchmarks for the existing deserialization functions. I think they are still relevant

docs/serde-2026.md
arvidn: this document still starts with:

> Every 2026-format blob begins with a 6-byte magic prefix:

My comment in this thread was essentially that in the `FullBlock` generator, that's 6 wasted bytes.

One important consideration is that it won't be possible to tell with 100% confidence that an unfinished block that you just farmed will be infused before or after the hard fork. So it might be good to have some mechanism where we allow the current serialization even after the hard fork, and indicate in the full block which one we use. We already have plenty of spare bits to do this in though.

On the other hand, if the farmer guesses wrong, the block cost will be invalid as well, so maybe we'll just have to live with blocks being discarded around the activation

docs/serde-2026.md
arvidn: right, I was (incorrectly) thinking that we had a length prefix for the whole block, but we don't

arvidn: However, I think we've suffered a bit from not having a length prefix with the current serialization. Just reading a `FullBlock` from the database requires us to parse the CLVM structure. It would be a lot more efficient if we would encode the whole block generator as a buffer at the Streamable level. If we would do that, then you know the size of the buffer here.

This hard fork is an opportunity to fix it.

fuzz/fuzz_targets/serde_2026.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+fn roundtrip_check(label: &str, a: &Allocator, original: NodePtr, blob: &[u8]) {
+    let mut a2 = Allocator::new();
+    let decoded = deserialize_2026(&mut a2, blob, FUZZ_MAX_ATOM_LEN, false)
+        .unwrap_or_else(|e| panic!("{label}: deserialize failed: {e:?}"));
+    assert_eq!(
+        canonical(a, original),
+        canonical(&a2, decoded),
+        "{label}: tree mismatch"
```
arvidn: I would expect this to blow up memory usage. Is it not doing that for you?
If you use the same allocator, you can use `node_eq()` to compare the trees instead. you can checkpoint the allocator to discard the extra space before returning.

fuzz/fuzz_targets/serde_2026.rs
```diff
@@ -0,0 +1,73 @@
...
+            check_tree(&program.allocator, program.tree);
+
+            let mut a2 = Allocator::new();
+            let blob = serialize_2026(&program.allocator, program.tree, 0).expect("Fast failed");
+            let decoded = deserialize_2026(&mut a2, &blob, FUZZ_MAX_ATOM_LEN, false)
+                .expect("deserialize fast failed");
+            assert_eq!(
+                canonical(&program.allocator, program.tree),
+                canonical(&a2, decoded)
+            );
```
arvidn: same comment here, about memory usage and `node_eq()`

fuzz/fuzz_targets/serde_2026_panic.rs
```diff
@@ -0,0 +1,47 @@
...
+            strict,
+        );
+
+        // Prefix-aware deserializer — same body parser, plus magic-prefix strip.
+        let mut a = Allocator::new();
+        let _ = deserialize_2026(&mut a, data, FUZZ_MAX_ATOM_LEN, strict);
+
+        // Length probe — walks the wire format without building a tree, so
+        // it has its own opportunity to allocate or recurse pathologically.
+        let _ = serialized_length_serde_2026(data, FUZZ_MAX_ATOM_LEN, strict);
```
arvidn: this is a great opportunity to assert that the length this function returns exactly matches the number of bytes consumed by the two previous parsing functions

fuzz/fuzz_targets/serde_2026_varint.rs
```diff
@@ -0,0 +1,52 @@
...
+    write_varint(&mut buf, value).unwrap();
+    buf
+}
+
+fuzz_target!(|data: &[u8]| {
+    // Non-strict decode must never panic on arbitrary input.
+    let mut cur = Cursor::new(data);
+    let non_strict = read_varint(&mut cur, false);
+
+    if let Ok(v) = non_strict {
```
arvidn: minor style improvement (I think).


```suggestion
    let Ok(v) = non_strict else {
        // If non-strict failed, strict must also fail (strict is stricter).
        assert!(read_varint(&mut Cursor::new(data), true).is_err());
        return;
     };
```

It saves you an indentation

src/serde_2026/de.rs
```diff
@@ -0,0 +1,223 @@
...
+                let pair = allocator.new_pair(left, right)?;
+                pairs.push(pair);
+                stack.push(pair);
+            }
+            n if n >= 2 => {
+                let ai = (n - 2) as usize;
+                stack.push(*atoms.get(ai).ok_or(EvalErr::SerializationError)?);
+            }
+            n => {
+                let pi = (-n - 2) as usize;
```
arvidn:
```suggestion
                let pi = -(n + 2) as usize;
```

this seems like a legitimate issue. consider `n == i64::MIN`, you can't negate it.

I would have expected the fuzzer to find this.

src/serde_2026/de.rs
```diff
@@ -0,0 +1,223 @@
...
+///
+/// `max_atom_len` caps any declared atom length and `strict` rejects
+/// overlong / non-minimal varint encodings — pass the same values you would
+/// pass to the deserializer. For framing-only callers that don't want a
+/// policy opinion, `usize::MAX` and `false` accept anything the deserializer
+/// could itself parse on a sufficiently permissive caller.
+///
+/// The input buffer may contain trailing data; only the bytes belonging to
+/// the serde_2026 blob are counted.
+pub fn serialized_length_serde_2026(buf: &[u8], max_atom_len: usize, strict: bool) -> Result<u64> {
```
arvidn, line 175: if we encode the block generator as a buffer, we might not need this function

src/serde_2026/tests.rs
```diff
@@ -0,0 +1,426 @@
...
+        blob.extend_from_slice(&SERDE_2026_MAGIC_PREFIX);
+        blob.extend_from_slice(&encode_varint(group_count));
+        blob.extend_from_slice(atom_table);
+        blob.extend_from_slice(&encode_varint(instruction_count));
+        blob
+    };
+
+    // Cases that deserialize_2026_body rejects at header time. For each one,
+    // serialized_length_serde_2026 must also reject.
+    let cases: &[(&str, Vec<u8>)] = &[
```
arvidn: you could move this to `#[case()]` with `rstest

src/serde_2026/varint.rs
```diff
@@ -0,0 +1,229 @@
...
+    use std::io::Cursor;
+
+    fn encode_varint(value: i64) -> Vec<u8> {
+        let mut buf = Vec::new();
+        write_varint(&mut buf, value).unwrap();
+        buf
+    }
+
+    #[test]
+    fn test_encode_varint() {
```
arvidn: if you use `rstest` and make the test cases `#[case]` you don't need a separate helper function, the helper just becomes the test body.

I think we should have negative tests for `encode_varint` as well. It has an upper limit of 56 bits, right? It would be good to have tests to ensure it rejects values that large

wheel/python/clvm_rs/clvm_rs.pyi
```diff
@@ -10,11 +10,41 @@ def deserialize_as_tree(
...
 
-NO_NEG_DIV: int
+# --- Deserialize functions ---
+def deser_legacy(blob: bytes) -> "LazyNode": ...
+def deser_backrefs(blob: bytes) -> "LazyNode": ...
+def deser_2026(
+    blob: bytes,
+    *,
+    max_atom_len: int = ...,
+    strict: bool = True,
```
arvidn, line 20: it seems suspicious to have `strict` default to `false`

docs/serde-2026.md
richardkiss: I agree that length-prefixing the generator at the `FullBlock` / `Streamable` layer could be a useful improvement. I think that’s the right layer for it, rather than baking it into the `serde_2026` format itself.

The reason I’m hesitant to put the length into `serde_2026` is that the serializer can currently stream directly to its writer. It doesn’t know the final compressed size until it has finished writing, so adding an internal length prefix would require buffering the whole serialization first, or writing and then rewinding/patching the header. If `FullBlock` wants the generator as a length-delimited byte buffer, it can wrap the serialized bytes at that layer and get the database/read benefit without taking streaming away from the serializer.

src/serde_2026/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+        };
+
+        // Resize buffer once per length group
+        atom_buffer.resize(actual_length, 0);
+
+        for _ in 0..atom_count {
+            cursor
+                .read_exact(&mut atom_buffer)
+                .map_err(|_| EvalErr::SerializationError)?;
+            let atom_node = allocator.new_atom(&atom_buffer)?;
```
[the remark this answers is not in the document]
richardkiss: This is partially addressed. Negative atom_length/count values are now rejected via `checked_usize()` / `checked_bounded_usize()`, `i64::MIN` is handled before negation, zero length/count are rejected, and `atom_buffer.resize()` is capped by `max_atom_len`.

The remaining-input precheck is implemented in `serialized_length_serde_2026()` using checked multiplication and cursor bounds checks. The streaming deserializer itself can’t generally check “remaining input” because it takes `Read`; for bounded slice callers, `read_exact()` returns `SerializationError` when the declared atom table exceeds the input, and it only allocates one max_atom_len-bounded buffer.

docs/serde-2026.md
richardkiss: I've always preferred serializers that can stream to `Write` and deserializers that can work with a `Read` because you can build a buffer-based api on this one but not the other way around. So in that sense, streaming is more fundamental.

src/serde_2026/de.rs
richardkiss: These are gone now, to be revived in `chia_rs` since they're consensus-based.

src/serde_2026/de.rs
richardkiss: Also gone, reviving in `chia_rs`.

src/serde_2026/de.rs
richardkiss: This is gone now.

src/serde_2026/de.rs
```diff
@@ -0,0 +1,223 @@
...
+                stack.push(pair);
+            }
+            n if n >= 2 => {
+                let ai = (n - 2) as usize;
+                stack.push(*atoms.get(ai).ok_or(EvalErr::SerializationError)?);
+            }
+            n => {
+                let pi = (-n - 2) as usize;
+                stack.push(*pairs.get(pi).ok_or(EvalErr::SerializationError)?);
+            }
```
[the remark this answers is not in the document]
richardkiss: Fixed

src/serde_2026/de.rs
richardkiss: I would love to to get rid of this function if at all possible. One idea would be to replace the last two fields of `FullBlock` with

```
#[streamable]
  pub struct GeneratorInfo {
      generator: Program,              // or Bytes for the blob
      ref_list: Vec<u32>,
  }
```
and then we can just read to EOF in the initial parse. When validating the block, we parse out this new structure, and by this point we know if HF2 has kicked in, and bad parsing just become another consensus failure step.

We should probably discuss this idea more.

wheel/python/clvm_rs/clvm_rs.pyi
richardkiss: It's now `True`

docs/serde-2026.md
arvidn: the field that says how many instructions follow is a length prefix though, right? I'm not sure I get the distinction. I imagine you mean relying on EOF is problematic in some way, it doesn't compose like other fields. But in order to not rely on EOF you have to store your own length (just like a stremable buffer does).

Anyway, I think the serialization would become simpler and cheaper by not having to first create all the instructions, count them and then stream them. If you don't need the counter, you can just stream them directly as you create them. But, that relies on the deserializer receiving EOF when there aren't any more instructions. i.e. the upper layer needs to wrap the generator, but I think we should do that anyway.

docs/serde-2026.md
arvidn: I see you have test and fuzzing. So the use case is the main question I think.

src/serde_2026/ser.rs
arvidn: CI runs almost all tests in debug mode, but it also enables debug asserts in release builds when it runs the tests

wheel/python/clvm_rs/clvm_rs.pyi
arvidn: I think the justification should be flipped. If you can't think of any use case, it shouldn't be allowed

src/serde_2026/de.rs
arvidn: we have some precedence for how to extend streamable object in backwards compatible ways in `ProofOfSpace` (for v2 proofs). I can make a PR to show what I have in mind

arvidn (dismissed):
looks good.

arvidn (approved):

PR #803, opened 2026-06-26
`LIMITS` have activated

arvidn:
This is best reviewed one commit at a time.

I made a mistake in clvm_rs (that hasn't been released yet). in the latest softfork we activated a flag called LIMITS which imposes size limits on operands to the problematic operators (`multiply`, `div`, `divmod`, `mod` and some BLS operators). Since this flag has activated, it's now part of consensus. Any changes to behavior based on this flag is now a consensus change (as opposed to just a mempool change). Yesterday I extended this flag to cover modpow and last week I extended it to limit the nesting depth of `softfork`, both of those are technically wrong, but it doesn't matter in the case of modpow, because it was entirely disabled in the soft fork. but the `softfork` operator change would be a hard fork.

This fixes the situation by making the `softfork` operator limit depend on a new flag (`LIMIT_SOFTFORK`) that's only enabled in mempool mode. and then remove the LIMITS flag and update the operators to just have the new behavior.

---

> [!NOTE]
> **High Risk**
> This is consensus-critical CLVM behavior: unconditional operand limits and separating softfork depth from the activated softfork flag can change which programs succeed on-chain versus in mempool.
> 
> **Overview**
> Renames **`ClvmFlags::LIMITS`** to **`LIMIT_SOFTFORK`** and uses it only for the **20-level `softfork` nesting cap** in `run_program` and **`MEMPOOL_MODE`**. Operand size limits on **`multiply`**, **`div`/`divmod`/`mod`**, **`modpow`**, and **BLS g1/g2 multiply** are no longer flag-gated—they always apply in consensus.
> 
> Fuzz targets, benchmarks, and tests are updated to match: **`InvalidOpArg`** for oversized operands (not only cost exceeded), **`LIMITS`** removed from flag sweeps, factorial bench input reduced, and recursive div/mul program expectations switched to operand-limit errors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68e3bd41639d72fb9d67328379e4c12c49947a31. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/bls_ops.rs
```diff
@@ -81,7 +81,7 @@ pub fn op_bls_g1_multiply(
     a: &mut Allocator,
     input: NodePtr,
     max_cost: Cost,
-    flags: ClvmFlags,
+    _flags: ClvmFlags,
```
AmineKhaldi, line 84: Should `_flags` be removed then here and in the other op functions?

arvidn: no, the operators have a fixed signature. the CLVM interpreter keeps them in an array

AmineKhaldi (approved):

PR #839, opened 2026-09-15
optimize tree_cache

arvidn:
reuse arena and seen across calls to `find_path()`, to avoid heap allocation churn.

This gives my stress-test benchmark a 1.5x speed-up.

---

> [!NOTE]
> **Medium Risk**
> Touches CLVM compression back-reference path finding and adds `unsafe impl Sync`; behavior is covered by existing tests but incorrect reuse or Sync assumptions could affect serialization correctness or threading.
> 
> **Overview**
> **Reuses scratch memory in `TreeCache::find_path()`** to cut heap churn on repeated back-reference lookups (reported ~1.5× on a stress benchmark).
> 
> `TreeCache` now keeps a **`Bump` arena** and **`BitSet`** as fields, reset/cleared at the start of each `find_path()` instead of allocating fresh buffers every call. `find_path` takes **`&mut self`** accordingly. An **`unsafe impl Sync for TreeCache`** documents why scratch `Bump` is safe when the cache is shared only for PyO3/GIL reasons, not concurrent mutation.
> 
> **`BitSet`** gains **`clear()`** (zero marks, keep capacity) and **`ensure_size()`** (grow without the assert in `extend()`), plus unit tests.
> 
> The final path-vs-length check is refactored to an equivalent **`Option`** match; tree cache tests call **`find_path` once** per assertion where they previously invoked it twice.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 252898920a0ab98ef808a1f28ce044279795f501. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


arvidn:
@cursor review

richardkiss (approved):
This one's a bit scary, but I suppose its shelf life is finite since it won't be used after serde 2026 kicks in.

src/serde/tree_cache.rs
```diff
@@ -135,8 +135,22 @@ pub struct TreeCache {
...
+    /// checkpointed state.
+    scratch_seen: BitSet,
 }
 
+// Bump is !Sync (interior mutability via Cell). TreeCache only touches
+// scratch_arena through &mut self (find_path), so sharing &TreeCache across
+// threads cannot race on it. Sync is required so Serializer can live in PyO3
+// types that release the GIL; that does not imply concurrent access.
+unsafe impl Sync for TreeCache {}
+
```
richardkiss, line 153: Claude is telling me you don't actually need `unsafe` here, and that it took it out and it compiled fine still. You may want to look into that.

arvidn: this new field `scratch_arena: Bump` is not `Sync`. The `Sync` trait is only required by the python wheel, because PyO3 uses it as a proxy for it being safe to release the GIL. So, the failure happens when building the python wheel, specifically in `chia_rs`. If I remove `unsafe` it fails with this error:

```
error[E0200]: the trait `Sync` requires an `unsafe impl` declaration
   --> /Users/arvid/Documents/dev/2/clvm_rs/src/serde/tree_cache.rs:152:1
    |
152 | impl Sync for TreeCache {}
    | ^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: the trait `Sync` enforces invariants that the compiler can't check. Review the trait documentation and make sure this implementation upholds those invariants before adding the `unsafe` keyword
help: add `unsafe` to this trait implementation
    |
152 | unsafe impl Sync for TreeCache {}
    | ++++++
```

If I remove the trait altogether, it fails with:

```
error[E0277]: `Cell<NonNull<bumpalo::ChunkFooter>>` cannot be shared between threads safely
   --> crates/chia-consensus/src/build_compressed_block.rs:42:12
    |
 42 | pub struct BlockBuilder {
    |            ^^^^^^^^^^^^ `Cell<NonNull<bumpalo::ChunkFooter>>` cannot be shared between threads safely
    |
    = help: within `BlockBuilder`, the trait `Sync` is not implemented for `Cell<NonNull<bumpalo::ChunkFooter>>`
    = note: if you want to do aliasing and mutation between multiple threads, use `std::sync::RwLock`
note: required because it appears within the type `bumpalo::Bump`
   --> /Users/arvid/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bumpalo-3.19.1/src/lib.rs:292:12
    |
292 | pub struct Bump<const MIN_ALIGN: usize = 1> {
    |            ^^^^
note: required because it appears within the type `clvmr::serde::TreeCache`
   --> /Users/arvid/Documents/dev/2/clvm_rs/src/serde/tree_cache.rs:95:12
    |
 95 | pub struct TreeCache {
    |            ^^^^^^^^^
note: required because it appears within the type `Serializer`
   --> /Users/arvid/Documents/dev/2/clvm_rs/src/serde/incremental.rs:19:12
    |
 19 | pub struct Serializer {
    |            ^^^^^^^^^^
note: required because it appears within the type `BlockBuilder`
   --> crates/chia-consensus/src/build_compressed_block.rs:42:12
    |
 42 | pub struct BlockBuilder {
    |            ^^^^^^^^^^^^
note: required by a bound in `assert_pyclass_send_sync`
   --> /Users/arvid/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/pyo3-0.29.0/src/impl_/pyclass/assertions.rs:6:15
    |
  4 | pub const fn assert_pyclass_send_sync<T>()
    |              ------------------------ required by a bound in this function
  5 | where
  6 |     T: Send + Sync,
    |               ^^^^ required by this bound in `assert_pyclass_send_sync`
```

